PR #1, opened 2019-08-09
Py2

joshmoore:
Following the style of https://github.com/ome/omero-py/pull/1, this PR looks to make this extracted ome/openmicroscopy:components/tools/OmeroWeb directory `pip`-installable with Python 2.7 _before_ work starts on upgrading to Python 3. From there, the main directory will be removed from ome/openmicroscopy and all development should occur on this repo.

An extra difficulty is the https://github.com/ome/omero-py/blob/master/src/omero/plugins/web.py script which has hard-coded assumptions about where `manage.py` lives. 

Once OMERO.web is starting, propose to tag and release this to PyPI as 5.5.dev1.


joshmoore:
Currently testing via:

```
[jamoore@idr1-slot2 omero-web-docker]$ git diff
diff --git a/Dockerfile b/Dockerfile
index 410ee0b..5a23f58 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,3 +31,11 @@ EXPOSE 4080
 VOLUME ["/opt/omero/web/OMERO.web/var"]

 ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
+
+USER root
+RUN rm -vrf /opt/omero/web/OMERO.web/lib/python/omeroweb
+RUN /opt/omero/web/venv/bin/pip install https://github.com/ome/omero-py/archive/master.zip
+RUN /opt/omero/web/venv/bin/pip install https://github.com/joshmoore/omero-web/archive/py2.zip
+ENV OMERODIR /opt/omero/web/OMERO.web
+ENV OMERO_HOME /opt/omero/web/OMERO.web
+USER omero-web
```

PR #4, opened 2019-08-22
OMERO.web use client (browser) timezone

manics:
Just testing a possible way of porting PRs:
`curl -sSL https://github.com/ome/openmicroscopy/pull/6010.patch | sed 's%/components/tools/OmeroWeb/%/%g' | git am`

Ported from https://github.com/ome/openmicroscopy/pull/6010

----

# What this PR does

This is an alternative solution to https://github.com/openmicroscopy/openmicroscopy/pull/5786 (and it reverts https://github.com/openmicroscopy/openmicroscopy/pull/6009)
@pwalczysko @will-moore 

This PR moves all OMERO.web datetime processing to the client browser using Javascript, which can optionally take account of the client's timezone. In cases where the timestamps were previously rendered with Django they are now stored as a `data-isodate` HTML attribute, and are automatically converted for display by `OME.formatDate`.

# Two configurations should be tested

## 1. All timestamps are UTC

The assumption is that the backend OMERO.server stores all timestamps in UTC. The timestamps are then converted to the client browsers local timezone for display. This is done client-side using Javascript, the OMERO.web server does not need to care about the client's locality, and this should therefore work for clients in multiple timezones.

OMERO.web configuration (this is the default in this PR):
```
omero config set omero.web.server_returns_utc true
```

### Current timezone
Open a browser. Create objects (e.g. projects, annotations) in OMERO.web. Check that the displayed timestamps are as expected. If they are not please verify that OMERO.server really is using UTC by requesting the timestamps using the OMERO API (e.g. Python BlitzGateway) and checking the raw timestamps.

### Different timezone
Fly to Australia, if you are using a Mac your laptop may auto-configure itself for the new timezone, otherwise update it in system preferences. If this is not possible you may be able to simulate a different timezone, e.g. On OS X with Chrome:
```
mkdir $HOME/chrome-profile
TZ='Australia/ACT' open -na "Google Chrome" --args "--user-data-dir=$HOME/chrome-profile"
```
Look at the objects created in the previous step, the timestamps should be adjusted to the new timezone. Note you can only have one instance of chrome using this custom profile directory.

## 2. Timestamps are localtime

The assumption is that the backend OMERO.server stores all timestamps in the client's timezone. Therefore no conversion is done when it is displayed in the browser. Obviously if you have clients in multiple timezones, of if the server timezone is not the same as the clients, they will all see the wrong time.

OMERO.web configuration:
```
omero config set omero.web.server_returns_utc false
```

As before open a browser in your current and different timezones. You should see that the displayed timestamps are always the same.

# Other notes
This PR changes the default OMERO.web timezone to `UTC` since nothing else really makes sense. I'm not sure if it makes sense for it to be configurable, but I guess there's no harm.

# Background

- https://github.com/openmicroscopy/openmicroscopy/pull/5786#issuecomment-484155143
- https://trello.com/c/HEXZNdSC/227-omeroweb-default-timezone-is-different-from-omeroserver

### DIFF
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -237,6 +237,18 @@ def identity(x):
     return x
 
 
+def json_or_plain_str(s):
+    """
+    Most omero.web properties are JSON objects but some are plain strings.
+    Use this to allow both JSON objects/strings and plain strings. Obviously
+    this can only be used when the plain string can't also be a JSON object.
+    """
+    try:
+        return json.loads(s)
+    except ValueError:
+        return s
+
+
 def str_slash(s):
     if s is not None:
         s = str(s)
@@ -850,6 +862,15 @@ CUSTOM_SETTINGS_MAPPINGS = {
          "Whether to allow OMERO.web to be loaded in a frame."
          ],
 
+    "omero.web.timezone":
+        ["TIME_ZONE",
+         "UTC",
+         json_or_plain_str,
+         ("Local time zone for this installation. Choices can be found here: "
+          "http://www.postgresql.org/docs/8.1/static/datetime-keywords.html"
+          "#DATETIME-TIMEZONE-SET-TABLE although not all variations may be "
+          "possible on all operating systems. Default ``\"UTC\"``")],
+
     "omero.web.django_additional_settings":
         ["DJANGO_ADDITIONAL_SETTINGS",
          "[]",
@@ -1086,10 +1107,6 @@ report_settings(sys.modules[__name__])
 
 SITE_ID = 1
 
-# Local time zone for this installation. Choices can be found here:
-# http://www.postgresql.org/docs/8.1/static/datetime-keywords.html#DATETIME-TIMEZONE-SET-TABLE
-# although not all variations may be possible on all operating systems.
-TIME_ZONE = 'Europe/London'
 FIRST_DAY_OF_WEEK = 0     # 0-Monday, ... 6-Sunday
 
 # LANGUAGE_CODE: A string representing the language code for this
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -862,7 +862,7 @@ CUSTOM_SETTINGS_MAPPINGS = {
          "Whether to allow OMERO.web to be loaded in a frame."
          ],
 
-    "omero.web.timezone":
+    "omero.web.time_zone":
         ["TIME_ZONE",
          "UTC",
          json_or_plain_str,
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -866,10 +866,10 @@ CUSTOM_SETTINGS_MAPPINGS = {
         ["TIME_ZONE",
          "UTC",
          json_or_plain_str,
-         ("Local time zone for this installation. Choices can be found here: "
-          "http://www.postgresql.org/docs/8.1/static/datetime-keywords.html"
-          "#DATETIME-TIMEZONE-SET-TABLE although not all variations may be "
-          "possible on all operating systems. Default ``\"UTC\"``")],
+         ("Time zone for this installation. Choices can be found in the "
+          "``TZ database name`` column of: "
+          "https://en.wikipedia.org/wiki/List_of_tz_database_time_zones "
+          "Default ``\"UTC\"``")],
 
     "omero.web.django_additional_settings":
         ["DJANGO_ADDITIONAL_SETTINGS",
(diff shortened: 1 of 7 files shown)


snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#54](https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/54/). See the [console output](https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/54/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - test/integration/test_tree_annotations.py
    - test/integration/test_tree.py

--conflicts

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#81](https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/81/). See the [console output](https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/81/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - test/integration/test_tree_annotations.py
    - test/integration/test_tree.py

--conflicts

manics:
GMT:
Screen Shot 2019-12-17 at 11 20 43

ACT (GMT+11)
Screen Shot 2019-12-17 at 11 20 49

manics:
@chris-allan any thoughts on how this will affect @glencoesoftware's fork?

chris-allan:
I'll have to dedicate some time to reviewing the implications specific to us. Off the top of my head I can't think of anything that is unique.

This PR is obviously largely scoped to the right hand side metadata panel and that is a quite narrow use of dates in OMERO.web. It's probably not a good thing to have dates mismatched between areas of the UI and that will very much happen with this PR as it stands. However, aside from reviewing date display across the entirety of OMERO.web and all plugins I don't have a good handle on how to approach resolving that dissonance. It is a very big job.

In the wild, I think it is very much the exception that the server has timestamps at UTC. It is for example not the case on *any* commercial install we have ever performed or open source install we have consulted on. Having that as the default seems wrong.

will-moore:
Seems like this is a lot of work to get right across the board.
I guess the short-term question is whether this PR is an improvement on the current status?

Maybe we need a matrix of:
 - Same timezone as server?
 - Is it Summer time?

And test those 4 cases against
 - Current status (before this PR)
 - With this PR & server_returns_utc = True
 - With this PR & server_returns_utc = False

asking the question: If I create a Project/Dataset/Screen, does the timestamp match my Wall Clock?

Timestamp displays on other Objects:
 - Annotations (Tags, Comment, File, Maps, Tables)
 - Image (Acquisition date & Import date).

It looks like e.g. Map Annotations on a browser in a different time zone, the annotation time doesn't match the Wall Clock with this PR on py3-ci.

Screen Shot 2019-12-19 at 11 22 32

Just tried https://py3-ci.openmicroscopy.org/web/webclient/?show=image-3065 (user-3) Tables:

Screen Shot 2019-12-19 at 11 11 37

Looks like Django timezone functionality may help, but that seems to have a few gotchas and would be a lot of work: https://docs.djangoproject.com/en/1.11/topics/i18n/timezones/

I'm afraid I don't think this is ready to merge without at least extending / fixing the logic for Annotations.
But we probably need a review of how bad the current situation is first. For whom is it currently broken (in the Same-Timezone/Summertime/Object matrix above)?

will-moore:
Reporting of time-stamp issues at https://forum.image.sc/t/omero-webclient-import-date-modify-timezone/54384

stephenogg:
Thanks for opening this again Will - I've been able to deploy this PR to compare how it renders times to the current released version of omero-web. I'll provide feedback as I can- but first, like you, I'll have to look at the changes Simon proposed and see if I can figure out what they do and why the annotations display this NaN time when omero.web.server_returns_utc=true

This is a difficult problem. One thing that I think needs to be thought out more - is the initial discussion above that timestamps somehow can be either in UTC or in local time. I didn't appreciate this until I started to look into this, but timestamps are just moments in time, _always_ from the unix epoch at **UTC**.  So an image uploaded with a timestamp of 0, would have an import date of Jan 1, 1970 00:00 if that timestamp is decoded to be imported from London, but would have an import date of Dec 31, 1969 16:00 if that 0 is decoded in the Mountain timezone. This 0 timestamp represents the exact same moment in time in the two locations, but has different local displays. Without knowing the location or the timezone of the import, you **can't** actually know how to decode the timestamp and display it consistently from one location to another. I don't know if this information is captured with the import?

You  could choose to always display it in the browser's local timezone, no matter who is browsing the image or from where. Or you choose to always display it in some reference timezone, like UTC - but then you have the original problem that for users who aren't in the UTC, it seems as if the import time is incorrect. Therefore a label becomes important. All of this will also be true for other things like annotations and creation times of other objects.

PR #5, opened 2019-08-23
Add .travis.yml

joshmoore:
Copied from omero-figure


joshmoore:
Merging to get the other PRs green. This will need some more work to actually do the testing we want. `flake8`, however, is going to take a while.

cc: @ome/python

PR #16, opened 2019-09-18
Link image

will-moore:
Ported PR from https://github.com/ome/openmicroscopy/pull/6087

To test, see PR.

### DIFF
--- a/omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
+++ b/omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
@@ -62,7 +62,14 @@ OME.getURLParameter = function(key) {
 
 var linkify = function(input) {
     var regex = /(https?|ftp|file):\/\/[-a-zA-Z0-9+&@#/%?=~_|!:,.;]*[-a-zA-Z0-9+&@#/%=~_|]/g;
-    return input.replace(regex, "<a href='$&' target='_blank'>$&</a>");
+    input = input.replace(regex, "<a href='$&' target='_blank'>$&</a>");
+    return linkObjects(input);
+};
+var linkObjects = function(input) {
+    var regex = /(Project|Dataset|Image|Screen|Plate|Well|ROI)(?: ID)*:(?: )*([0-9]+)/g;
+    return input.replace(regex, function($0, $1, $2) {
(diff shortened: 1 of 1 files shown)


jburel:
tested  against
https://merge-ci.openmicroscopy.org/web/webclient/?show=image-91366
with and without ID in the description
space no space after ``:``

omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
```diff
@@ -62,7 +62,14 @@ OME.getURLParameter = function(key) {
 
 var linkify = function(input) {
     var regex = /(https?|ftp|file):\/\/[-a-zA-Z0-9+&@#/%?=~_|!:,.;]*[-a-zA-Z0-9+&@#/%=~_|]/g;
-    return input.replace(regex, "<a href='$&' target='_blank'>$&</a>");
+    input = input.replace(regex, "<a href='$&' target='_blank'>$&</a>");
+    return linkObjects(input);
+};
+var linkObjects = function(input) {
+    var regex = /(Project|Dataset|Image|Screen|Plate|Well|ROI)(?: ID)*:(?: )*([0-9]+)/g;
```
jburel: Maybe we want to support the old insight case for projection
i.e. Image|image and ID|id same for the other type
this makes the regex a bit more complicated but it will cover the case

will-moore: This already supports:
 - Dataset:1
 - Project: 2
 - Image ID: 3
I already think this is more than ideal since it's harder to search, handle, remember all the different formats in various clients if we allow more formats. E.g. other 3rd party clients or scripts will be less strict in what they produce if we support more.
This already satisfies the original requirement of handling Image links from Figure (and now handles a lot more objects and formats) so I think it's time to get it merged and not think of every possible variation.
It was already approved of at https://github.com/ome/openmicroscopy/pull/6087 so if it's working here then 👍.

jburel: but not ``image id`` i.e. lower case

jburel: it handles figure true but it will be good to support previous link.
Insight has been adjusted to support new and old formats

jburel:
Thanks both lower case and upper case work
Ready to merge. 
I will hold off before merging since @joshmoore mentioned tagging the repo

jburel:
Merging this PR
js changes that work

PR #20, opened 2019-09-25
Replaces assert (True) in multiline webgateway tests, flake8

manics:
Replaces assert (True) in multiline webgateway tests
Add flake8 (somehow got lost), version is 2.4.0 to match https://github.com/ome/openmicroscopy/blob/v5.5.1/Dockerfile#L32


joshmoore:
Changes all make sense and look fine. I do wonder if the flake call shouldn't be in tox though. I'm trying that now on the omero-py repository. I'll open a comparison PR.

PR #22, opened 2019-10-03
Filter by map anns webclient

will-moore:
As requested in various workshops, allow filtering of images by Map Annotations (without needing to install OMERO.parade).
This also recognises when all the values for a chosen Key are numbers, and then lets you do numerical filtering.

Screen Shot 2019-10-03 at 10 24 03

To test:
 - Pick a Dataset of Images with a good selection of Map Annotations (all namespaces are used), including some 'numerical' values.
 - Choose to filter by 'Key-Values'.
 - Pick a Key - images without that key will be immediately filtered out
 - If you've picked a Key where all the values are numbers, you will see a ```>``` chooser and the min-max range will be shown as a placeholder.
 - Otherwise, you'll get a regular text filter.
 - Typing in the filter field will start filtering.
 - Additional Map-Annotation filters can be added, also with other filters.
 - Clicking the X will remove the filter and un-filter the images.

cc @pwalczysko


will-moore:
Travis failing with ```.. ERROR:: You need Python 3.4 or greater to install PyTables! ```.

joshmoore:
Pushed a fix, @will-moore

pwalczysko:
this works as expected

2 RFEs (maybe for a follow-up PR):

  1. (more important) - would it be possible for the non-numerical values to have a pre-population as you are typing into the "Value" box ? It is hard to know what is really there. Best would be pre-suggestions aka mapr.
  2. For the numerical values, the suggestion of the available range is there at the beginning in the box which is good (see screenshot). But, would it be possible to have it also in the tooltip somewhere aka parade ? I started to type values, then forgot the range, and the suggestion was not there as typed over it in the previous step.

Screen Shot 2019-10-23 at 10 09 02

will-moore:
Auto-complete should now be working for Map values.
Also, tooltip for range of numbers.

pwalczysko:
Tested on py3-ci. Both additional features work as expected. 

Ready to merge fmpov

joshmoore:
Minor issue with the layout on small displays:
Screen Shot 2019-12-18 at 09 45 39

joshmoore:
Merging to not let @pwalczysko's testing be kept from the user. @will-moore: can you create an issue for the layout issue and for looking at after the 5.6.0 release?

Issue #23, opened 2019-10-03
webadmin handle lookupLdapAuthExperimenter


will-moore:
When loading an Experimenter, for e.g. ```/webadmin/experimenter/edit/id``` this fails if there is an exception when looking up LDAP user, e.g:

https://www.openmicroscopy.org/qa2/qa/feedback/27937/
https://www.openmicroscopy.org/qa2/qa/feedback/27865/
 https://www.openmicroscopy.org/qa2/qa/feedback/27767/
 https://www.openmicroscopy.org/qa2/qa/feedback/18791/
```
Traceback (most recent call last):

File "/usr/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/home/omero/OMERO.py/lib/python/omeroweb/decorators.py", line 486, in wrapped
retval = f(request, *args, **kwargs)

File "/home/omero/OMERO.py/lib/python/omeroweb/decorators.py", line 536, in wrapper
context = f(request, *args, **kwargs)

File "/home/omero/OMERO.py/lib/python/omeroweb/webadmin/views.py", line 478, in manage_experimenter
prepare_experimenter(conn, eid)

File "/home/omero/OMERO.py/lib/python/omeroweb/webadmin/views.py", line 107, in prepare_experimenter
isLdapUser = experimenter.isLdapUser()

File "/home/omero/OMERO.py/lib/python/omeroweb/webclient/webclient_gateway.py", line 2289, in isLdapUser
self.ldapUser = admin_serv.lookupLdapAuthExperimenter(self.id)

File "/home/omero/OMERO.py/lib/python/omero/gateway/__init__.py", line 4671, in __call__
return self.handle_exception(e, *args, **kwargs)

File "/home/omero/OMERO.py/lib/python/omeroweb/webclient/webclient_gateway.py", line 2109, in handle_exception
e, *args, **kwargs)

File "/home/omero/OMERO.py/lib/python/omero/gateway/__init__.py", line 4668, in __call__
return self.f(*args, **kwargs)

File "/home/omero/OMERO.py/lib/python/omero_api_IAdmin_ice.py", line 1123, in lookupLdapAuthExperimenter
return _M_omero.api.IAdmin._op_lookupLdapAuthExperimenter.invoke(self, ((id, ), _ctx))

ApiUsageException: exception ::omero::ApiUsageException
{
[...]

PR #24, opened 2019-10-03
Handle lookupLdapAuthExperimenter exception

will-moore:
See https://github.com/ome/omero-web/issues/23

E.g. for experimenter/edit/id this would still allow the page to load and show message in place of LDAP info:

Screen Shot 2019-10-03 at 11 05 03

NB: screenshot was created by manually manipulating the code, not via real LDAP testing.
To test: need to fail to look up LDAP user.

### DIFF
--- a/omeroweb/webadmin/views.py
+++ b/omeroweb/webadmin/views.py
@@ -53,6 +53,7 @@ from forms import UploadPhotoForm, EmailForm
 from omeroweb.http import HttpJPEGResponse
 from omeroweb.webclient.decorators import login_required, render_response
 from omeroweb.connector import Connector
+from omero import ApiUsageException
 
 logger = logging.getLogger(__name__)
 
@@ -104,7 +105,12 @@ def prepare_experimenter(conn, eid=None):
     defaultGroup = experimenter.getDefaultGroup()
     otherGroups = list(experimenter.getOtherGroups())
     hasAvatar = conn.hasExperimenterPhoto()
-    isLdapUser = experimenter.isLdapUser()
+    try:
+        isLdapUser = experimenter.isLdapUser()
+    except ApiUsageException as x:
+        # e.g. "Cannot find unique user DistinguishedName: found=0"
+        logger.error(traceback.format_exc())
+        isLdapUser = x.message
     return experimenter, defaultGroup, otherGroups, isLdapUser, hasAvatar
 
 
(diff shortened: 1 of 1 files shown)


omeroweb/webadmin/views.py
```diff
@@ -104,7 +105,12 @@ def prepare_experimenter(conn, eid=None):
     defaultGroup = experimenter.getDefaultGroup()
     otherGroups = list(experimenter.getOtherGroups())
     hasAvatar = conn.hasExperimenterPhoto()
-    isLdapUser = experimenter.isLdapUser()
+    try:
+        isLdapUser = experimenter.isLdapUser()
+    except ApiUsageException as x:
+        # e.g. "Cannot find unique user DistinguishedName: found=0"
+        logger.error(traceback.format_exc())
+        isLdapUser = x.message
```
joshmoore: Wouldn't this resolve to `true`?

omeroweb/webadmin/views.py
```diff
@@ -104,7 +105,12 @@ def prepare_experimenter(conn, eid=None):
     defaultGroup = experimenter.getDefaultGroup()
     otherGroups = list(experimenter.getOtherGroups())
     hasAvatar = conn.hasExperimenterPhoto()
-    isLdapUser = experimenter.isLdapUser()
+    try:
+        isLdapUser = experimenter.isLdapUser()
+    except ApiUsageException as x:
+        # e.g. "Cannot find unique user DistinguishedName: found=0"
+        logger.error(traceback.format_exc())
```
joshmoore: I see that this might be a different case, but in general, ApiUsageException's are _not_ errors.

omeroweb/webadmin/views.py
will-moore: ```isLdapUser``` is named confusingly. It is actually a string, as returned by ```lookupLdapAuthExperimenter()``` https://downloads.openmicroscopy.org/omero/5.4.10/api/slice2html/omero/api/IAdmin.html#lookupLdapAuthExperimenter (see https://github.com/ome/omero-web/blob/master/omeroweb/webclient/webclient_gateway.py#L2298)
and displayed as in the screenshot above.

If the user is not an LDAP user then this returns ```None``` which evaluates to false in the experimenter_form. 

```
    {% if ldapAuth %}
        <label>LDAP:</label> {{ ldapAuth }}
    {% else %}
```

joshmoore: Weird! Thanks.

joshmoore:
Merging as a straight-forward exception handling though I imagine there's room for refactoring (e.g. deprecating `isLdapUser` and creating a more understandably named method) if we ever get around to it.

Issue #6, opened 2019-07-30
Bug: web closes used table


joshmoore:
see:
 * https://forum.image.sc/t/server-error-after-populating-metadata-for-images-using-web-script/27844
 * https://www.openmicroscopy.org/qa2/qa/feedback/27793
 * https://github.com/openmicroscopy/openmicroscopy/pull/6070

Multiple (browser?) threads accessing the same table proxy can lead to a `NoneType` exception on `table.getHeader()` if one thread calls `close()` while other threads are still active.

Options seem to be:
 - client: use multiple services
 - client: only close when complete
 - client: re-open on the fly
 - server: re-open on the fly
 - (longer-term): use a stateless interface

joshmoore:
Sorry, @will-moore. I missed this go by. Does the above mean that https://github.com/ome/omero-web/issues/6#issuecomment-522952150 isn't a possible solution?

will-moore:
No, I think the JavaScript calls "in series" will fix this issue.
But longer term, it seems that a "lower level" fix would be more robust.
Also it would be nice not to try 6 different ways to get the data since if we don't get it till the 6th try, then it's going to feel a bit slower than you'd like.

will-moore:
https://github.com/ome/omero-web/pull/25 for a potential fix/workaround.

will-moore:
Reported again at https://www.openmicroscopy.org/qa2/qa/feedback/28005/

will-moore:
This seems to be fixed by #25. Closing...

PR #25, opened 2019-10-03
Avoid parallel table.close() on the same table

will-moore:
For tables on Projects or Datasets, we try 2 queries ```image=1``` (lowercase) and ```Image=1``` on the same Table at the same time, which leads to errors. See https://github.com/ome/omero-web/issues/6

This changes the webclient right panel to make each pair of queries run in series instead of parallel.

Screenshot shows calls in Chrome dev-tools before this PR, where we try to load table data on a parent Screen, Plate, Project (x2) and Dataset (x2) all at once. 

Screen Shot 2019-10-03 at 14 47 58

With this PR, we wait for the first queries on Project and Dataset to return, and if they have no data then we make another call with the alternative query:

Screen Shot 2019-10-03 at 14 40 45

To test, check that table data is shown as expected with no error dialogs.

This can be tested at https://merge-ci.openmicroscopy.org/web/webclient/?show=dataset-4919 (user-3), which was the sample data originally used to replicate the error.
NB: In this data, the first call to get table data from Dataset is successful, so a second call is never made.

### DIFF
--- a/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
+++ b/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
@@ -356,6 +356,7 @@
                     {% endif %}
 
                     var showBulkAnnTooltip = function(data) {
+                        if (!data || !data.id) return;
                         var bulkAnnTooltip = "<span class='tooltip_html' style='display:none'>" +
                             "Data from tables file:<br />" +
                             "<b>File ID:</b> " + data.id + "<br />" +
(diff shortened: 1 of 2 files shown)


jburel:
Time to go back to this PR?

will-moore:
This can be tested at https://merge-ci.openmicroscopy.org/web/webclient/?show=dataset-4919 (user-3), which was the sample data originally used to replicate the error.
NB: In this data, the first call to get table data from Dataset is successful, so a second call is never made.

pwalczysko:
Tables are shown as expected on py3-ci and merge-ci. I tried (on merge-ci) both the orphaned dataset indicated in the PR header as well as the idr0021 project table. On py3-ci, I have tried only the idr0021 project table.

I found the "waterfall" feature of the developer tools in chrome (at least temporarily) but did not see any of the green bars as shown in the header of this PR - maybe I need ot readjust something. ?

joshmoore:
@pwalczysko : i.e. you need to help to finish this review?

pwalczysko:
@will-moore helped me to find the "waterfall" feature in the browser. 

Again, confirmed that I am not getting the errors in the UI and also, my screenshots (see below) look as expected in the sense of the two calls for image and Image waiting for one another as expected (i.e. the image call waits for the Image call to finish before being sent off). This was verified by reading the tooltips on the lines in the "waterfall" table. Tested both table on a Project as well as table on a Dataset.
Screen Shot 2019-11-28 at 15 20 28

Screen Shot 2019-11-28 at 15 23 50


Ready to merge fmpov

PR #30, opened 2019-10-24
Add Developer installation to README

will-moore:
Stuff I did in addition to README steps:

- Did dev installation of  ```omero-py```
- pip install future  (should this have installed as dependency? need to add it somewhere?)
- Set ```$OMERODIR``` to ```path/to/openmicroscopy/dist/```
- Installed omero-marshal from origin/master (released version is installed with omero-web but we need https://github.com/ome/omero-marshal/pull/63
```
pip uninstall omero-marshal
cd 
cd Desktop/MARSHAL/omero-marshal/
git fetch origin
git checkout origin/master
pip install -e .
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,7 +50,7 @@ For a development installation we recommend creating a virtualenv with the follo
     cd omero-web
     pip install -e .
 
-This will install OMERO.web into your virtualenv as an editable package, so any edits to ``src`` files should be reflected in your installation.
+This will install OMERO.web into your virtualenv as an editable package, so any edits to source files should be reflected in your installation.
 Note that if you add or remove files you must rerun the last step.
 
 Running tests
--- a/README.rst
+++ b/README.rst
@@ -32,6 +32,27 @@ Contributing
 
 See: `OMERO`_ documentation
 
+Developer installation
+----------------------
+
+OMERO.web depends on OMERO.py. If you want a developer installation of OMERO.py, replace ``pip install omero-py``
+with instructions at https://github.com/ome/omero-py.
+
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -32,6 +32,27 @@ Contributing
...
+
+    python3.6 -mvenv venv
+    . venv/bin/activate
+    pip install zeroc-ice==3.6.5
+    pip install omero-py          # OR dev install (see above)
+    git clone https://github.com/ome/omero-web
+    cd omero-web
+    pip install -e .
+
+This will install OMERO.web into your virtualenv as an editable package, so any edits to ``src`` files should be reflected in your installation.
```
joshmoore:
```suggestion
This will install OMERO.web into your virtualenv as an editable package, so any edits to source files should be reflected in your installation.
```

joshmoore: Reminds me that edits to `target/` files in omero-py are also reflected. cc: @manics

will-moore:
Would be good to tag ```omero-marshall``` and update the requirement so we don't need the manual install above.

joshmoore:
What version would you propose? `0.7.0`, `0.7.dev1`?

cc: @ome/python

README.rst
```diff
@@ -32,6 +32,27 @@ Contributing
...
+    python3.6 -mvenv venv
+    . venv/bin/activate
+    pip install zeroc-ice==3.6.5
+    pip install omero-py          # OR dev install (see above)
+    git clone https://github.com/ome/omero-web
+    cd omero-web
+    pip install -e .
+
+This will install OMERO.web into your virtualenv as an editable package, so any edits to source files should be reflected in your installation.
+Note that if you add or remove files you must rerun the last step.
```
manics: This shouldn't be necessary, it's only required in omero-py because of the separate external blitz target that has to be merged in

PR #33, opened 2019-10-29
futurize testlib

joshmoore:
Needed to enable integration tests under py3-ci


joshmoore:
Py3 tests now running :+1:

PR #35, opened 2019-11-01
Annotation marshal: fix Python2/3 compatibility

sbesson:
Required to fix https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/31/testReport/junit/OmeroWeb.test.integration.test_tree_annotations/TestTreeAnnotation

7757875 allows to use `list.sort(key=key)`
97d7e4e  ensures the format of the timestamp returned by `datetime` is identical between Python 2 and Python 3. Without this change, `isoformat()` contains millisecond information.

### DIFF
--- a/omeroweb/webclient/tree.py
+++ b/omeroweb/webclient/tree.py
@@ -22,6 +22,7 @@
 import time
 import omero
 from builtins import bytes
+from past.utils import old_div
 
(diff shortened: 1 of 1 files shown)


joshmoore:
test_tree_annotation passed on merge (222) and py3 (37)

PR #39, opened 2019-11-01
history

jburel:
remove usage of old functions

This should fix some of the history tests

See also https://github.com/ome/openmicroscopy/pull/6160


joshmoore:
Don't see any history failures in merge (222) or py3 (37)

PR #45, opened 2019-11-02
Remove long in webclient gateway

will-moore:
Since a lot of tests have needed these fixes, and they also occur in a lot of places not covered by tests, this aims to remove all usages of ```long(id)``` and use ```int(id)``` instead.

As described in https://portingguide.readthedocs.io/en/latest/numbers.html#unification-of-int-and-long, we don't really need ```long``` in python2 and it's removed in python3.

### DIFF
--- a/omeroweb/webadmin/views.py
+++ b/omeroweb/webadmin/views.py
@@ -280,7 +280,7 @@ def drivespace_json(request, query=None, groupId=None, userId=None, conn=None,
                                   "userId": e.getId()})
 
     elif userId is not None:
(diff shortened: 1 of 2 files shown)


joshmoore:
Very green day on rebuild yesterday (merge 227 and py3 43). :+1:

sbesson:
The removal of `long` in `omeroweb/webadmin/views.py` broke the Travis flake8 checks due to an unnecessary import. Unclear why this did not get picked up by the PR status but this should be fixed by 54b74f2

PR #55, opened 2019-11-06
Bump omero-web OME dependencies

sbesson:
Increment the minimal versions of the omero-py and omero-marshal packages to include the Python 3 fixes


sbesson:
Merging and tagging

PR #50, opened 2019-11-04
Fix Tag creation in webclient

will-moore:
Fix bug reported at:
https://docs.google.com/spreadsheets/d/1ry6vS0dmsylFQDAiIZDrPC3kXNpnHSMtK-2yj5X8jaU/edit#gid=420881033

"Add Tag fails if tag is new: Click on the plus button in Tags harmonice in RHP and try to create a new tag. This fails after clicking "Save" with no error displayed (just a blank form "Send feedback" appears). The addition of tags works if these tags are created previously via the Tags tab."

PR #56, opened 2019-11-06
copy/paste right_plugin.general.js.html instead of include it

will-moore:
Trying to fix https://github.com/ome/omero-web/pull/50#issuecomment-549707554
which is only seen on merge-ci (works fine locally and on py3-ci).

### DIFF
--- a/omeroweb/webclient/templates/webclient/base/base_container.html
+++ b/omeroweb/webclient/templates/webclient/base/base_container.html
@@ -274,7 +274,146 @@
     {% include "webclient/base/includes/toolbar_forms.html" %}
 
     <!-- set-up right panel tabs -->
-    {% include "webclient/data/includes/right_plugin.general.js.html" %}
+
+    <!-- TEMP pasting script here instead of 'include' to try bug-fix include failure -->
+<script type="text/javascript">
+    /**
+     * This script is included in the main containers.html page as well as public.html,
+     * adding itself as a selection listener to the jsTree in each case.
+     * It loads appropriate data into the right-hand 'general' tab on selection changes in the jsTree.
+    **/
+    
+    $(function () {
+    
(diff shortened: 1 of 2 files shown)


joshmoore:
Tag creation _does_ work today on merge-ci/web. Does that point to a solution that doesn't require the TEMP copying of the script?

will-moore:
I still don't know why the ```include``` failed before on merge-ci, but it has nothing to do with the tagging fix in https://github.com/ome/omero-web/pull/50/.

will-moore:
As discussed, since the include is only used in one place (apart from old unused front page) we don't need to duplicate code if we simply in-line the include code (as committed above).
So can clean-up for merge...

will-moore:
Update - As Simon suggested, this is likely due to mapr right panel not loading (it overwrites the included template). This would explain why this is failing on merge-ci but not locally or on py3-ci.
Also, it highlights that in-lining the include as suggested above, would break mapr!
Closing...

will-moore:
Just confirmed that merge-ci (python 2.7 and Django 1.11) can't run the current release of mapr. Fails with:

```
File "/Users/willadmin/Desktop/WEB/omero-web/omeroweb/urls.py", line 85, in <module>
    __import__(urlmodule)
  File "/Users/willadmin/Desktop/MAPR/omero-mapr/omero_mapr/urls.py", line 23, in <module>
    from . import views
  File "/Users/willadmin/Desktop/MAPR/omero-mapr/omero_mapr/views.py", line 62, in <module>
    from omeroweb.http import HttpJPEGResponse
ImportError: No module named http
```

But, since mapr is installed there, the mapr right_plugin template overwrites the webclient one, but fails to render (imports fail etc) so we get no right panel.
This is fixed in https://github.com/ome/omero-mapr/pull/54

PR #10, opened 2019-09-02
unit tests

will-moore:
This is on top of @joshmoore's ```remove integration tests``` branch.

Work in progress.
At current state, we have

```
$ python -m pytest
=================================================================================================== test session starts ===================================================================================================
platform darwin -- Python 2.7.16, pytest-4.6.5, py-1.8.0, pluggy-0.12.0
rootdir: /Users/willadmin/Desktop/WEB/omero-web
collected 91 items                                                                                                                                                                                                        

test/unit/test_marshal.py .............                                                                                                                                                                             [ 14%]
test/unit/test_render_response.py FFFFFFFFFFFFF                                                                                                                                                                     [ 28%]
test/unit/test_settings.py ...                                                                                                                                                                                      [ 31%]
test/unit/test_show.py ...................                                                                                                                                                                          [ 52%]
test/unit/test_tree.py .........                                                                                                                                                                                    [ 62%]
test/unit/test_util.py ..FFFFFF............                                                                                                                                                                         [ 84%]
test/unit/test_webgateway.py ...........F..   
```

PR #57, opened 2019-11-06
Enable py36 unit tests and fix failures

joshmoore:
This got turned off some time ago (gh-10) and was never subsequently turned on. 😱 Fortunately, the failures were fairly minimal.


joshmoore:
@will-moore : you ok with the new method here?

will-moore:
Looks good 👍

PR #38, opened 2019-11-01
fix api test_container_crud

will-moore:
Fixes https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/31/testReport/OmeroWeb.test.integration.test_api_containers/TestContainers/test_container_crud_Project_/

Needs omero-marshal fix too...

PR #59, opened 2019-11-07
Tox improvements (flake8) and cleanup

sbesson:
While reviewing https://github.com/ome/omero-web/pull/38, I was surprised that the missing imports were not picked up via a minimal check. Executing the `flake8` on Python 3 suffices to flag additional `long` and `unicode` imports which are probably not covered by integration tests.

This PR makes the following changes:

- adds `flake8` to the tox environment defined by `tox.ini` for execution under `py27` and `py36`
-  fixes a few Python 3 flake8 errors by importing the relevant `past.builtins` methods
- cleans up dependencies declared in `tox.ini` redundant with `setup.py`
- uncaps Travis CI to run on Python 3 and removes flake8 (now executed via `tox` under multiple Python environment)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,9 +9,7 @@ requires = pip >= 19.0.0
 deps =
     flake8==2.4.0
     future
-    Django>=1.8,<1.9
     numpy>=1.9
-    Pillow
     pytest
     PyYAML
     py27: tables < 3.6.0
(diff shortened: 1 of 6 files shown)


sbesson:
That's related to the fact that `tox.ini` does not configure `flake8` at the moment. I will look into an extra commit  to enable it. That way `flake8` warnings can be flagged during local development rather than on Travis CI only.

sbesson:
Last commits migrate the `flake8` logic out of `.travis.yml` and into `tox.ini` and perform various cleanups of redundant dependencies. Description updated.

tox.ini
```diff
@@ -7,19 +7,16 @@ requires = pip >= 19.0.0
 
 [testenv]
 deps =
+    flake8==2.4.0
```
joshmoore: Does a higher version work on py3?

sbesson: I have an in-progress branch to uncap `flake8` (primarily revolving about `except: -> except Exception:` and similar cleanups). If we wanted to go there, I would propose to get this in first and open a follow-up PR.

joshmoore:
Merging proactively so we can get the follow-up in.

sbesson:
Opened https://github.com/ome/omero-web/pull/62 for discussion

PR #74, opened 2019-11-14
Extend sleep time for webgateway cachesettings test

sbesson:
This unit test has been proven to be intermittently failing with a Time limit failed assertion (see https://travis-ci.org/ome/omero-web/builds/611736343 for an example)

In agreement with the other unit tests, this sets the sleep time to 1 second more than the timeout before testing the cache deletion.


joshmoore:
I've run into the same error as well. :+1: for bumping the time. If it persists, I assume we bump again.

PR #78, opened 2019-11-20
Changelog

will-moore:
On top of https://github.com/ome/omero-web/pull/31

Do we want to remove the various empty ```.dev``` headers instead of trying to retrospectively fill them out?


joshmoore:
Merging after a quick sign-off for posting on image.sc. Happy to have follow-ups. thanks, @will-moore & @mtbc.

Issue #65, opened 2019-11-12
Add "arbitrary extra nginx config" to `omero web config`


manics:
Previously https://github.com/ome/ansible-role-omero-web/releases/tag/2.0.4 only regenerated the nginx config when OMERO.web was upgraded, changes in configuration that weren't part of an upgrade would be ignored unless `/etc/nginx/conf.d/omero-web.conf` was manually deleted.

https://github.com/ome/ansible-role-omero-web/pull/26/commits/cd4ded47eef7c9afc5c82e94c74e573cb2d44cd4 in https://github.com/ome/ansible-role-omero-web/pull/26 fixes this and ensures `omero-web.conf` is updated when necessary. The downside is that playbooks that relied on the previously incorrect behaviour to be able to insert extra text will now fail an idempotence check:
https://github.com/ome/prod-playbooks/blob/142d8c675c56d465686e792eb6d69668e408e537/omero/training-server/playbook.yml#L82-L89

PR #80, opened 2019-11-21
Add omero.web.nginx_server_extra_config

manics:
Closes https://github.com/ome/omero-web/issues/65

Testing:
```
omero config set omero.web.nginx_server_extra_config '["listen 443 ssl;", "ssl_certificate /dummy/fullchain.pem;", "ssl_certificate_key /dummy/private.key;"]'
omero web config nginx
```
Generated configuration should include
```
listen 443 ssl;
ssl_certificate /dummy/fullchain.pem;
ssl_certificate_key /dummy/private.key;
```

### DIFF
--- a/omeroweb/templates/nginx.conf.template
+++ b/omeroweb/templates/nginx.conf.template
@@ -9,6 +9,8 @@ server {
     sendfile on;
     client_max_body_size %(MAX_BODY_SIZE)s;
 
+%(NGINX_SERVER_EXTRA_CONFIG)s
+
     # maintenance page serve from here
     location @maintenance%(PREFIX_NAME)s {
         root %(ROOT)s/etc/templates/error;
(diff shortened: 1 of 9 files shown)


omeroweb/templates/nginx.conf.template
```diff
@@ -9,6 +9,8 @@ server {
     sendfile on;
     client_max_body_size %(MAX_BODY_SIZE)s;
 
+%(NGINX_SERVER_EXTRA_CONFIG)s
```
manics: This lack of indentation is deliberate since if there are multiple lines only the first will be indented, and it's better to be consistent and require the user to expicitly indent if they need it.

manics:
For example this will replace:
https://github.com/ome/prod-playbooks/blob/7d8e59ced0b47a224d6be02907d913c392ef9379/nightshade-webclients.yml#L92-L99

joshmoore:
Merging. Only potential follow-up I could think of is prepending a comment to say where the values come from, which might hold for a number of different variable sources.

manics:
@joshmoore Is this what you had in mind? https://github.com/ome/omero-web/pull/90

PR #82, opened 2019-11-21
omero-py>=5.6.dev5

manics:
Bump omero-py to https://github.com/ome/omero-py/releases/tag/v5.6.dev5


joshmoore:
I debated this but wasn't sure. Is there a dependency? or is it just more a matter of best practice?

manics:
It'll make things a bit easier if anyone pip installs omero-web, with all the bug fixes I assume we'd want them to update the whole stack for testing.

joshmoore:
Fair enough merging. I'll try to get in the habit.

Issue #75, opened 2019-11-15
load_server_settings() fails to open config


mtbc:
Servers like `ns-web` log plenty of errors like,
```
2019-11-15 10:05:21,320 ERROR [                     omeroweb.decorators] (proc.21412) load_server_settings():264 Traceback (most recent call last):
  File "/opt/omero/web/OMERO.web/lib/python/omeroweb/decorators.py", line 261, in load_server_settings
    propertiesToDict(conn.getClientSettings(),
  File "/opt/omero/web/OMERO.web/lib/python/omeroweb/webclient/webclient_gateway.py", line 250, in getClientSettings
    localcfg = AdminControl().open_config()
  File "/opt/omero/web/OMERO.web/lib/python/omero/plugins/admin.py", line 1686, in open_config
    self.ctx.die(577, str(e))
  File "/opt/omero/web/OMERO.web/lib/python/omero/cli.py", line 592, in die
    raise Exception((rc, args))
Exception: (577, "(1, 'Resource temporarily unavailable')")
```

mtbc:
@sbesson recalls that the problem may be related to concurrent access to the locked `config.xml` client-side.

PR #83, opened 2019-11-21
Use config.xml cached at startup in Django settings

manics:
Closes https://github.com/ome/omero-web/issues/75

A nicer alternative to https://github.com/openmicroscopy/openmicroscopy/pull/6053

Testing: on omero.web running separately from omero.server.
```
omero config set omero.client.ui.tree.orphans.name whatever
```

Login as the right kind of user (non-admin, non-group-leader) and `Orphaned images` should be called `whatever`


joshmoore:
This would be a nice one to get out to nightshade to silence check_mk!

PR #85, opened 2019-11-25
If OMERO_HOME not found, use omero_userdir()

will-moore:
To be used with corresponding change in ```omero-py```: https://github.com/ome/omero-py/pull/14

See testing instructions there.


will-moore:
This shouldn't be needed anymore, since we now enforce the setting of ```$OMERODIR``` in https://github.com/ome/omero-py/pull/14

PR #87, opened 2019-11-26
Fix script results failures

will-moore:
This prevents multiple timeouts running at the same time, which allows several simultaneous calls
to get script results, causing race conditions.

See https://forum.image.sc/t/omero-scripts-producing-failed-to-get-results/31601

To test:
 - Run multiple scripts at once. E.g. Open lots of "Batch Image Export" script dialogs and then hit Run on all of them at once.
 - Should get all results returned as expected. No "Failed to get results" as described on forum.
 - Looking at the console Network tab, you should see that we only ping the server once every 5 secs. Previously, each script started it's own 5-second ping. So we were had multiple simultaneous calls to getResults(). Only the first call succeeded, but the Django ```request.session``` containing the results was over-written by the other request. E.g. as described at https://stackoverflow.com/questions/13748166/django-session-race-condition

### DIFF
--- a/omeroweb/webclient/templates/webclient/base/includes/activities_info.html
+++ b/omeroweb/webclient/templates/webclient/base/includes/activities_info.html
@@ -32,7 +32,8 @@
     }
 
     // this is called by the setInterval loop below
-    OME.activitiesUpdate = function(i) {
+    OME.activitiesUpdate = function() {
+        var i = OME.activitiesInteval;
         $.get("{% url 'activities' %}", function(data) {
             $('#activities_spinner').hide();
             var inprogress = $("#inprogress", data).text();
@@ -42,12 +43,16 @@
             // if we've got no jobs still running, stop checking
             if ((typeof inprogress == 'undefined') || (inprogress.length == 0)) {
                 if (i) clearInterval(i);
+                OME.activitiesInteval = undefined;
                 showJobCount(0);
                 //return;
             }
             inprogress = parseInt( inprogress );
             if (inprogress==0) {
-                if (i != undefined) clearInterval(i);
+                if (i != undefined) {
+                    clearInterval(i);
+                    OME.activitiesInteval = undefined;
+                }
             }
             
             // display what we recieved, bind events etc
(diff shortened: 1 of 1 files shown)


austinMLB:
I patched this into my version and could no longer reproduce the error I was seeing.  I noticed that "activitiesInteval" should perhaps be "activitiesInterval" (missing 'r'), but I don't know the code well enough to comment on correctness.  --Michael (original poster on the forum)

will-moore:
Thanks for the feedback @austinMLB. I fixed that typo and removed a stray ```console.log()```.

pwalczysko:
Repeating the problem on ome-training-4
Needs to select all the idr0021 datasets (10) and run 12 "simultaneous" as possible runs of Batch image export on all the 10 datasets.

Got one failure
Screen Shot 2019-11-27 at 11 56 45

pwalczysko:
Unfortunately, on py3-ci, I am still getting the failures as well.

Screen Shot 2019-11-27 at 12 08 36

will-moore:
Apologies @pwalczysko, it appears that *again* this PR wasn't included. Seems to be happening a lot these days.
Snoopy says:
```
Excluded PRs:
  - PR 87 will-moore 'Fix script results failures' (status: )
```
at https://github.com/snoopycrimecop/omero-web/commit/82bf95a93fccbca07ac8974f2451fad4b4d6b87f
and indeed, this PR is not X but is not green either.

pwalczysko:
This PR seems to be excluded again on https://github.com/snoopycrimecop/omero-web/commit/82bf95a93fccbca07ac8974f2451fad4b4d6b87f ? I cannot see any apparent reason reported there for it.

joshmoore:
@pwalczysko : likely caused by a delayed job on py3-ci. Should be there now.

pwalczysko:
Repeated on py3-ci just now.

Batch image export run in close succession 18 times on the ca 400 images (executed on 10 datasets) of the idr0021 project.
No reported problems in the activities, all ended as expected.

The zips;were also created and attached as expected.
Clearing of tha Activities table went fine too.

Ready to merge FMPOV

PR #92, opened 2019-11-27
Fix roi masks and thumbnails

### LINKED ISSUE #89: Bug: masks not displayed with Python 3
Demo on idr-experimental failed to display masks.

will-moore:
Fixes #89.

To test:
 - Check that Masks are loaded in iViewer AND old viewer.
 - Also check that ROI thumbnails are shown in old viewer (noticed this was broken while fixing the masks)

### DIFF
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -755,7 +755,7 @@ def render_shape_mask(request, shapeId, conn=None, **kwargs):
         if x > width - 1:
             x = 0
             y += 1
-    rv = StringIO()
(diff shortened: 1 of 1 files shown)


manics:
Doesn't seem to work on py3-ci: https://py3-ci.openmicroscopy.org/web/webclient/img_detail/2966/?dataset=2269 (`user-3`), the bounding box is shown but not the mask overlay.

joshmoore:
py3 should now be updated: https://py3-ci.openmicroscopy.org/jenkins/job/Trigger/210/console

PR #94, opened 2019-12-02
Fix handling of exceptions in process_custom_settings

will-moore:
See https://github.com/ome/omero-mapr/pull/57#issuecomment-560472871

To test:
 - set some invalid JSON
 - should see a useful error message on web start (not sure where the Error printing text is coming from, but I saw the same in web-install testing earlier).

```
$ omero config set omero.web.server_list '[["some", "invalid JSON]'
$ omero web start
Error printing text
Invalid SERVER_LIST (omero.web.server_list = '[["some", "invalid JSON]'). Unterminated string starting at: line 1 column 11 (char 10). A list of servers the Web client can connect to.
```

NB: I seem to remember that ```omero config set``` would fail with invalid JSON before? cc @manics

### DIFF
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -1036,7 +1036,7 @@ def process_custom_settings(
         except ValueError as e:
             raise ValueError(
                 "Invalid %s (%s = %r). %s. %s" %
-                (global_name, key, global_value, e.message, description))
+                (global_name, key, global_value, e.args[0], description))
         except ImportError as e:
             raise ImportError(
                 "ImportError: %s. %s (%s = %r).\n%s" %


joshmoore:
Seeing the same behavior. :+1:  If you want to see where the "Error printing text" is coming from, you can turn on debugging:

```
(omero-virtualenv) bash-4.2$ bin/omero -d1 web start
Error printing text
Invalid SERVER_LIST (omero.web.server_list = '[["testintegration",14064,"testintegration"]'). Expecting object: line 1 column 44 (char 43). A list of servers the Web client can connect to.
Traceback (most recent call last):
  File "/home/omero/workspace/OMERO-test-integration/src/dist/lib/python/omero/cli.py", line 523, in safePrint
    stream.write(text)
TypeError: expected a character buffer object
```

(`DEBUG=1 bin/omero ...` would work as well)

PR #97, opened 2019-12-12
wheel

jburel:
add wheel to requirement
Following comment https://forum.image.sc/t/omero-python-3-5-6-milestone-for-upgrade-testing/32120/2
I had a closer look at the log of omeroweb-install. I could see error like invalid command 'bdist_wheel' when installing omero-web (same for omero-py) on all OS

Check that you cannot see the error in the travis log with this PR


will-moore:
No error messages like that in travis log. 👍

jburel:
Discussed today. 
Plan is to review the dependencies during the bump 
We agree that ``wheel`` is not directly a dependency of omero-py.

PR #98, opened 2019-12-16
Ignore OMERO_HOME in omeroweb/settings.py

will-moore:
See https://forum.image.sc/t/omero-python-3-5-6-milestone-for-upgrade-testing/32120/3

If ``OMERO_HOME`` is set, we should ignore it and show warning:
Same as in ``omero-py``: https://github.com/ome/omero-py/pull/148

To test, setting OMERO_HOME should not break ``omero web start``:
```
$ export OMERO_HOME=/path/to/somewhere
$ omero web start
...
```

### DIFF
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -56,13 +56,13 @@ if 'OMERO_HOME' in os.environ:
     logger.warn("OMERO_HOME usage is deprecated in OMERO.web")
 
 if 'OMERODIR' in os.environ:
-    OMERO_HOME = os.environ.get('OMERODIR')
+    OMERODIR = os.environ.get('OMERODIR')
 else:
-    OMERO_HOME = os.path.join(os.path.dirname(__file__), '..', '..', '..')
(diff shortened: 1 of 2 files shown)


omeroweb/settings.py
```diff
@@ -53,15 +53,16 @@
 # A boolean that turns on/off debug mode.
 # handler404 and handler500 works only when False
 if 'OMERO_HOME' in os.environ:
-    OMERO_HOME = os.environ.get('OMERO_HOME')
-elif 'OMERODIR' in os.environ:
-    OMERO_HOME = os.environ.get('OMERODIR')
+    logger.warn("OMERO_HOME usage is deprecated in OMERO.web")
```
manics:
```suggestion
    logger.warn("OMERO_HOME usage is ignored in OMERO.web")
```
Same comment as https://github.com/ome/omero-py/pull/148/files#r358707608

joshmoore:
Cheers. Only decision left to make is if we want to just ignore OMERO_HOME without warning to leave it to userspace.

Issue #99, opened 2019-12-18
`omero web start` should require OMERODIR


manics:
Whilst testing https://github.com/ome/omero-web/pull/98 I discovered `omero web start` works if `OMERODIR` isn't set, and creates a `var` dir inside the virtualenv `lib` directory:
`604 static files copied to '/VENV/lib/var/static', 2 post-processed.`

PR #100, opened 2019-12-18
web requires OMERODIR

manics:
closes https://github.com/ome/omero-web/issues/99


will-moore:
Works fine:
```
$ omero web start
ERROR: OMERODIR not set
```
👍

joshmoore:
Does `stop` or any of the other commands also need `OMERODIR`?

will-moore:
Possibly. If OMERODIR is not set then config may not be found, so web won't know if it's running in developer mode, debug etc.
We'd need to do a full review to know which commands really need OMERODIR, bu
unless there's a case for allowing some commands when it's not set, it's easier/safer to apply the blanket ban.

PR #109, opened 2020-01-13
Quick fix of the show downloads regex

joshmoore:
The 5.6.0 version is broken for non-dev version strings.
This should handle 5.6.0, 5.6.dev1, and 5.6.0.dev1 style
version strings. Ultimately, this method should either be
refactored out to a util method so we can write unit tests
or removed in favor of a more complete solution a la
https://github.com/ome/design/issues/103


joshmoore:
Pushed a small fix:

Screen Shot 2020-01-13 at 12 37 23

PR #111, opened 2020-01-14
Django 22

will-moore:
Work in progress...
Upgrading Django to 2.2 LTS (supported till April 2022) since current 1.11 is unsupported after April 2020.
See https://www.djangoproject.com/download/

Exclude until 5.6.0 is out...

--exclude


jburel:
do you want to include that PR now that 5.6 is out?

will-moore:
No, I think it's not ready yet. Maybe I'll close for now...

PR #114, opened 2020-01-21
Fix color picker hex input box

chris-allan:
Problem was jQuery access to input box value via .attr('value') instead of the newer .val(). The latter method sets a data attribute now and no longer accesses the input box value.

### DIFF
--- a/omeroweb/webgateway/static/webgateway/js/ome.colorbtn.js
+++ b/omeroweb/webgateway/static/webgateway/js/ome.colorbtn.js
@@ -33,7 +33,7 @@ $.fn.colorbtn = function(cfg) {
 
     var callback = function (color) {
       self.attr('data-picked-color', color);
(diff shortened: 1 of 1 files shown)


chris-allan:
@will-moore / @jburel: Can we get this merged?

jburel:
listing it for review

will-moore:
Just testing locally without this PR before looking at the fix... I don't see any bugs with the functionality? Or was this fixing some warning etc?

chris-allan:
Maybe you already have the change included or are looking at something else @will-moore? For example this is broken on IDR:

 * http://idr.openmicroscopy.org/webgateway/img_detail/9036344/?dataset=9008

Changing the hex value does nothing and there is no synchronization between the elements.

will-moore:
Fixed for me. Good to merge 👍

PR #115, opened 2020-01-21
Confine projections to no more than 256MiB of raw data

chris-allan:
# What this PR does

Projections from the OMERO.web default viewer are currently not confined. In 2019 the stack sizes across all active channels can easily be 1-2 GiB which will, on a reasonable machine, result in projection times > minute. The stack size can also be > Java maxint which will result in integer overflows and HTTP 500s.

This PR confines the maximum projection size as 256MiB of raw data, disabling the "Maximum Intensity" viewing option of the viewer if this condition is met.

# Testing this PR

1. Find an image where the projection size as defined as `sizeX * sizeY * sizeZ * activeChannelCount * bytesPerPixel` is greater than 256MiB as well as one that is not

2. Open the viewer

3. Ensure that the "Maximum Intensity" viewing option is disabled or enabled accordingly

### DIFF
--- a/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
+++ b/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
@@ -1116,7 +1116,10 @@
       // on pixel type byte width
       var bytes_per_pixel = Math.ceil(
         Math.log2(viewport.loadedImg.pixel_range[1]) / 8.0);
-      var stack_size = sizes.width * sizes.height * sizes.z * sizez.c * bytes_per_pixel;
+      var active_channel_count = viewport.loadedImg.channels
+        .filter(function(x) { return x.active }).length;
+      var stack_size = sizes.width * sizes.height * sizes.z
+                       * active_channel_count * bytes_per_pixel;
 
       // disable 'Max Intensity' projection for single-Z images or if total
       // amount of data to be projected is > 256MiB
--- a/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
+++ b/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html
@@ -1111,8 +1111,19 @@
         $("#rdef-fullrange-btn").attr("disabled", "disabled").addClass("button-disabled");
       }
 
-      // disable 'Max Intensity' projection for single-Z images
-      var disable_intmax = (viewport.loadedImg.rdefs.invertAxis || viewport.getSizes().z < 2);
+      var sizes = viewport.getSizes();
+      // Currently relies on `pixel_range` coming from the server being based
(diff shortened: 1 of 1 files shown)


jburel:
The size limit `` 256 * 1024 * 1024`` should probably be turned into a config value that other clients e.g. iviewer can access  so that they can adjust the controls accordingly too

chris-allan:
@jburel: Agreed. I would have pursued that but it's 3+ PRs to do.

Shall we get this in and I can follow up with those changes and we can decide if this should actually be enforced at the rendering engine level as well? Since we're using the microservices predominantly it's less of an issue for us and adding a check there is very easy.

jburel:
I will do some functional testing this am

jburel:
Imported an fake image ``my-special-test-file&pixelType=uint8&sizeX=3000&sizeY=3000&sizeZ=300&sizeC=4.fake``
Without this PR:
Screenshot 2020-02-14 at 10 29 28

With this PR:
I cannot click on the Maximum intensity radio button

From a "internal" viewer perspective, this works fine.

Do you want me to create an issue for the config parameter?

chris-allan:
Sure, I should be able to start the work on it next week and we can decide on when we want to get it in and how.

jburel:
Discussed with @chris-allan,
open a PR in https://github.com/ome/omero-common property ``omero.pixeldata.max_projection_bytes`` 
We can then consume that property when ready

jburel:
merging
Adjustment to be made when we have the property in place

will-moore:
idr0077 (currently importing) has some images 1920 x 1920 x 285 or more and I can see users wanting to play a movie of the projection as at https://twitter.com/karlriha/status/1227154739198033920, but this PR won't be in that release I guess.

will-moore:
Apologies for not looking closer at the code before, but I noticed that the active channel counting only happens when you first launch the viewer (not when you turn channels on/off).
So it's possible to start off below the threshold, then go over the threshold by turning on multiple channels and projection doesn't get disabled.
I noticed this because of a similar issue porting this to iviewer.

To fix it properly, we'd need to check for disabling the projection when channels turn on/off.
Also, if we're already doing projection, to turn projection off when too many channels turn on. OR we disable channel buttons when projection is on, but that would probably be harder and more confusing.
Don't know if it's worth it in this viewer? or in iviewer?

jburel:
The safest at this stage will be to check the number of channels regardless of the number of active channels
We should probably disable it deeper in the stack

chris-allan:
I can definitely open a follow up PR that switches the checks from the number of active channels to the total number of channels; that would err on the side of caution.

chris-allan:
Follow up is in #130.

/cc @jburel, @will-moore

PR #119, opened 2020-02-01
Map ann parent ID

will-moore:
I recently found I needed this:

With multiple objects e.g. Images selected, map annotations are shown for all of them. We currently display the Name of the object that each map annotation is linked to, but names can be hard to recognise. This adds the Object ID to the tooltip.
To test, select a few Objects that have some map annotations:
 - Tooltip on each map annotation should show e.g. Image ID: 1

Screenshot 2020-02-01 at 23 17 16

### DIFF
--- a/omeroweb/webgateway/static/3rdparty/jquery.jstree-3.0.8/jstree.js
+++ b/omeroweb/webgateway/static/3rdparty/jquery.jstree-3.0.8/jstree.js
@@ -7045,6 +7045,8 @@
 
 
 (function ($) {
+	// NB: document.registerElement is deprecated - see https://github.com/vakata/jstree/issues/2094
+	// But seems this is not used by OMERO.web currently - see https://github.com/ome/omero-web/pull/119
 	if(document.registerElement && Object && Object.create) {
 		var proto = Object.create(HTMLElement.prototype);
 		proto.createdCallback = function () {
--- a/omeroweb/webgateway/static/3rdparty/jquery.jstree-3.0.8/jstree.js
+++ b/omeroweb/webgateway/static/3rdparty/jquery.jstree-3.0.8/jstree.js
@@ -7066,7 +7066,7 @@
 		};
 		// proto.attributeChangedCallback = function (name, previous, value) { };
 		try {
-			document.registerElement("vakata-jstree", { prototype: proto });
+			window.customElements.define("vakata-jstree", { prototype: proto });
 		} catch(ignore) { }
 	}
 }(jQuery));
--- a/omeroweb/webclient/templates/webclient/annotations/mapanns_underscore.html
+++ b/omeroweb/webclient/templates/webclient/annotations/mapanns_underscore.html
@@ -31,6 +31,11 @@
                       <% print (ann.permissions.canEdit && clientMapAnn ? 'editing' : 'viewing') %>
                       <b><%= ann.parentNames.length %></b> identical Key-Value annotations:
                     <% } %>
+                    <% if (!ann.parentNames) { %>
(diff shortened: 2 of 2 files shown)


pwalczysko:
but you removed the name ? not sure this really improves the experience

will-moore:
No, the name is not in the tool-tip. It is above the table itself and that hasn't changed.

pwalczysko:
Aha, I did not get that. Removed my previous comment.

pwalczysko:
Yes, understand now. So the situation is:

  - have a MapAnn which is on one or more images, Datasets, Projects, Plates, Screens
  - select more of the objects of the same quality (such as having a MapAnn on one Image (image1), you select image1 and image2 as well)
  - in the RHP, see that the tooltip on that MapAnn gives you not only the Annotation ID, but also the ImageID (the ID of image1 from the above example)

This is working fine on merge-ci and makes sense as an improvement

Ready to merge fmpov

omeroweb/webgateway/static/3rdparty/jquery.jstree-3.0.8/jstree.js
```diff
@@ -7066,7 +7066,7 @@
 		};
 		// proto.attributeChangedCallback = function (name, previous, value) { };
 		try {
-			document.registerElement("vakata-jstree", { prototype: proto });
+			window.customElements.define("vakata-jstree", { prototype: proto });
```
joshmoore: Are there any other changes to jquery.jstree that we should consider? i.e. does this need bumping?

will-moore: I don't think it *needs* bumping or that we are currently looking to add features that require newer releases. So not high priority just now, but could look into it if anything changes...

chris-allan: https://www.chromestatus.com/features/4642138092470272 which was removed in M80. 

Reference post:

 * https://developers.google.com/web/updates/2019/07/web-components-time-to-upgrade

M80 was released on 4 Feb so currently this functionality likely silently fails or throws hard JavaScript errors. There is at least one other location in jstree where `registerElement` is called and it is also called within panojs. We should most definitely look at a jstree, and possible a panojs, upgrade and/or full patching.

/cc @knabar, @jburel

chris-allan: jstree issue for further reference:

 * https://github.com/vakata/jstree/issues/2094

jburel: line 7084 is also more likely to create issue
`` if(document.registerElement ``

will-moore: @jburel That was also pointed out on the issue that Chris linked: https://github.com/vakata/jstree/issues/2094#issuecomment-433926602
But the issue hasn't been re-opened. I tested, removing that whole block of code and it doesn't break any jsTree functionality. It creates a ```createdCallback()``` function but I can't see that function called anywhere or find it in the docs.

jburel: In that case, maybe put a note so we know we can remove it later on

will-moore: Done

jburel:
Thanks for the note

jburel:
This is an improvement
Further work will probably be required.

PR #123, opened 2020-02-10
omero.web.root_application: override / handler

manics:
Allows the root handler in OMERO.web (`/`) to be overridden.
The default is for `/` to redirect to `/webclient`. This can be used to make omero-gallery appear as `/`, and should mean the IDR no-longer needs a custom patch https://github.com/IDR/deployment/pull/211/commits/60f2161ac7f6882be826788bc3b59b6ddc285e00
```
omero config set omero.web.root_application gallery
```


will-moore:
The code looks fine and works for me locally.
Good to merge.

PR #124, opened 2020-02-11
Delete check parent links

will-moore:
See https://forum.image.sc/t/caution-with-copy-links-in-omero/33680

We do a simple check for multiple parents and if found, show a Warning message:

Screenshot 2020-02-11 at 15 25 19

To test:
 - Select 1 or more Images or Datasets or Plates where at least one of the selected objects has multiple Parents.
 - Click Delete.
 - Should see the warning above.
 - Also check warning is not shown when not needed.

### DIFF
--- a/omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
+++ b/omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
@@ -520,6 +520,29 @@ OME.truncateNames = (function(){
     return truncateNames;
 }());
 
+OME.checkForMultipleParents = function(objList, callback) {
+    // objList is list of ['image=1'] etc.
+    var url = WEBCLIENT.URLS.api_parent_links + '?' + objList.join('&');
+
+    $.getJSON(url, function(data){
+        // look for any child that has > 1 parent:
+        var multiple = false;
+        var links = {};
+        data.data.forEach(function(link){
+            var childId = link.child.type + link.child.id;
+            var parentId = link.parent.type + link.parent.id;
+            // do we already have a parent link for that child?
+            if (links[childId]) {
+                multiple = true;
+            }
+            links[childId] = parentId;
+        });
+        if (callback) {
+            callback(multiple);
+        }
+    });
+}
+
 // Handle deletion of selected objects in jsTree in containers.html
 OME.handleDelete = function(deleteUrl, filesetCheckUrl, userId) {
     var datatree = $.jstree.reference($('#dataTree'));
@@ -586,6 +609,17 @@ OME.handleDelete = function(deleteUrl, filesetCheckUrl, userId) {
         datatree.disable_node(node);
     });
 
+    // Hide warning, show it if e.g Image is in multiple groups
+    // https://forum.image.sc/t/caution-with-copy-links-in-omero/33680
+    $("#deleteCopyWarning").hide();
+    $("#delete-dialog-form").css('height', '92px');
+    OME.checkForMultipleParents(ajax_data, function(multiple){
+        if (multiple) {
+            $("#deleteCopyWarning").show();
+            $("#delete-dialog-form").css('height', '192px');
+        }
+    })
+
     if (notOwned) {
         $("#deleteOthersWarning").show();
     } else {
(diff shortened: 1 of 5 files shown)


joshmoore:
@dominikl, @jburel : how hard would it be to have a similar dialog in insight?

jburel:
We already have a confirmation dialog so it will be a matter of adding a check and add the warning message.
Insight has already a warning message (divergence between web/insight)
Screenshot 2020-02-12 at 08 28 10

omeroweb/webclient/templates/webclient/data/containers.html
```diff
@@ -895,7 +895,16 @@
     <p id="deleteOthersWarning" class='error' style="font-size: 120%; font-weight: bold">
         Warning: Some objects you selected are owned by other users.
     </p>
-    <p>Are you sure you want to delete the selected <span id="delete_type">Images</span>?</p>
+    <p id="deleteCopyWarning" class='error' style="font-size: 120%; font-weight: bold">
+        Warning: One or more <span class="delete_type">Images</span> are linked in multiple
```
jburel: should the text be ``linked to``?

jburel: The text will be true if you are dealing with Dataset linked to Project
Plate to screen so the message should be adjusted accordingly

omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
```diff
@@ -520,6 +520,29 @@ OME.truncateNames = (function(){
...
+OME.checkForMultipleParents = function(objList, callback) {
+    // objList is list of ['image=1'] etc.
+    var url = WEBCLIENT.URLS.api_parent_links + '?' + objList.join('&');
+
+    $.getJSON(url, function(data){
+        // look for any child that has > 1 parent:
+        var multiple = false;
+        var links = {};
+        data.data.forEach(function(link){
+            var childId = link.child.type + link.child.id;
```
jburel: if I understand the dialog correctly as soon as one image has more than one parent the warning is displayed. So the loop could be stop as soon as multiple is true

will-moore: If I want to break out of the loop early then I have to switch to using a ```for loop```, which is slightly less nice coding style (I prefer the 'functional' style). I don't think you will notice much performance improvement to break early but can refactor if you want?

omeroweb/webclient/templates/webclient/data/containers.html
will-moore: Try deleting a Dataset linked to Project and see how it looks. I see:

Screenshot 2020-02-12 at 10 43 41

omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
jburel: i understand but we have seen cases where people were trying to delete a large number of objects from the UI
so it will be good to see if it makes a major change in performance.

will-moore: This code is running after we make a call to the server, which takes e.g. 162 millisecs for 46 images (just tried that). I imagine the refactor will save maybe a few millisecs and won't be noticed compared to the time to load the data. With a large number of images the effect will be even smaller since it will take longer to load the data.
If you want me to spend time to analyse the performance under various conditions (for loop vv forEach and many images where the first has multiple parents vv many images where none of the images have multiple parents (so we don't break early) then I can do, but I think that would not be the best use of time.

jburel: Thanks for checking 

> If you want me to spend time to analyse the performance under various conditions (for loop vv forEach and many images where the first has multiple parents vv many images where none of the images have multiple parents (so we don't break early) then I can do, but I think that would not be the best use of time.

This is not what I was suggesting. Nonetheless we need to keep in mind performance and large volume since it is the reality! and break early if required.
For this call, it seems that we might not have a significant performance degradation.

jburel:
Actually insight has already a check in place for images
Screenshot 2020-02-12 at 11 24 00

jburel:
Unfortunately this will partially prevent the lost of image data.
Follow is a use case that will delete the image. The user is very likely to delete a dataset.

* Add an image to 2 datasets, d1 and d2. 
* d1 is only in one project or orphan
* Select to delete d1.
* The image will be also deleted and "disappear" from d2

will-moore:
@jburel If you delete a Dataset, any Images in it that are also in another Dataset won't get deleted.

jburel:
I got mislead by the UI, you are correct
It is deleting from the display but a refresh "brings" it back
I forgot about that strange behaviour

jburel:
The UI bug should be fixed

will-moore:
@jburel That last commit should fix the bug you noticed.

jburel:
Thanks @will-moore , the shared image is not removed from the second dataset when the first dataset is deleted.

PR #125, opened 2020-02-11
fix webadmin search firefox

will-moore:
Fixes layout of the search field on Groups and Experimenters webadmin pages on firefox.
To test:
 - Search field should not be obscured behind the edge of the table header.


chris-allan:
@jburel: We've also seen this so would be nice to get in, ready to merge?

jburel:
PR approved. Merging

PR #128, opened 2020-02-14
script GUI file upload

will-moore:
Ported from https://github.com/ome/openmicroscopy/pull/6096

See https://github.com/ome/openmicroscopy/issues/6095

Currently, some scripts that require a File use the convention of a parameter named ```File_Annotation``` to allow the user to choose a previously created ```FileAnnotation ID```.

This PR adds a File chooser input to parameter named ```File_Annotation``` which allows users to pick a local file from the Run-Script dialog in webclient. This will be used to create a File Annotation in OMERO and the ID will be passed to the ```File_Annotation``` script parameter.

To test:
 - Choose a script that has a ```File_Annotation``` parameter, such as "Populate Metadata" (https://github.com/ome/scripts/blob/develop/omero/import_scripts/Populate_Metadata.py)
 - The script dialog will allow you to choose a file:

Screen Shot 2019-08-02 at 23 53 57

 - This will allow you to pick a local File to upload
 - When the script is run, a new ```FileAnnotation``` is created in OMERO (not linked to anything) and the ID is passed to the ```File_Annotation``` script parameter (so the script should work as expected)

### DIFF
--- a/omeroweb/webclient/templates/webclient/scripts/include_param.html
+++ b/omeroweb/webclient/templates/webclient/scripts/include_param.html
@@ -90,6 +90,12 @@
 
 {% if i.min %}<span style="color: #666"> Min: {{ i.min }} </span>{% endif %}
 {% if i.max %}<span style="color: #666"> Max: {{ i.max }} </span>{% endif %}
+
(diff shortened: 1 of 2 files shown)


pwalczysko:
This works as described in the header of this PR.

Ready to merge FMPOV

PR #130, opened 2020-02-18
Use sizeC in stack size to be as safe as possible

chris-allan:
The disabling of the projection controls happens once per viewport
initialization.  If the viewport is initialized with a small number of
active channels but a the image has a large number of total channels an
unfavourable projection environment may exist.  It is safest to make the
availability check based on the total number of channels the image has.


jburel:
Works as expected.
Merging

PR #137, opened 2020-02-25
Fix shape_thumbnail using integer division

will-moore:
Fixes the rendering of shape thumbnails in old viewer.

This was previously failing when using python 2 division in python 3 gives floats not int:
```
File "/Users/wmoore/Desktop/WEB/omero-web/omeroweb/webgateway/views.py" in render_shape_thumbnail
  483.     return get_shape_thumbnail(request, conn, image, shape, compress_quality)

File "/Users/wmoore/Desktop/WEB/omero-web/omeroweb/webgateway/views.py" in get_shape_thumbnail
  633.         xs_image = Image.new('RGB', (xs_w, xs_h), bg_color)

File "/Users/wmoore/opt/miniconda3/envs/omeroweb/lib/python3.6/site-packages/PIL/Image.py" in new
  2544.     return im._new(core.fill(mode, size, color))

Exception Type: TypeError at /webgateway/render_shape_thumbnail/0175127/
Exception Value: integer argument expected, got float
```

To test, view shape thumbnail in old viewer.

### DIFF
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -563,15 +563,15 @@ def get_shape_thumbnail(request, conn, image, s, compress_quality):
     # we want to render a region larger than the bounding box
     x, y, w, h = bBox
     # make the aspect ratio (w/h) = 3/2
-    requiredWidth = max(w, h*3/2)
-    requiredHeight = requiredWidth*2/3
+    requiredWidth = max(w, h*3//2)
+    requiredHeight = requiredWidth*2//3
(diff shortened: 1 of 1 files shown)


jburel:
Checked yesterday without this PR and the preview of the shapes did not work
Checked today and it works fine with this PR.
Shapes created by iviewer and not correctly displayed in the old viewer.
I will create an issue. This is a problem since users have requested a read-only view of the rois in the preview tab of the datamanager

Screenshot 2020-02-26 at 09 04 54

PR #149, opened 2020-03-12
Chown webclient

will-moore:
Adds ```chown``` functionality to webclient.

Screenshot 2020-03-12 at 11 23 57

To test:
 - Log-in as a Group Owner or Admin that has 'Chown' privilege (other users should have chown option disabled and tooltip to explain why)
 - Choose P/D/I or S/P and use right-click menu "Change Owner..."
 - Dialog should list other members of the group (not the owner of the data if all the selected objects belong to the same user)
 - Also, dry-run will run immediately (as long as there is at least 1 other member of the group)
 - Dry-run has a cancel button, which attempts to cancel - but not sure this really prevents much work on the server.
 - Dry-run results should show what won't be chowned.
 - Choose new Owner and click OK.
 - Activities dialog will show when Done.
 - The selected object(s) and any child objects should all now belong to the new owner

NB: Some links can be lost when chown.
E.g. As a group Owner, I put MY Image into another user's Dataset. Then I Change owner of the Dataset to ME. The Image will get removed from the Dataset (because it didn't belong to the same owner as the Dataset?).


will-moore:
Looks like those last 3 commits are causing merge conflict: https://github.com/ome/omero-web/pull/180#issuecomment-670798616

pwalczysko:
Screen Shot 2020-08-11 at 11 36 08


It is not possible for a full admin to edit a group atm on merge-ci. See above. Not sure  whether or not this is caused by the PR here.

pwalczysko:
Deleted the comment about the Edit group fail - it was caused by using a non-privileged admin.

pwalczysko:
The functionality principally works as expected for 

  - full admin
  - light admin with chown
  - group owner

I would suggest though a re-think of the usage of the activities report. This feature should be handled very similarly to ``chgrp`` feature imho, with:

   - a separate dialog
   - a warning, meaning:
     - first check what will be deleted by the chown (such as links between objects, unlinked attachments etc.)
     - give user a list of those to-be-lost links
     - give user a hint that almost inevitably the objects will land in another users portfolio
     - still keep the "navigate to the data button"
     - think through the situation where the image goes into a new owner portfolio and lands in Orphaned images, I guess some sort of hint to that should be there too

This might sound tedious, but the direct experience of admin users in the field with the chgrp feature shows that they expect a high level of explanation.

pwalczysko:
Screen Shot 2020-08-14 at 15 05 44

Could that possibly say "Change owner of" ? ``Move`` is not precise enough here

pwalczysko:
There is no report about annotations ? I have for example tested a following setup:

  - 1 Dataset to it linked
  - 1 Image
 
Annotations (always two, one on Dataset, one on Image, each annotation is separate and unique to Image and Dataset):
  - KVP
  - FileAttachment
  - Tag
All attachments, the Dataset, Image and the links belong to the same user (user-3) in a read-only group.
After chowning of the Dataset to user-9, I get following:
  - no links were broken, the whole tree is cohesive, all attachments, tags and KVPs are there (this is great for the workflow, big plus imho)
  - some attachments changed owner, others did not
      - KVP changed owner (a funny side-effect is that the tooltip now claims "Added by user-9", although it is not true, it was added by user-3 and then chowned to user-9)
      - Tag did not change owner (probably it is used somewhere else, and/or tags never change owner, @mtbc  ?
      - FileAttachment did not change owner (it again could have been used somewhere else, or FileAttachments never change owner, @mtbc ?
 
I think for the workflow this is not bad as a result, but see further comments below.

pwalczysko:
Does the report take annotations into account ? It seems that no. Why ? In analogy with a similar dialog in webclient where the Move into Group action is executed, there are some informations missing, like the report about non-included annotations, see below a screenshot from a ``Move to Group``

Screen Shot 2020-08-14 at 15 23 22

Why is that,  I would think it would be great to have the annotations report for the chown case as well.

pwalczysko:
There is an element of a potential unpleasant surprise when I change ownership of a linked Dataset or Image, and after that change, the Dataset is not linked anymore to the Project and Image not anymore linked to the Dataset. Thsi is not possible to figure out by a simple user and principally it could be deemed to be a loss of data (if taken strictly). Whereas during the Move to Group there is a justification that the meaning of the word ``Move`` gives the user a hint about what will happen with the linkage of the object, with ``Change owner`` there is not such warning from a plain common sense. Unless you have a very deep knowledge of OMERO, you would never figure that out that the tree "must split".

I think we should somehow warn about this in the ``Change owner`` dialog. Although there is a message in the Activities, saying ``Dataset moved to Owner user-9``, this is a post-fact message, and it is too late to repair the link breakage - even if the reverse chown is executed, the link is lost.

will-moore:
@pwalczysko It seems that `chown` never unlinks Annotations (it may or may not change their owner, but the links aren't broken). So Annotations are not reported in the "Not included" output.

I guess this might not be what you'd expect.
I would probably expect that NO links get broken (such as P/D/I links) since there seems no need for it, except possibly in a private group. If that were the case, then the "Not included" would refer to the objects where ownership doesn't change.

But it seems that what we have is:
 - Ownership not changing is not reported as "Not included"
 - Quite a strict graph traversal of P/D/I is used to decide what can be `chown`ed, and items outside that graph are unlinked (even if they belong to the target new owner)?!

I wonder if the webclient could/should try to "repair" the broken P/D/I links after the move (if group permissions allow)?
It already does some linking of objects after `chgrp`, so the same strategy could work.
Then, we wouldn't need to unlink any objects??
However, the new links may not have same owner as the old links (if a Group owner is doing the `chown`, the new links would belong to the Group owner - not possible for the data owners to break them)!
Maybe needs some thought.

pwalczysko:
> It already does some linking of objects after chgrp,

Do you mean it links the objects to the target containers selected in the Move to Group dialog by the user ? Such as when create a New Project to move data into ?

Or do you mean something else (or this and yet something else) ?

will-moore:
>Do you mean it links the objects to the target containers selected in the Move to Group dialog by the user ? Such as when create a New Project to move data into ?

Yes, exactly.

I guess it would be nicer if the server didn't break links that didn't have to be broken in the first place.
As @joshmoore said - Having different client vv server behaviour is not ideal.
@mtbc How hard is it to change the `chown` behaviour in the server to not unlink P/D/I containers (except for private / read-only? groups)?

I don't think it should be necessary to "Select target container". We just try to keep/repair the existing links.

mtbc:
> `chown` never unlinks Annotations

It should if, for instance, you are giving your image that is tagged with somebody else's tag?

> How hard is it to change the `chown` behaviour in the server to not unlink P/D/I containers (except for private / read-only? groups)?

The rules are fairly easily changed. It should only be unlinking when necessary to avoid violating the rules of the group's permissions, though, e.g., in read-write groups I'd not expect it to unlink?

mtbc:
For read-annotate if you chown a project and you own all the datasets and images and they are not also in some other project or whatever then by default it should chown the lot.

pwalczysko:
> It should if, for instance, you are giving only some subset of images that have your tag in a read-only group?

@mtbc I just tested this case, this is not so. My guess is you are thinking in terms "that tag will not be chowned" and this is accurrate. But, the link between that tag and the image has not been cut. The transfer of ownership of that image is between two members of the same read-only group. The tag is used by the original owner on another Dataset in that group, which has not been chowned - I mean that Dataset has not been intended by the user actioning the chown action to be transferred, and it was not as expected.

mtbc:
Ah cool, thanks for testing, it probably should have been unlinked, one of the cases I missed. :smiley: I'd hope it's at least unlinked in a private group, heh.

pwalczysko:
> For read-annotate if you chown a project and you own all the datasets and images and they are not also in some other project or whatever then by default it should chown the lot.

I can confirm that from the present testing. But what @will-moore and myself are trying to explore is rather a following case:

   - Dataset belonging to user-1 in read-annotate group
   - Image belonging to user-1 in linked to that dataset in read-annotate group
   - a third party, say, light admin, chowns the image (not the Dataset) to user-2, another member of the very same read-annotate group
   - the Image gets chowned as expected, but it also gets unlinked in the process from the Dataset (which, correctly, still belongs to user-1)
   - we are pondering if it would be possible to leave that link in peace in this situation ?

Of course, at least in my opinion, this is just an exploration of possibilities, not even a crystalized opinion yet. From the perspective of a graphical client, the behaviour of the feature looks quite different than without it, so trying to find out where can we go with that.

pwalczysko:
>  it probably should have been unlinked, one of the cases I missed

We are trying to make the case that we are happy that it did not get unlinked. That unlinking makes sense only in lines of "purity" by sticking to the maxim that nothing should be there which the users could not have achieved themselves in that group type. But now, as we typically let a third party do the action (light admin, admin , GO), then this maxim makes much less sense - the third party can link whatever they like, and change ownership of links too if they choose so.

From the GUI perspective, it is a small disaster if things get unlinked unncecesarily. Maybe we can discuss this in a call next week @will-moore @mtbc  ?

mtbc:
Can group owners put someone's dataset into somebody else's project in a read-annotate group? I honestly hadn't realized. I think it's horribly confusing if group permissions are so casually violable. All sorts of these changes can be made, the system's intended to be flexible, although there are many interdependencies (and integration tests!) so expect teething troubles with unforeseen consequences. The question of what approach *should* be taken is one more for the management team, I'd say: the larger issues here are very much more what makes sense than what's technically possible.

pwalczysko:
> Can group owners put someone's dataset into somebody else's proj

Yes, they can. Freshly tested and confirmed. I was always thinking they should be able to do so too.

> although there are many interdependencies (and integration tests!) so expect teething troubles with unforeseen consequences. 

That sounds like a blocker to me.

mtbc:
> teething troubles

No worse than getting OMERO 5.1 together; the main issue is that many graph policy rules are written in the context of others. Probably good to resurrect some of the group permissions testing for the graphs operations; too easy for edge cases to creep through (like with the "Permissions Testing" list on the QA board).

If changes are limited to "let chown create mixed-ownership container hierarchies in read-annotate" then that's not too bad, the philosophical pill is harder to swallow than the technical work. If scope creeps toward "let graph operations by admins violate the permissions system willy-nilly" then, well, that's another thing. :smiley:

joshmoore:
Just adding onto the "teething troubles" thread, I imagine that this will be a case where someone other than @mtbc will need to try to adjust the existing chown rules if the current behavior is broken (or even just non-ideal).

will-moore:
It seems that this PR is exposing/clarifying more of the server-side `chown` behaviour, which in some cases is maybe not what we'd expected?
But it's tricky to work out what needs addressing for this PR (apart from https://github.com/ome/omero-web/pull/149#issuecomment-674092512 'Change Owner of' which I'll do soon) and if the server-side work needs to happen before this PR can be considered for merging?

Is the `chown` behaviour so unpredictable that we don't want to expose it in the webclient?
Or can we expose it as is, then update server to improve it (maybe before releasing webclient)?

mtbc:
The behavior's fairly predictable. When one chowns something that contains things, it'll tend to chown those too if they have the same owner and aren't multi-linked to outside what's being chowned. Then it'll make at least some obvious link breaks to preserve group permission rules. As @pwalczysko discovered, I do seem to have missed a category of link-breaking for read-only but that's easily fixed and it sounds like the opposite of where things might now go so maybe not worth effort presently, I might have much overestimated the importance of chown not being an easy way to violate permissions.

pwalczysko:
> But it's tricky to work out what needs addressing for this PR

@will-moore  I would suggest

   - https://github.com/ome/omero-web/pull/149#issuecomment-674092512 (improve the message, which is possibly a carryover from move)
   - have the user more clearly warned about the linkage breaks, see below

> Then it'll make at least some obvious link breaks to preserve group permission rules

@mtbc @will-moore What would be my priority here is : Can I get these links breakages as warnings in a dry run in the webclient ?  Or do we have to rely on a documentation somewhere else (this could of course be the guides, and we could link to them from the webclient ?)

mtbc:
Yes, dry-run reports broken links and the client can query what it is that they link.

mtbc:
(Could be that chgrp could get similar warnings.)

pwalczysko:
Of course out of scope of this PR, but 

> (Could be that chgrp could get similar warnings.)

Yes. User reports show that especially link breakages of object-annotation links are percieved as unpleasant surprises.

will-moore:
> Can I get these links breakages as warnings in a dry run in the webclient ?

This should already be working, e.g:

Screenshot 2020-08-17 at 21 56 44

will-moore:
Screenshot 2020-08-17 at 22 19 11

will-moore:
Build failed with `gcc not installed`! Close and re-open to run again...

joshmoore:
This is the pytest-xdist issue I mentioned elsewhere. Restarting won't suffice.

pwalczysko:
> This should already be working, e.g:

Studying the screenshot more closely, I realize what is missing: Links. The ``Not included`` list is enumerating Images only. This means, if I select a single image and chown it, it very probably gets a report: ``All fine, all included.`` But the link will be broken and the image will end up in Orphaned, and I will be surprised. Can we list the links too ?

will-moore:
Hmm - I don't think the server gives you the parent link, but I think that will *always* get broken when you `chown` the child, except in a `read-write` group. In which case, the webclient could simply warn you about that.

But I guess we need to decide whether OMERO should try to avoid mixed-owner links in `read-annotate` and 'read-only' groups. Currently this is allowed by the server and these can be created in webclient. But the server tries to avoid mixed-owner links when doing `chown` (except in `read-write` groups).

mtbc:
The server should give you any links it's breaking.

pwalczysko:
> But I guess we need to decide whether OMERO should try to avoid mixed-owner links in read-annotate and 'read-only' groups. Currently this is allowed by the server and these can be created in webclient

This is fine fmpov. I do not see the problem explaining the present situation. So, fmpov, lets keep it as it is now.

But we should get the warning about the link breakage. This is a surprise to a user, because we have to assume that the user is completely naive with regard to intricacies of the OMERO permissions.

will-moore:
With that last commit, we show any parent objects that are 'not included' in the `chown`.

Screenshot 2020-08-18 at 16 28 01

However, this can still be confusing (see screenshot)... You might assume that when I change the Dataset (owned by Ben) to 'Will Moore', the fact that Project is not included doesn't matter since it already belongs to 'Will Moore'. But, they will still get unlinked, even though they both end up belonging to 'Will Moore'!

manics:
--exclude
https://github.com/ome/omero-web/pull/218

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#462](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/462/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/462/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - omeroweb/webgateway/marshal.py
    - omeroweb/webclient/urls.py
    - omeroweb/webclient/views.py

--conflicts

will-moore:
Picking this up from last year...
With the changes in https://github.com/ome/omero-web/pull/149#issuecomment-675554192 it should give users a better idea what's going on ahead of time.
There is still potential for confusion, but I don't think the `chown` graph logic is going to change anytime soon, so we need to decide if the added feature benefit is worth it?

will-moore:
I think this functionality is useful to users as it stands and we should consider merging this PR.
I'll re-list for testing the display of broken links...

pwalczysko:
Retested again the behaviour on merge-ci. The warnings are quite good actually. It is quite hard to create a situation where an annotation is "not included", which is a good thing imho (this is the type of situation where the user goes into the action and then reports "...and the annotations were lost". The one case where the annotation gets "not included" is the case of annotation of multiple objects in private group, then the chown is performed on one of them only. But, even in such case, I get

Screenshot 2021-03-01 at 14 04 02

which is very nice. Also, in the activities, it tells me clearly the image is moved out of the present portfolio and moved to the portfolio of a new user, and there is a link to it.

I would vote for going ahead with this PR.

will-moore:
cc @jburel Is this "good to merge"?
Would be nice to get it in if possible?

PR #151, opened 2020-03-13
Close sessions where user is anonymous but is_valid_public_url is false

kkoz:
Previously, if a request is made where is_valid_public_url is false but is_anonymous is true, a connection would be created but would never be closed. This PR closes those connections to prevent resource leaks.

# Testing this PR

1. required setup
- Install this version of omeroweb
- You need a public user (to make `is_anonymous` True)
- Need to have `omero.web.public.url_filter` configured without `webclient/keepalive_ping` (to make `is_valid_public_url` False for `webclient/keepalive_ping` requests)

2. actions to perform
- Navigate to a public page (e.g. a pathviewer image) which uses a keepalive ping.
- Open several tabs like this.

3. expected observations
- The thread count used by omeroweb or gunicorn should not increase over time, even as keepalive requests are made by these pages.
- Without this PR, the number of threads in use (e.g. result of `ps -Fw -u omero Hh |grep gunicorn |wc -l`) will increase over time.

### DIFF
--- a/omeroweb/decorators.py
+++ b/omeroweb/decorators.py
@@ -337,6 +337,9 @@ class login_required(object):
             is_anonymous = connection.isAnonymous()
             logger.debug('Is anonymous? %s' % is_anonymous)
             if is_anonymous and not is_valid_public_url:
+                if connection.c is not None:
+                    logger.debug("Closing anonymous connection")
+                    connection.close(hard=False)
                 return None
         return connection
 


chris-allan:
Builds failing due to Teemu/pytest-sugar#187; see #152.

/cc @sbesson, @will-moore, @joshmoore

chris-allan:
Build now passing after merge of #152 and Travis rebuild.

sbesson:
--include

will-moore:
Code change makes total sense.
To test the fix as described, I need to set-up public user with the right url regex etc...

manics:
@kkoz Just out of interest, how did you find this leak? Is there anything we can do to detect these in future?

chris-allan:
@manics We found it on a customer system where this configuration was present. I can't think of any way off hand to easily detect these in the future aside from monitoring OMERO.web thread counts.

will-moore:
Instead of pathviewer I added ping every second to webgateway viewer.
Running locally (Mac) as described at https://docs.openmicroscopy.org/omero/5.6.0/developers/Web/Deployment.html#using-wsgi and with 3 viewer tabs open, I don't see any thread increases using 
```
$ pstree | grep gunicorn | wc
       7     132    1903
```
but I get "Internal Server Error" pages quite quickly without this PR.

With this PR, everything continues to work fine.
👍

mtbc:
Thank you all for your efforts here.

PR #152, opened 2020-03-14
pytest-sugar not compatible with pytest 5.4.0+

chris-allan:
See also:

 * Teemu/pytest-sugar#187
 * Teemu/pytest-sugar#188

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -14,7 +14,6 @@ deps =
     PyYAML
     py27: tables < 3.6.0
     py36: tables
-    pytest-sugar
     pytest-xdist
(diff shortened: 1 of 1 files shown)


joshmoore:
Thanks @chris-allan. The addition of `pytest-sugar` was an attempt from my side to improve the readability of the test output, but having prompt updates to pytest are certainly more important. Unless you know of other reasons to pin pytest, I'd vote for just dropping sugar.

chris-allan:
Can't think of any reason whatsoever at this point @joshmoore.

PR #154, opened 2020-03-18
public user workflow crash

will-moore:
Fixes https://trello.com/c/sBdAN4L6/375-bug-public-user-workflow-crash

Several URLs load data to the jsTree and include ```?group=ID``` to load data for the correct group. This causes ```SecurityViolation``` if you are not a member of that group.
A common cause of this is your session has timed-out, but when you try to connect, you are logged-in as public user (who is not a member of your group).

This PR checks for invalid group when loading 'containers', Datasets, Images, Plates in the jsTree and returns a 403 (as if you were logged out) which causes the webclient to refresh.

To test:
 - you need to log into webclient as non-public user, browse to a group that Public User is not a member of.
 - Wait for session to time out (e.g. close laptop to stop the webclient pinging keep-alive) or use @mtbc's handy command (from trello card) to kill it. E.g. if you can ```psql omero``` then:
```
for s in `echo "select uuid from session where useragent = 'OMERO.web' and closed is null" | psql omero | grep '[[:digit:]]'-` ; do omero sessions timeout --session $s 1 ; done
```
 - Then try to refresh the tree (refresh button) or expand Project, Dataset or Screen.
 - The webclient page should refresh to show that you are now logged-in as public user.

NB: If the public user IS a member of the group that you're in, then you won't see the refresh. Data will be loaded (if not a private group) but you'll be viewing it as pubic user, so you won't be able to edit etc. This is not ideal but has been the behaviour for a long time. It doesn't cause any errors and simply needs a manual refresh of the webclient page to 'fix'.

### DIFF
--- a/omeroweb/webclient/webclient_gateway.py
+++ b/omeroweb/webclient/webclient_gateway.py
@@ -149,8 +149,8 @@ class OmeroWebGateway(omero.gateway.BlitzGateway):
         return self._shareId
 
     def isValidGroup(self, group_id):
-        group_ids = self.getEventContext().memberOfGroups
-        return group_id == -1 or group_id in group_ids
+        ec = self.getEventContext()
+        return ec.isAdmin or group_id == -1 or group_id in ec.memberOfGroups
 
     ##############################################
     #    Session methods                         #
--- a/omeroweb/webclient/webclient_gateway.py
+++ b/omeroweb/webclient/webclient_gateway.py
@@ -148,6 +148,10 @@ class OmeroWebGateway(omero.gateway.BlitzGateway):
                 self._shareId = self.getEventContext().shareId
         return self._shareId
 
+    def isValidGroup(self, group_id):
+        group_ids = self.getEventContext().memberOfGroups
+        return group_id == -1 or group_id in group_ids
+
     ##############################################
     #    Session methods                         #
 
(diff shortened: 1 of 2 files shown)


mtbc:
Will this still be okay for admins browsing outside their groups?

will-moore:
@mtbc Good point, thanks. I'll add ```isAdmin()``` into my check.

joshmoore:
:+1: for getting this merged and released. I do wonder though if the check should be part of a decorator somewhere. If so, happy for that to happen either now or later.

will-moore:
I originally looked at the idea of putting this in a decorator, but the problem is that it relies on the ```?group=1``` in the URL query string, and we don't use that in many places.
Potentially more problematic is someone using ```?group=1``` to mean something else and getting a 403 when they didn't need to.

omeroweb/webclient/webclient_gateway.py
```diff
@@ -148,6 +148,10 @@ def getShareId(self):
                 self._shareId = self.getEventContext().shareId
         return self._shareId
 
+    def isValidGroup(self, group_id):
+        ec = self.getEventContext()
+        return ec.isAdmin or group_id == -1 or group_id in ec.memberOfGroups
```
joshmoore, line 153: Does `isAdmin` correctly handle the light admin situation?

will-moore: Yes, since we are just trying to read the data here, and all admins (and light admins) have read access.

joshmoore: :+1:

PR #162, opened 2020-04-09
Fix edit of shares

will-moore:
Fix bug with editing of shares, reported at https://www.openmicroscopy.org/qa2/qa/feedback/29218/

```
File "/opt/omero/web/venv3/lib/python3.7/site-packages/omeroweb/webclient/views.py", line 2719, in manage_action_containers
if o_type == "share" and o_id > 0:
TypeError: '>' not supported between instances of 'str' and 'int'
```

To test:
 - Select some Images and create a 'share' in the webclient
 - Under 'Public' tab, select the Share then click the 'Edit' (green pen) in the right panel
 - Should see no error - can make a change and save it.
 - /webclient/action/edit/share/ should return a 404 (not an exception)

### DIFF
--- a/omeroweb/webclient/views.py
+++ b/omeroweb/webclient/views.py
@@ -2716,6 +2716,8 @@ def manage_action_containers(request, action, o_type=None, o_id=None,
 
     elif action == 'edit':
         # form for editing Shares only
+        if o_id is None:
+            raise Http404("No share ID")
         if o_type == "share" and int(o_id) > 0:
             template = "webclient/public/share_form.html"
             manager.getMembers(o_id)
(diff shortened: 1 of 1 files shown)


joshmoore:
Is it possibly for the `o_id` to be a non-`int`? Otherwise, change makes sense to me.

will-moore:
It *could* be ```None``` if you manually edit the URL to ```http://localhost:4080/webclient/action/edit/share/```.

```url(r'^action/(?P<action>[a-zA-Z]+)/(?:(?P<o_type>[a-zA-Z]+)/)?(?:(?P<o_id>[0-9]+)/)?$',```

That URL has always caused an error with ```None > 0``` but I guess we can handle that better.

will-moore:
Fixed. Now /webclient/action/edit/share/ should return a 404 (not an exception).

joshmoore:
Thanks, @will-moore!

PR #165, opened 2020-04-26
Fix thumbnail size from config

### LINKED ISSUE #164: increase thumbnail size
Requested again at https://forum.image.sc/t/increase-thumbnail-size-in-omero/36835

Previously this was logged at
https://trello.com/c/cZ2Ww8nA/560-bug-webclient-ignores-thumbnail-size-config

Originally requested at https://github.com/ome/design/issues/96#issuecomment-394691752 and  https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=8545

Will: "Unfortunately I think this configuration is not going to work in the webclient since we changed the way that we load thumbnails.
Instead of loading individual images, we now batch load thumbnails as JSON data using webclient/get_thumbnails/?

https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webclient/urls.py#L116

but this currently has no way to pass the size to the webgateway get_thumnails_json method https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webgateway/views.py#L1541 even though that method does have a width 'w' parameter.

I've created a card https://trello.com/c/cZ2Ww8nA/16-bug-webclient-ignores-thumbnail-size-config to fix this since it was also requested at https://github.com/ome/design/issues/96#issuecomment-394691752 although the performance issues as mentioned by Sebastien will still be an issue."

will-moore:
This fixes #164 a long-standing bug: the batch JSON loading of thumbnails ignores the thumbnail size setting.

To test:

``` $ omero config set omero.client.browser.thumb_default_size 250```

Then check in browser dev tools the 'natural' size of the thumbnails in webclient.

Screenshot 2020-04-26 at 23 05 24

### DIFF
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -1566,8 +1566,11 @@ def get_thumbnails_json(request, w=None, conn=None, **kwargs):
     @param w:           Thumbnail max width. 96 by default
     @return:            http response containing base64 encoded thumbnails
     """
+    server_settings = request.session.get('server_settings', {}) \
+                                     .get('browser', {})
+    defaultSize = server_settings.get('thumb_default_size', 96)
     if w is None:
-        w = 96
+        w = defaultSize
     image_ids = get_longs(request, 'id')
     image_ids = list(set(image_ids))    # remove any duplicates
     # If we only have a single ID, simply use getThumbnail()

### CODE REVIEW

**Comment by joshmoore** [comment]
I don't see any other hard-codings of `96`:

```
/opt/omero-web $git grep 96 -- '*.py'
omeroweb/webgateway/views.py:    @param w:           Thumbnail max width. 96 by default
omeroweb/webgateway/views.py:    defaultSize = server_settings.get('thumb_default_size', 96)
omeroweb/webgateway/views.py:    @param w:           Thumbnail max width. 96 by default
omeroweb/webgateway/views.py:    @param w:           Thumbnail max width. 96 by default
omeroweb/webgateway/views.py:    defaultSize = server_settings.get('thumb_default_size', 96)
omeroweb/webgateway/views.py:    @param w:           Thumbnail max width. 96 by default
```

If more locations end up needing the look up, might be worth an internal helper.

PR #177, opened 2020-06-16
host property doc

jburel:
Move doc from ome-doc
cc @manics


omeroweb/settings.py
```diff
@@ -363,7 +363,10 @@ def leave_none_unset_int(s):
         ["APPLICATION_SERVER_HOST",
          "127.0.0.1",
          str,
-         "Upstream application host"],
+         "The front-end webserver e.g. NGINX can be set up to run on a"
```
manics, line 366: Python will concatenate these lines without a space, so you'll need to add it yourself.

jburel: done

PR #191, opened 2020-07-21
login_required() doesn't conn.close() for ConnCleaningHttpResponse

will-moore:
This allows you to use `@login_required()` decorator without `doConnectionCleanup=False`
and `conn.close()` won't get called if you return an instance of `ConnCleaningHttpResponse` (which will close the conn itself). 

If you previously needed to use `@login_required(doConnectionCleanup=False)`, but ended up returning a `Http404` then the connection wouldn't not get closed.
Now, you can use `@login_required()` and if you fail to return a `ConnCleaningHttpResponse` then the conn will get closed.

This also allows you to be more flexible and your view methods can return either a Streaming HttpResponse OR a standard HttpResponse. Previously if you wanted to return a  `StreamingHttpResponse` then you had to use `@login_required(doConnectionCleanup=False)` but if you wanted to also return a non-streaming response from the same view method then you needed to make sure you called `conn.close()` manually.

I needed this logic for another PR (coming soon on top of this commit).

Testing:
 - Need to check web logs to see when conn is closed e.g.
```
2020-08-12 09:00:15,207  INFO [                           omero.gateway] (proc.37498) close():1988 closed connection (uuid=410000b1-0e43-49d9-a911-6bf250750134)
```
 - Test download of archived files will valid Image ID and with invalid ID (to get Not Found response - actually 500) /webgateway/archived_files/download/?image=12795345345 or with no query at all /webgateway/archived_files/download/
 - Test original files at /webclient/get_original_file/398157/ with valid ID (e.g. open script dialog and click on View Script) and invalid ID - should get a custom 404 page (from `handlerInternalError()`) - check conn is closed in logs for both.

Screenshot 2020-08-12 at 09 11 48

cc @chris-allan

### DIFF
--- a/omeroweb/decorators.py
+++ b/omeroweb/decorators.py
@@ -500,10 +500,9 @@ class login_required(object):
                 )
                 if doConnectionCleanup and delayConnectionCleanup:
                     raise Exception(
-                        (
-                            "FIXME: methods that return a ConnCleaningHttpResponse"
-                            " must be marked with @login_required(doConnectionCleanup=False)"
-                        )
+                        "FIXME: methods that return a"
+                        " ConnCleaningHttpResponse must be marked with"
+                        " @login_required(doConnectionCleanup=False)"
                     )
                 doConnectionCleanup = not delayConnectionCleanup
                 logger.debug(
--- a/omeroweb/decorators.py
+++ b/omeroweb/decorators.py
@@ -491,6 +491,7 @@ class login_required(object):
 
                     # kwargs['error'] = request.GET.get('error')
                     kwargs['url'] = url
+            retval = None
             try:
                 retval = f(request, *args, **kwargs)
             finally:
--- a/omeroweb/decorators.py
+++ b/omeroweb/decorators.py
@@ -494,21 +494,20 @@ class login_required(object):
                 retval = f(request, *args, **kwargs)
             finally:
                 # If f() raised Exception, e.g. Http404() we must still cleanup
-                try:
-                    delayConnectionCleanup = isinstance(
-                        retval, ConnCleaningHttpResponse
-                    )
-                    if doConnectionCleanup and delayConnectionCleanup:
-                        raise Exception(
-                            (
-                                "FIXME: do whatever here is "
-                                "appropriate to force a developer to "
-                                "fix the issue _before_ release"
-                            )
+                delayConnectionCleanup = isinstance(
+                    retval, ConnCleaningHttpResponse
+                )
+                if doConnectionCleanup and delayConnectionCleanup:
+                    raise Exception(
+                        (
+                            "FIXME: methods that return a ConnCleaningHttpResponse"
+                            " must be marked with @login_required(doConnectionCleanup=False)"
                         )
-                    doConnectionCleanup = not delayConnectionCleanup
-                    logger.debug(
-                        'Doing connection cleanup? %s' % doConnectionCleanup)
+                    )
+                doConnectionCleanup = not delayConnectionCleanup
+                logger.debug(
+                    'Doing connection cleanup? %s' % doConnectionCleanup)
+                try:
                     if doConnectionCleanup:
                         if conn is not None and conn.c is not None:
                             conn.close(hard=False)
--- a/omeroweb/decorators.py
+++ b/omeroweb/decorators.py
@@ -96,10 +96,11 @@ class login_required(object):
     configurable by various options.
 
     doConnectionCleanup:
+        Used to indicate methods that may return ConnCleaningHttpResponse.
         If True (default), then returning a ConnCleaningHttpResponse will
         raise an Exception since cleanup is intended to be immediate; if
-        False, connection cleanup will be skipped for all
-        ConnCleaningHttpResponse values.
+        False, connection cleanup will be skipped ONLY when a
+        ConnCleaningHttpResponse is returned.
     """
 
     def __init__(self, useragent='OMERO.web', isAdmin=False,
--- a/omeroweb/decorators.py
+++ b/omeroweb/decorators.py
@@ -489,6 +489,9 @@ class login_required(object):
             finally:
                 # If f() raised Exception, e.g. Http404() we must still cleanup
                 try:
+                    if isinstance(retval, ConnCleaningHttpResponse):
+                        # conn will be closed by response.close()
+                        doConnectionCleanup = False
                     logger.debug(
                         'Doing connection cleanup? %s' % doConnectionCleanup)
                     if doConnectionCleanup:
--- a/omeroweb/decorators.py
+++ b/omeroweb/decorators.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 #
-# Copyright (C) 2011-2013 University of Dundee & Open Microscopy Environment.
+# Copyright (C) 2011-2020 University of Dundee & Open Microscopy Environment.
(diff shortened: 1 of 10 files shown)


chris-allan:
My preference is the original style.

Yes, if you have set `doConnectionCleanup` to `False` you are responsible for closing the connection on error or if you return a different response type. I don't see having to call `conn.close()` a particularly difficult thing to ask and knowing the difference is important.

will-moore:
If you wish to prevent cleanup, you can still set `doConnectionCleanup` to `False` as before and nothing will change.

All that this PR does is prevent `@login_required` calling `conn.close()` when you return an instance of `ConnCleaningHttpResponse`.
Is there ever a use-case where you'd want `conn.close()` to be called for a `ConnCleaningHttpResponse`?

It sounds easy to call `conn.close()` but we have found many cases before where we've failed to do that.
This makes it even easier to "do the right thing".

chris-allan:
I understand what the PR does and I appreciate it may make it easier to do the right thing for your use case.

However, I don't agree that hiding connection semantics behind response class type checks makes these sorts of scenarios easier to not screw up overall. It is now just as easy to have changes in your response class hierarchy cause a scenario where you do not get your expected behaviour and then encourages people to use `doConnectionCleanup=False`. Equally so if you override `ConnCleaningHttpResponse.close()`. Dealing with streaming responses that rely on an OMERO session is very tricky to get right regardless.

will-moore:
Hi, Apologies - This was a bit of a last-minute PR before I was away for 2 weeks...

Having revisited these changes, I realise that the removal of `doConnectionCleanup=False` allows
us to simplify the handling of exceptions or 404s. See commit above: https://github.com/ome/omero-web/pull/191/commits/85e33b3bc6316a4874d26dfb54b46519e85cd2c5

Previously with the use of `doConnectionCleanup=False` we needed to diligently return an instance of
`ConnCleaningHttpResponse` in every case, as in https://github.com/ome/omero-web/commit/451f959d9486ffc9e44d71e3e4aca5e6254bd437
However, in a method with many return statements, it becomes easy to miss one.
Also, this restricted what you could return, e.g. couldn't return a redirect or delegate to `return handlerInternalError()` as before.

Re https://github.com/ome/omero-web/pull/191#issuecomment-662428281: If `ConnCleaningHttpResponse.close()` is overridden and doesn't actually call `conn.close()` then
this would cause issues even without this PR.

chris-allan:
> If ConnCleaningHttpResponse.close() is overridden and doesn't actually call conn.close() then this would cause issues even without this PR.

Of course it would. My point is that there is inherent complexity here and it's not good design to pretend that there isn't. Stating `doConnectionCleanup=False` emphatically in the `login_required()` decorator declaratively states what the intent is. If you are then overriding `close()` on a class whose literal name is `ConnCleaningHttpResponse` you at least then know that `close()` was likely to have been required and looking at its implementation is a smart thing to do.

Trading one set of complexity (return values) for another (class hierarchy) is just that. One is better in certain circumstances, one is better in others. One is not objectively better universally.

If you want to discuss making connection handling easier in streaming response cases more widely there are lots of ways to approach that. Likely better ones than return values or class hierarchy checks. Django, our use of it, and our experience has evolved a lot in the 8 years since this code was added.

We are now conflating a fundamental platform and design change in this PR with real, useful, and desired functionality from #192. The changes in #192 do not require the changes from this PR in order to be implemented and should be proposed utilizing the style already established.

omeroweb/decorators.py
```diff
@@ -489,6 +495,18 @@ def wrapped(request, *args, **kwargs):
             finally:
                 # If f() raised Exception, e.g. Http404() we must still cleanup
                 try:
+                    delayConnectionCleanup = isinstance(
+                        retval, ConnCleaningHttpResponse
+                    )
+                    if doConnectionCleanup and delayConnectionCleanup:
+                        raise Exception(
+                            (
```
manics, line 503: There's an unnecessary level of parenthesis here

omeroweb/decorators.py
```diff
@@ -488,9 +495,19 @@ def wrapped(request, *args, **kwargs):
                 retval = f(request, *args, **kwargs)
             finally:
                 # If f() raised Exception, e.g. Http404() we must still cleanup
+                delayConnectionCleanup = isinstance(
+                    retval, ConnCleaningHttpResponse
+                )
+                if doConnectionCleanup and delayConnectionCleanup:
+                    raise Exception(
```
manics, line 502: How about throw an `omero.ApiUsageException`?

will-moore: Done in 085890f

will-moore:
8af5949 fixes exception handling as in https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/486/testReport/OmeroWeb.test.integration.test_links/TestLinks/test_link_datasets_images/

sbesson:
Briefly discussed during standup. It might be worth looking into whether https://github.com/ome/omero-weberror can simply be extended to simplify the testing of the new functionality/contract introduced by this PR.

will-moore:
Added tests without using weberror. See https://github.com/ome/openmicroscopy/pull/6248

omeroweb/decorators.py
```diff
@@ -484,13 +492,24 @@ def wrapped(request, *args, **kwargs):
...
             try:
                 retval = f(request, *args, **kwargs)
             finally:
                 # If f() raised Exception, e.g. Http404() we must still cleanup
+                delayConnectionCleanup = isinstance(
+                    retval, ConnCleaningHttpResponse
+                )
+                if doConnectionCleanup and delayConnectionCleanup:
+                    raise ApiUsageException(
+                        "FIXME: methods that return a"
```
joshmoore, line 505: Side note: this FIXME was largely for you, but I defer to whether or not you want to show it to users. Note that some developer tools will mark it as a warning which may lead to confusion.

will-moore: Ah, sorry - I thought it was a message to the developer to FIX their missing usage of ```@login_required(doConnectionCleanup=False)```. I'll remove the FIXME.

joshmoore:
Looks like this may well need the test fix from https://github.com/ome/devspace/pull/169

cc: @manics @sbesson

sbesson:
I suspect we'll need to review all places where `pytest-xdist` is used across our codebase indeed - https://github.com/search?q=org%3Aome+pytest-xdist&type=Code

will-moore:
Travis is failing with:
```
py36 run-test: commands[0] | flake8 .
./omeroweb/webgateway/views.py:2551:44: F821 undefined name 'StringIO'
ERROR: InvocationError for command /src/.tox/py36/bin/flake8 . (exited with code 1)
```
But that file doesn't contain `StringIO` anywhere?!
https://github.com/ome/omero-web/blob/056ad63740d4440fb454bc2100e0e3074aa33810/omeroweb/webgateway/views.py

will-moore:
Hmmm - the 'Details' link above goes to https://travis-ci.org/github/ome/omero-web/builds/719232030 which is for commit https://github.com/ome/omero-web/commit/1d0e394807683c2cb2d09765861b82604b3d7881 which isn't right.
Need to force Travis to use the correct commit!

will-moore:
So, travis seems to be merging each commit into https://github.com/ome/omero-web/commit/ba645162d7e93ee0f16551037e9d05df6165dc15 before running flake8, which is then failing.

joshmoore:
:+1:

PR #195, opened 2020-07-28
Be defensive and return default value for scripts_to_ignore

sbesson:
Reported in a QA feedback, under some condition there might be no `scripts_to_ignore` key in the `server-settings` dictionary leading to an `AttributeError`.

This change fixes this issue by returning a default value as done elsewhere in the code e.g.

https://github.com/ome/omero-web/blob/9284f6a5a6f335a1a2708c74f2ebd825142ca2ae/omeroweb/webgateway/marshal.py#L190-L191


will-moore:
Code looks good - 👍

PR #196, opened 2020-07-29
OMERO.web default log can be changed, e.g. to stdout

manics:
OMERO.web can be configured to send its default logs to stdout instead of a file. This is useful when running in the foreground, for example in Docker. The default behaviour is unchanged.

For example:
```
omero config set omero.web.default_log_handlers '["stdout"]'
omero web start --foreground
```
You can also have multiple loggers `'["default", "stdout"]'`

----

Ported from https://github.com/ome/openmicroscopy/pull/5912

### DIFF
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -133,6 +133,11 @@ LOGGING = {
             'class': 'logging.StreamHandler',
             'formatter': 'standard'
         },
+        'stdout': {
+            'level': 'DEBUG',
+            'class': 'logging.StreamHandler',
+            'formatter': 'standard'
+        },
         'mail_admins': {
             'level': 'ERROR',
             'filters': ['require_debug_false'],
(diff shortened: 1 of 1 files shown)


joshmoore:
Few things:
 - Generally looks fine.
 - slightly odd behaviors:
   - The log file is still created with `stdout`. With debug=true, the debug out is logged to OMEROweb.log.
   - With `omero.web.default_log_handlers=["default"]`, `--foreground` still seems to print quite a lot to the console. (Edit: this is with debug=true)
 - question: When  would `console` be used?
 - rfe: Especially for foreground, being able to set the logger from the CLI might be useful.
 - comment: I'm not sure when `request_handler` was dropped from the default.

manics:
I don't think any of us fully understand how the current OMERO.web logging system works 😀. I can probably deal with some of your points through trial and error? However the impression I got from the [original PR](https://github.com/ome/openmicroscopy/pull/5912) was there's concern about unintentionally changing the Django logging without fully understanding what's going on, which is why the PR was updated to minimise changes to the existing configurations, and instead only add a new config- I think this explains some of the "odd behaviour".

The other factor is that logging is already active before the settings file is parsed, and in addition logs are output during parsing, so unless we move the log configuration outside the file (not sure if that's actually possible, since the parsing will occur when the settings file is imported) there will always be logs sent somewhere.

joshmoore:
Ok. So from your text, I assume you don't have any more info on `console` or `request_handler`. In the sense of "reclaiming this file", we likely should decide whether or not console is being removed in favor of stdout.  Happy to leave the RFE for `--stdout` (or similar) for the future.

manics:
I can think of two options for taking this forward:
- Attempt to clean up the log configuration, and rely on functional testing to guard against unknown breakages. I'm happy to try this if everyones agrees with this.
- Wait for 6.0.0 when we have more leeway to make breaking changes (intentional or otherwise).

@ome/omero-apps what do you think?

joshmoore:
With this still dragging, unless there are any strong feelings I imagine it's better to get it in pre-6.0.0.

manics:
--exclude
https://github.com/ome/omero-web/pull/218

jburel:
Closing we can go back to it later

Issue #118, opened 2020-01-30
Limit download sizes (or document how to control them)


manics:
Recently the UoD production server has temporarily run out of space, presumably due to large temporary files (several GB).

Nginx error log:

`2020/01/29 15:01:10 [error] 1623#1623: *433114 upstream timed out (110: Connection timed out) while reading response header from upstream, client: X.X.X.X, server: X, request: "GET /webgateway/archived_files/download/?image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&zipname=OriginalFileDownload.zip HTTP/1.0", upstream: " [http://127.0.0.1:4080/webgateway/archived_files/download/?image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&amp:image=X&zipname=OriginalFileDownload.zip] , host: "nightshade.openmicroscopy.org", referrer: " [https://nightshade.openmicroscopy.org/webclient/download_placeholder/?ids=image-X❘image-X❘image-X❘image-X❘image-X❘image-X❘image-X❘image-X❘image-X❘image-X❘image-X❘image-X❘image-X❘image-X❘image-X❘image-X❘image-X❘image-X❘image-X❘image-X❘image-X)]`

I checked one of the images, the associated file is 360 MB.

manics:
[...]

PR #192, opened 2020-07-21
Download table as csv

will-moore:
NB: this is on top of #191. 

This allows downloading of much larger OMERO.table files as csv.

Currently, downloading of OMERO.tables is slow and will timeout with a reasonable sized table because:
 - we use table.read() to get 1 row at a time
 - we read all the rows in hand, then format string as csv and return as HttpResponse

This PR improves this by:
 - reading rows in batches of 1000 (when downloading ALL rows `query=*`)
 - streaming the response - download starts immediately

To test:
 - Find a large OMERO.table. - I'll create an example one on merge-ci soon...
 - Click on 'eye' to open as html table
 - Click to download as CSV, either with a query to filter rows or without query (whole table)

cc @manics

PR #197, opened 2020-07-30
Prevent large zip files for archived files from being created

manics:
This prevents too many large files being zipped up which can lead to excessive use of disk space, either on the node creating the zip or due to buffering in a proxy.

Closes https://github.com/ome/omero-web/issues/118 (at least it closes the particular URL that's causing problems, it's quite likely other URLs could cause problems too).

Testing:
1. Optionally set the maximum size limit to something small to make testing easier, e.g. 1MB: `omero config set omero.web.maximum_multifile_download_size 1048576`.
2. Find some images for which you can download the original archived files
3. Select enough images to take you over the limit you set in step 1. If you didn't set a limit the default is 1GB (should it be larger? Or unlimited?).
4. Attempt to download them. Instead of a zip link you should see an error message:

image

### DIFF
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -598,7 +598,7 @@ CUSTOM_SETTINGS_MAPPINGS = {
          1024 ** 3,
          int,
          "Prevent multiple files with total aggregate size greater than this "
-         " value in bytes from being downloaded as a zip archive."],
+         "value in bytes from being downloaded as a zip archive."],
 
     # VIEWER
     "omero.web.viewer.view":
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -593,6 +593,12 @@ CUSTOM_SETTINGS_MAPPINGS = {
          "Size, in bytes, of the “chunk”"],
     "omero.web.webgateway_cache":
         ["WEBGATEWAY_CACHE", None, leave_none_unset, None],
+    "omero.web.maximum_multifile_download_size":
+        ["MAXIMUM_MULTIFILE_DOWNLOAD_ZIP_SIZE",
+         1024 ** 3,
+         int,
+         "Prevent multiple files with total aggregate size greater than this "
+         " value in bytes from being downloaded as a zip archive."],
 
     # VIEWER
     "omero.web.viewer.view":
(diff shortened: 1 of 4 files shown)


omeroweb/settings.py
```diff
@@ -593,6 +593,12 @@ def leave_none_unset_int(s):
          "Size, in bytes, of the “chunk”"],
     "omero.web.webgateway_cache":
         ["WEBGATEWAY_CACHE", None, leave_none_unset, None],
+    "omero.web.maximum_multifile_download_size":
+        ["MAXIMUM_MULTIFILE_DOWNLOAD_ZIP_SIZE",
+         1024 ** 3,
+         int,
+         "Prevent multiple files with total aggregate size greater than this "
+         " value in bytes from being downloaded as a zip archive."],
```
mtbc, line 601: seem to start with a space here but end with a space above

joshmoore, line 601:
```suggestion
         "value in bytes from being downloaded as a zip archive."],
```

pwalczysko:
Tested on merge-ci.

Works as described in the header of the PR.

  - Two files just under 1GB - zip created and downloaded successfully.
  - Five files over 1GB - message in the zip part of the dialog appeared as expected.

The only suggestion:

``...requesting....`` in the error message sounds not too user-freindly. Also, the situation is quite clear from the context of the dialog, and selecting fewer files might not be a feasible suggestion for many situations. How about simply:

``Not available: Total size of files 2.3 GB is larger than the 1.0 GB limit.``

manics:
Updated:
image

What should we set the limit to? You could argue unlimited downloads is a bug and therefore we should add one, or that it's a feature and therefore adding a limit is a breaking change.

joshmoore:
I'd err on the side of a :bug: but we can help point at it in changelogs, announcements, and the production setup documentation to make it less painful.

pwalczysko:
The updated dialog looks good to me, thanks @manics 

Note the questions about the breaking/not-breaking change tactics.

will-moore:
Looks good. Not tested myself, but I trust @pwalczysko's testing.
1 GB seems like a good limit to me.
👍

sbesson:
Given the impact of the original issue in production deployment like UoD but probably others, what is the general feeling about the timeline for rolling out this fix into a patch/minor release? Would the OMERO documentation need an update as per https://github.com/ome/omero-web/pull/197#issuecomment-668138640?

will-moore:
Similar bug-fix is https://github.com/ome/omero-web/pull/192 (download table as csv) which fails for medium to large tables.
I think that is close to being merged (with #191)

jburel:
Some clarification will be required [here](https://omero-guides.readthedocs.io/en/latest/download/docs/download.html) at least

PR #198, opened 2020-08-05
shortening filter fix

### LINKED ISSUE #194: TypeError in common_filters
from ns-web-pub today:

```
2020-07-24 09:19:45,864 ERROR [omeroweb.webgateway.templatetags.common_] (proc.24008) shortening():155 Traceback (most recent call last):
File "/opt/omero/web/venv3/lib64/python3.6/site-packages/omeroweb/webgateway/templatetags/common_filters.py", line 153, in shortening
return value[:front]+"..."+value[length-end:]
TypeError: slice indices must be integers or None or have an __index__ method
```

cc: @will-moore

will-moore:
Fixes #194 

Another python 2 > 3 fix. Use integer division to fix error from issue above.

To test:
 - Check "Companion Files" tab of https://merge-ci.openmicroscopy.org/web/webclient/?show=image-83221 (user-3)
 - Should be shortened with `start...end` (3 dots in the middle).
 - Currently, the shortening fails (and is logged) but the full name is still displayed:

Screenshot 2020-08-05 at 15 39 31

### DIFF
--- a/omeroweb/webgateway/templatetags/common_filters.py
+++ b/omeroweb/webgateway/templatetags/common_filters.py
@@ -140,17 +140,15 @@ def shortening(value, arg):
         length = int(arg)
     except ValueError:  # invalid literal for int()
         return value  # Fail silently.
-    front = length/2-3
-    end = length/2-3
+    chunk = length//2-3
 
     if not isinstance(value, basestring):
         value = str(value)
     try:
-        length = len(value)
-        if length < length:
+        if len(value) < length:
             return value
-        elif length >= length:
-            return value[:front]+"..."+value[length-end:]
+        else:
+            return value[:chunk]+"..."+value[length-chunk:]
     except Exception:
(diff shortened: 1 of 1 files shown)


joshmoore:
I assume it would be good to roll this into a release, no? (I also wonder if there's any potential linting that would help us detect this type of error)

will-moore:
It's not particularly urgent - I think it can wait for the next release, whenever that is.
For the user, it's a silent failure, and you just see the full name for companion files instead of the truncated one.

will-moore:
I guess it would be nice to detect this kind of TypeErrors, but don't know if it's possible.
On the plus side, we don't tend to introduce these types of errors during development (as far as I'm aware!) - just still catching one's from Python 2 -> 3 migration.

will-moore:
#203 attempts to catch remaining integer division errors.

PR #203, opened 2020-08-26
Use integer division in doPagination()

will-moore:
Fixes https://www.openmicroscopy.org/qa2/qa/feedback/29485/

To test:
 - Go to https://merge-ci.openmicroscopy.org/web/webclient/history/ (user-3)
 - Click on today (with over 200 Datasets)
 - Page should load correctly (not with error as on QA)

### DIFF
--- a/omeroweb/webclient/controller/__init__.py
+++ b/omeroweb/webclient/controller/__init__.py
@@ -42,7 +42,7 @@ class BaseController(object):
 
     def doPaging(self, page, page_size, total_size, limit=PAGE):
         total = list()
-        t = (total_size/limit) + (total_size % limit > 0 and 1 or 0)
+        t = (total_size//limit) + (total_size % limit > 0 and 1 or 0)
         if total_size > (limit * 10):
             if page > 10:
                 total.append(-1)
--- a/omeroweb/webclient/views.py
+++ b/omeroweb/webclient/views.py
@@ -3011,9 +3011,9 @@ def omero_table(request, file_id, mtype=None, conn=None, **kwargs):
     context['data']['id'] = file_id
     context['meta']['query'] = query
 
-    # if we're on an exact page:
-    if offset == 0 or float(offset)/limit == offset/limit:
-        context['meta']['page'] = (offset/limit) + 1 if offset > 0 else 1
+    # check if offset matches an integer page number:
+    if offset == 0 or offset/limit == offset//limit:
+        context['meta']['page'] = (offset//limit) + 1 if offset > 0 else 1
 
     # pagination links
     url = reverse('omero_table', args=[file_id])
--- a/omeroweb/webgateway/templatetags/common_filters.py
+++ b/omeroweb/webgateway/templatetags/common_filters.py
@@ -79,6 +79,7 @@ def ago(value):
 
     if ago.total_seconds() < 0:
         return "Future times not supported"
+
     def plurals(val):
         return val != 1 and "s" or ""
     hours, remainder = divmod(ago.seconds, 3600)
--- a/omeroweb/webgateway/templatetags/common_filters.py
+++ b/omeroweb/webgateway/templatetags/common_filters.py
@@ -84,7 +84,7 @@ def ago(value):
     if ago.days >= 365:
         years = ago.days // 365
         return "%s year%s" % (years, plurals(years))
-    if ago.days > 28:
+    if ago.days > 31:
         months = ago.days // 30
         return "%s month%s" % (months, plurals(months))
     if ago.days > 0:
--- a/omeroweb/webgateway/templatetags/common_filters.py
+++ b/omeroweb/webgateway/templatetags/common_filters.py
@@ -77,6 +77,8 @@ def ago(value):
     except TypeError:
         return str(value)
 
+    if ago.total_seconds() < 0:
+        return "Future times not supported"
     def plurals(val):
         return val != 1 and "s" or ""
     hours, remainder = divmod(ago.seconds, 3600)
--- a/omeroweb/webgateway/templatetags/common_filters.py
+++ b/omeroweb/webgateway/templatetags/common_filters.py
@@ -82,10 +82,10 @@ def ago(value):
     hours, remainder = divmod(ago.seconds, 3600)
     mins, secs = divmod(remainder, 60)
     if ago.days >= 365:
-        years = ago.days / 365
+        years = ago.days // 365
         return "%s year%s" % (years, plurals(years))
     if ago.days > 28:
-        months = ago.days / 30
+        months = ago.days // 30
         return "%s month%s" % (months, plurals(months))
     if ago.days > 0:
         return "%s day%s" % (ago.days, plurals(ago.days))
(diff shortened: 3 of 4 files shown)


mtbc:
What/where is "today"?

mtbc:
Ah, if I go to History and pick yesterday, that has a big project, that's probably it? :smiley:

mtbc (approved):

jburel:
A full review of such problem should happen before the next release
That is the second one see also #198

will-moore:
Trying to find places where we have non-integer division using grep...
Find all files containing `/`, excluding 'Binary' files, those with integer division ` // ` and html closing brackets
still leaves quite a few:

```
$ grep -r "/" ./ | grep -v "Binary" | grep -v " // " | grep -v "</" | wc
   13374   88361 1804993
```

Need more targeted grep via regex I think...

will-moore:
Searching for digits like `d/d` or `d / d` is better...

```
$ grep -r "\d/\d" ./ | grep -v "Binary" | grep "\.py" | grep -v "\.js" 
.//test/unit/test_webgateway.py:        fname = r'1/2_3!"\'#$%&()=@€£‰¶÷[]≠§±+*~^\,.;:'
.//test/unit/test_webgateway.py:        # ext2/3/4 limit is 255 bytes, most others are equal to or larger
.//omeroweb/webgateway/views.py:    # make the aspect ratio (w/h) = 3/2
.//omeroweb/settings.py:        # http://blog.dscpl.com.au/2010/03/improved-wsgi-script-for-use-with.html

$ grep -r "\d / \d" ./ | grep -v "Binary" | grep "\.py" | grep -v "\.js" 
.//omeroweb/webgateway/templatetags/common_filters.py:        return value / 1000 / 100
.//omeroweb/webgateway/templatetags/common_filters.py:        return value / 1000 / 100 / 10
.//omeroweb/webgateway/templatetags/common_filters.py:        return value / 1000 / 100 / 10 / 1000
```

will-moore:
However, searching for digits won't find any cases where we are diving a variable by another.
Don't know how to do that!

joshmoore:
It's pretty tough with a regex. ```find . -name "*.py" -exec git grep -e "=.*\w/\w" {} \;``` at least matches the original issue.

mtbc:
:+1: Including an "n" with `grep -ne` can be helpful for this kind of thing.

will-moore:
Thanks @joshmoore . That gives me 222 lines which I can check manually.
These need more investigation..

```
omeroweb/webclient/controller/__init__.py:        t = (total_size/limit) + (total_size % limit > 0 and 1 or 0)
omeroweb/webclient/views.py:    if offset == 0 or float(offset)/limit == offset/limit:
omeroweb/webclient/views.py:        context['meta']['page'] = (offset/limit) + 1 if offset > 0 else 1
omeroweb/webgateway/templatetags/common_filters.py:    front = length/2-3
omeroweb/webgateway/templatetags/common_filters.py:    end = length/2-3
```
Searching for the same but with spaces gives a few more"
```
$ find . -name "*.py" -exec git grep -e "=.*\w / \w" {} \;
omeroweb/webgateway/templatetags/common_filters.py:        years = ago.days / 365
omeroweb/webgateway/templatetags/common_filters.py:        months = ago.days / 30
omeroweb/webgateway/webgateway_cache.py:                exp = time.time() + timeout + (timeout / 5 * random())
omero/plugins/web.py:                args = [sys.executable, location / app / "scripts" /
```

will-moore:
After ignoring the grep hits above for this PR and #198, 2 more fixes were needed (see above).

To test first fix:
 - Open an OMERO.table (>200 rows) in the table view (eye icon). Use 'Next' to go to next page. This should then say
`Showing page 2 (200 rows).` instead of `Showing page 2.0 (200 rows)`. 
 - Then edit the URL to an offset that is not an exact page. e.g. ?offset=100. Page should say
`Pagination offset: 100 limit: 200` instead of `Showing page 1.5 (200 rows)`. 

To test the other fix, you need to have created an OME-TIFF over 28 days ago, as an attachment on the image with the 
`omero.constants.namespaces.NSOMETIFF` (`openmicroscopy.org/omero/ome_tiff`) namespace. But, this isn't created by the Batch_Image_Export script (uses a 'Batch_Image_Export' namespace). Needs a fix elsewhere.

will-moore:
Used the same search for `omero-py` and found

```
src/omero/gateway/__init__.py:                    normalize = 1.0/chw[c][1]*(range-1)
test/unit/test_jvmcfg.py:        assert table[1] == int(2**15 * 15 / 100)
src/omero/util/image_utils.py:    b = check_rgb_range(rgba / 256 % 256)
src/omero/util/image_utils.py:    g = check_rgb_range(rgba / 256 / 256 % 256)
src/omero/util/image_utils.py:    r = check_rgb_range(rgba / 256 / 256 / 256 % 256)
src/omero/util/cleanse.py:    loops = ceil(wait * 1000.0 / ms)
src/omero/plugins/sessions.py:        val = float(timeout) / 60 / 1000
src/omero/plugins/admin.py:        loops = ceil(wait * 1000.0 / ms)
```
All of these are fine, but don't know about `test/unit/test_jvmcfg.py:        assert table[1] == int(2**15 * 15 / 100)`.

jburel:
First problem:
idr
Screenshot 2020-09-04 at 10 41 28

merge-ci
Screenshot 2020-09-04 at 10 43 59

Offset change:
idr
Screenshot 2020-09-04 at 10 46 03

merge-ci
Screenshot 2020-09-04 at 10 46 23

jburel:
For the second commit: difficult to find a file but I think a test should be available.

Testing locally the logic of the filter:
With or without the commit:
* ``value = datetime.datetime(2020, 8, 6)`` returns ``0 months``. Days should be used in that case.
* ``value = datetime.datetime(2020, 8, 1)`` returns ``1 month``
* ``value = datetime.datetime(2020, 8, 27)`` returns ``8 days``
* ``value = datetime.datetime(2000, 9, 27)`` returns ``19 years``
* Date in the future ``value = datetime.datetime(2020, 9, 27)`` returns ``11 hours``.

will-moore:
Test added.

jburel:
```
./test/unit/test_filters.py:28:1: E302 expected 2 blank lines, found 1
```

will-moore:
Grrr - I was trying to setup pre-commit hooks to avoid those flake8 failures, but need help: https://github.com/ome/omero-web/pull/207

jburel:
Thanks for adding the test but the problematic used cases are not covered
The function ``ago`` needs to be adjusted to cf.
```
assert ago(now - timedelta(days=29)) == '29 days'
E       AssertionError: assert '0 months' == '29 days'
```
The case of date in the future needs to be handle to

will-moore:
I added that test and fixed the handling of 29 days.
But I don't think we need to support future dates. This function is used exclusively for taking a creation date from the past.
Future dates will return a 'not supported' message (added a test for that too).

jburel:
The ``not supported`` message makes sense. I don't think we need to support it. It was returning values that were not valid.
There is a flake8 issue

jburel:
Thanks @will-moore for the changes. Merging

PR #209, opened 2020-08-28
Return 404 for imgData if image not found

will-moore:
See https://forum.image.sc/t/opening-omero-images-through-qupath/33703/37

Previously if you access `webclient/imgData/123/` and that image is not found or you don't have permission to view it, you would get returned a 500 response.
Now we correctly return a 404.

To test:
 - go to `webclient/imgData/ID/` with an invalid ID, logged-in as a regular user and check the 404 status in browser dev-tools Console.
 - logout then (on a server with public user configured) go to the same URL and you should get a 403 response.

### DIFF
--- a/omeroweb/decorators.py
+++ b/omeroweb/decorators.py
@@ -76,10 +76,15 @@ def get_client_ip(request):
     return ip
 
 
-def is_public_user(conn):
-    """Is the current user the same as the configured public user?"""
-    return (hasattr(settings, 'PUBLIC_USER') and
-            settings.PUBLIC_USER == conn.getUser().getOmeName())
+def is_public_user(request):
+    """
+    Is the session connector created for public user?
+
+    Returns None if no connector found
+    """
+    connector = request.session.get('connector')
+    if connector is not None:
+        return connector.is_public
 
 
 class ConnCleaningHttpResponse(StreamingHttpResponse):
--- a/omeroweb/decorators.py
+++ b/omeroweb/decorators.py
@@ -76,6 +76,12 @@ def get_client_ip(request):
     return ip
 
 
+def is_pubic_user(conn):
+    """Is the current user the same as the configured public user?"""
+    return (hasattr(settings, 'PUBLIC_USER') and
+            settings.PUBLIC_USER == conn.getUser().getOmeName())
+
+
 class ConnCleaningHttpResponse(StreamingHttpResponse):
     """Extension of L{HttpResponse} which closes the OMERO connection."""
 
--- a/omeroweb/decorators.py
+++ b/omeroweb/decorators.py
@@ -76,7 +76,7 @@ def get_client_ip(request):
     return ip
 
 
-def is_pubic_user(conn):
+def is_public_user(conn):
     """Is the current user the same as the configured public user?"""
     return (hasattr(settings, 'PUBLIC_USER') and
             settings.PUBLIC_USER == conn.getUser().getOmeName())
--- a/omeroweb/webclient/decorators.py
+++ b/omeroweb/webclient/decorators.py
@@ -127,9 +127,9 @@ class render_response(omeroweb.decorators.render_response):
         }}
 
         context.setdefault('ome', {})   # don't overwrite existing ome
-        connector = request.session.get('connector')
-        if connector is not None:
-            context['ome']['is_public_user'] = connector.is_public
+        public_user = omeroweb.decorators.is_public_user(request)
+        if public_user is not None:
+            context['ome']['is_public_user'] = public_user
         context['ome']['eventContext'] = eventContextMarshal(
             conn.getEventContext())
         context['ome']['user'] = conn.getUser
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -43,7 +43,7 @@ from django.contrib.staticfiles.templatetags.staticfiles import static
 from django.views.generic import View
 from django.shortcuts import render
 from omeroweb.webadmin.forms import LoginForm
-from omeroweb.decorators import get_client_ip
+from omeroweb.decorators import get_client_ip, is_pubic_user
 from omeroweb.webadmin.webadmin_utils import upgradeCheck
 
 try:
@@ -1494,7 +1494,11 @@ def imageData_json(request, conn=None, _internal=False, **kwargs):
     key = kwargs.get('key', None)
     image = conn.getObject("Image", iid)
     if image is None:
-        return HttpResponseNotFound(f'Image:{iid} not found')
+        if is_pubic_user(conn):
+            # 403 - Should try logging in
+            return HttpResponseForbidden()
+        else:
+            return HttpResponseNotFound(f'Image:{iid} not found')
     if request.GET.get('getDefaults') == 'true':
         image.resetDefaults(save=False)
     rv = imageMarshal(image, key=key, request=request)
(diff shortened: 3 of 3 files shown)


will-moore:
When QPath is trying to load an image from a server that requires a login (get a 403 response from `webclient/imgData/1/`) then it displays a login popup.
However, if the server has a public user configured (but the image you want is not public), then you previously got a 500 and now you will get a 404. But, what you really want is a login to access the non-public image.

This is equivalent to the situation we used to have in webclient. E.g. if you went to `/webclient/?show=image-1` on a server that had public user configured (but the image wasn't public) then you'd just get the webclient UI loading but no image found. 
This was fixed in https://github.com/ome/omero-web/commit/96d04ab1586678b691765482dadc958afe4ada57, where if you have e.g. `?show=image-1` and that image is not found AND you're logged in as public user, then redirect to login page (handle the same as a 403).

So, I'm wondering if the `webclient/imgData/1/` end-point, and possibly many other endpoints, should behave in the same way? If the image isn't found AND you're logged in as public user then return a 403 response, to indicate a permission denied, rather than 404 - not found?

cc @joshmoore

sbesson:
Interesting thoughts.

I am all for unifying the error codes across the code base and making them as informative as possible for the end user. Adding a quick thought, we could also consider returning an HTTP 401 in the case of accessing data as the public user - see https://tools.ietf.org/html/rfc7235#section-3.1. As far as I understood, 401 should be returned in the `Unauthenticated` case while 403 should be returned in the `Unauthorized` case.

will-moore:
Hmmm - I think 403 is correct for public user since they are Authenticated (as a public user) but not Authorised.
But we may not be technically correct to return 403 when user has made no attempt to login (or has failed to login). That should possibly return a 401.
However, it probably reduces the burden on client developers if they only have to handle 403 (and then attempt to login).
Also I notice that Django doesn't provide a standard HttpResponse for 401 (see https://docs.djangoproject.com/en/2.2/_modules/django/http/response/) so it may be generally less used.
And probably it's not worth breaking the `@login_required` API by switching from 403 to 401.

We can never be entirely correct, since if we try to find an Image in OMERO by ID, and it's not found, we don't know whether the Image actually exists (but we can't access it - 403) or if it doesn't exist (404)!

mtbc (approved):
It's possible for clients to tell if something exists but is forbidden, e.g., the query service's "get" will throw a `SecurityViolation` instead of a `ValidationException`, but I think best to stick to 404 to leave it open that one may not really be supposed to be able to tell the difference.

omeroweb/decorators.py
```diff
@@ -76,6 +76,12 @@ def get_client_ip(request):
     return ip
 
 
+def is_pubic_user(conn):
```
mtbc, line 79:
```suggestion
def is_public_user(conn):
```
:smiley:

omeroweb/decorators.py
```diff
@@ -76,6 +76,12 @@ def get_client_ip(request):
     return ip
 
 
+def is_public_user(conn):
+    """Is the current user the same as the configured public user?"""
+    return (hasattr(settings, 'PUBLIC_USER') and
+            settings.PUBLIC_USER == conn.getUser().getOmeName())
```
sbesson, line 82: Is there a difference between this logic and the one used in another place of the decorators https://github.com/ome/omero-web/blob/fcd02c18db0c03761f8ca666e66868483fa72ff6/omeroweb/webclient/decorators.py#L130-L132 ?

will-moore: Yes, that logic uses the request to get the stored `connector` that was created by the `@login_required` decorator and testing that for `is_public`.
So, for that logic you need access to the `request`, and for this PR, `is_public_user` uses the `conn`, but does potentially more work to look-up the user.
I guess I could reuse the existing logic since you mostly have access to the `request`.

joshmoore:
In general, :+1: for making the switch to 40x that most resembles what we're doing elsewhere now. (Is that 404?)

Then, we probably move quite a bit of this discussion to a new issue. Things that occur to me:

(1) https://github.com/ome/omero-web/pull/209#issuecomment-682762813
> Hmmm - I think 403 is correct for public user since they are Authenticated (as a public user) but not Authorised.

I think we need to decide where we are going with public user. Does someone need to know that they are public user or is it just public access. This wasn't worked through with PU was rolled out.

> we don't know whether the Image actually exists (but we can't access it - 403) or if it doesn't exist (404)

I'm not too terribly worried about the difference between these two. (I certainly saw exactly these issues with S3) But I *do* think we should provide a complete list of exception codes which mean, "this isn't a temporary failure, you're simply not going to be able to see this image etc. with your current permissions"


(2) https://github.com/ome/omero-web/pull/209#pullrequestreview-478459925

>  t's possible for clients to tell if something exists but is forbidden, e.g., the query service's "get" will throw a SecurityViolation instead of a ValidationException

Darn. We certainly try to avoid such leakage. But for logged in users, that's probably ok. With _might_ want to review how PU handles this and try to make it as information-neutral as possible.

pwalczysko:
Test 1:

  - logged in as user-3 to merge-ci, construct an invalid id url https://merge-ci.openmicroscopy.org/web/webclient/imgData/12000000000/
  - I am getting a 404 message in Console
Screen Shot 2020-09-01 at 14 53 53

Test 2:

  - logout as user-3 and then do not log in
  - go to the same url as in Test 1 above
  - in Console, observe the similar output as for Test 1

Screen Shot 2020-09-01 at 14 58 05


Test 3: (bonus)

   - still logged out as user-3, not log in, but navigae to a valid image url
   - I end up on a login page, which is expected


This seems to pass the test.

will-moore:
@pwalczysko yes, importantly the "similar output as for Test 1" you see is actually a different response code: 403 (as expected) instead of a 404.
Thanks.

will-moore:
Unfortunately I can't open https://merge-ci.openmicroscopy.org/web/webclient/img_detail/105665/ in QuPath to test that this is working. It *should* allow QuPath to show a login to access a non-public image on a server that has public user configured.

I also can't get it to work against my local dev web server (even without public user enabled).
In both cases I see
```
ERROR: Unable to connect to OMERO server
ERROR: Unable to build server: Unable to build ImageServer for http://localhost:4080/webclient/img_detail/1254/.
```

pwalczysko:
When trying to open an image from merge-ci, I have (similarly to @will-moore ) an error. 

```
INFO: Project set to Project: project3-project
ERROR: Unable to connect to OMERO server
```

will-moore:
Thanks for trying Petr.
But, as I was suggesting when we discussed this:
I think that the `500 -> 404/403` fix is valid (and has been tested).
Although the original bug was identified via QuPath, testing login with QuPath shouldn't be a blocker to merging this, especially since we don't have a test server where we can connect with QuPath.

If we find that we *still* can't connect with QuPath (for some other reason) then we can address that in a follow-up PR.

will-moore:
Discussion moved to a new issue https://github.com/ome/omero-web/issues/213
as suggested above: https://github.com/ome/omero-web/pull/209#issuecomment-684707995

joshmoore:
* :+1: for the issue to continue the discussion. 
 * :+1: for this change happy to get it in.
 * I do wonder if there are other locations that need the same fix, e.g. looking at the next use of HttpJavascriptResponseServerError:

```
def wellData_json(request, conn=None, _internal=False, **kwargs):
    wid = kwargs['wid']
    well = conn.getObject("Well", wid)
    if well is None:
        return HttpJavascriptResponseServerError('""')
```
I assume the same would happen? Fine for that review to be a follow on.

will-moore:
Created a new issue to address other places we need to fix HttpJavascriptResponseServerError('""') for Not Found.
https://github.com/ome/omero-web/issues/215

PR #211, opened 2020-09-01
disk usage percent fix

will-moore:
See https://forum.image.sc/t/omero-web-reports-disk-usage-incorrectly/42219

This rounds the usagePercent value before casting it to Int, to avoid strange JavaScript parseInt behaviour of ignoring exponentials for very small numbers (see example in link above).

To test: - probably not so easy to create multi-petabyte free disk space.
So might have to just check the code change against the examples in the link above.

### DIFF
--- a/omeroweb/webadmin/templates/webadmin/includes/drivespaceStats.html
+++ b/omeroweb/webadmin/templates/webadmin/includes/drivespaceStats.html
@@ -34,8 +34,17 @@
 
             <div class="summary">
                 <!-- show 'Total usage' or 'Group usage' or 'User usage' -->
-                <h3><span class="totalUsage">Total</span> {% trans "usage" %}: <span id="total"></span></h3>
-                <h3>{% trans "Free space" %}: <span id="free_space"></span></h3>
+                <p>
+                    <span class="totalUsage">Total</span> {% trans "usage" %}:
+                    <span id="total" style="color:black"></span><br>
+                    {% trans "Free space" %}:
+                    <span id="free_space" style="color:black"></span>
+                </p>
+                <p>
+                    NB: Total usage is calculated as the sum of all data, including any 'in-place' imports,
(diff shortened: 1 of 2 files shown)


pwalczysko:
Screen Shot 2020-09-02 at 15 25 52

On merge-ci, this seems to be okay (?) Although how many TB do we have for our peruse there ?
https://merge-ci.openmicroscopy.org/web/webadmin/stats/
Even 1 TB seems excessive to the best of my knowledge, so not sure here.
The ratio counted seems correct, it is 3%

joshmoore:
```
bash-4.2$ du -sh *
4.8G	BioFormatsCache
1.0K	DropBox
2.3G	Files
3.5G	FullText
66G	ManagedRepository
51G	Pixels
53M	Thumbnails
```

joshmoore:
Confirm separately. Free size comes from `df`. Used size comes from a query in the database (which ignores symlinks).

will-moore:
The web UI reports `diskFree / (diskFree + usedSize)` as a percentage.
If anyone wants to propose an update to the UI that tries to take into account the in-place imported data, please go ahead and open a new issue.
But I think the bug-fix here to the current functionality is working and can be merged?

joshmoore:
Certainly no objections for this PR: taking in-place into account is going to require server-side changes. :+1:

sbesson:
As a minor doc RFE, is there some place in the client UI where the caveat about the in-place imports could be mentioned?

pwalczysko:
Agree with @sbesson about the explanation/warning in the UI.

Screen Shot 2020-09-04 at 11 38 23

Could an additional line under "Drive space usage" saying ``The in-place imported data are not counted correctly. The disk usage might be smaller than indicated."

?

will-moore:
Added warning message:

Screenshot 2020-09-11 at 12 21 52

manics (approved):

Issue #221, opened 2020-09-16
Move production Dockerfile into this repo


manics:
https://github.com/ome/omero-web-docker predates the decoupling of the omero-web repo. It now results in unnecessary work as rebuilds and tags have to be triggered on that repo. We could simplify things by moving the Dockerfiles into this repo and using a GitHub action or something else to orchestrate the building and release of all Docker image

The main downside is we'd loose the ability to easily update the Docker images on their own, unless we have multiple tagging schemes (e.g. prefix Docker tags with `docker-`).

joshmoore:
One issue might be whether or not we start including other products in omero-web, e.g. https://github.com/ome/docker-example-omero/issues/6#issuecomment-694097399

One option might be to have this repo produce a base image which others then build on.

manics:
I can see that working- minimal base image in this repo, and use dependabot to update the Dockerfiles for downstream images in the other repo.

PR #224, opened 2020-09-23
WIP: Rewrite Dockerfile

manics:
Since it's almost pure Python (apart from Ice) the base image doesn't matter as much. `ubuntu:20.04` is slightly smaller than `centos:7`

This will eventually become the base production image, but I think we should wait for `6.0.0` before switching the `openmicroscopy/omero-web` to use this build, just in case someone is relying on a feature of the existing image. We could optionally tag it with a different name or tag though.

This is on top of https://github.com/ome/omero-web/pull/218 to avoid future conflicts.

Closes https://github.com/ome/omero-web/issues/221
Uses https://github.com/manics/omero-cli-externalconfig/

--exclude


jburel:
We can look at it again later. Closing

PR #17, opened 2019-09-18
Right panel performance fixes

will-moore:
Port PR from https://github.com/ome/openmicroscopy/pull/6075/

This PR contains 2 approaches to improve right panel performance:
 - tries to reduce the times that `conn.getExperimenter()` is called for right panel - e.g. single P/D/I etc (see link above)
 - improves loading speed of batch-annotate panel (when multiple objects selected) by NOT loading the toolbar download menu until the user clicks on it. This was being very slow when many Images are selected (as in https://github.com/ome/omero-web/pull/225) because the 'canDownloadAsJpg()` to enable png/tiff/jpeg download was calling 'image.getSizeX()` etc for every image (loading the pixels on the fly each time). To test, select multiple objects and test behaviour of the Download menu.

PR #225, opened 2020-09-24
Query string ids

### LINKED ISSUE #223: URL query for filter or multi-select too long with many images
If you have e.g. 500 images per page and you try to filter those images or select too many of them, the URL to load the right panel or to load annotations to filter by is too long. E.g. `?image=1234&image=3456...` etc. 
For 500 images, this query string is about 7k and the limit is generally around 4k.
The `&image=` adds 7 "unnecessary" characters for each Image.

E.g. can't filter by Annotations at:
http://idr.openmicroscopy.org/mapr/gene/?value=ARL13B&show=project-501

One solution is to use a different notation for image IDs in the query, which we already do in some places:
`?image=1234,3456, etc`. This will likely reduce the query string length for 500 images below 4k

Possibly, we could support both notations in a bunch of places, so as not to make breaking API changes.
Need to investigate all the various places we do this.

will-moore:
Fix #223 
Where URLs use e.g. `?image=1&image=2`, we now also support more concise `?image=1,2`.
This allows us to avoid query strings becoming too long (greater than 4k) with e.g. 500 images.

Also, where we use `?Image=1,2`, we now also support `?Image=1&Image=2`.

The webclient now uses the more concise form for loading right panel and loading annotations (required when filtering centre panel).

To test:
 - Need to have `omero.web.page_size` set to 500
 - Open a Dataset with > 500 images - e.g. https://merge-ci.openmicroscopy.org/web/webclient/?show=dataset-13651
 - Select them all - right panel should load
 - Expand the various annotation panels in right panel
 - Try adding e.g Tag, Rating, Key-Value pairs, Comments to ~500 images
 - Filter centre panel by Tag, Rating, Key-Value pairs

NB: Need to check which parts of this PR need to be ported to mapr, which overrides some parts of webclient.


will-moore:
Seems `black` and `flake8` passed locally when creating that last commit...
```
(omeroweb2) LS30778:omero-web wmoore$ git commit -m "Handle AnnotationCounts with ?dataset=1&image=2"
[WARNING] Unstaged files detected.
[INFO] Stashing unstaged files to /Users/wmoore/.cache/pre-commit/patch1601029874.
[INFO] Initializing environment for https://github.com/psf/black.
[INFO] Installing environment for https://github.com/psf/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
black....................................................................Passed
flake8...................................................................Passed
[INFO] Restored changes from /Users/wmoore/.cache/pre-commit/patch1601029874.
[query_string_ids a5ee03581] Handle AnnotationCounts with ?dataset=1&image=2
 1 file changed, 3 insertions(+)
```

joshmoore:
> Seems `black` and `flake8` passed locally when creating that last commit...

Is it possible it was only testing the changes in that one commit?

will-moore:
Yes, I think so. I didn't have pre-commit enabled for the previous commit.

manics (commented):
If I select multiple images and click on the `Link to this` button I get `https://merge-ci.openmicroscopy.org/web/webclient/?show=image-85518|image-85519|image-85523|image-85564|...` Is this query format in scope for the PR, or a separate issue?

Editting map-annotations seems to be broken, both individual and batch edit.

Tagging multiple images sometimes works but there's some weird behaviour. In a dataset with >500 images:
1. Add a tag to the first image in the dataset.
2. Select 500 images including the first one. The tag appears in the right hand panel even though it's only applied to the first image:
image
3. `Save`. The tag is _not_ applied to the other 499 images.
4. If instead I select just 4 images including the first, the dialog appears as expected.
image
Adding and saving the tag applies to the selected images.

Adding and deleting a comment on 500 images works.

Select 500 images. Click the `+` under attachments. The `Choose attachments` seems to get stuck on `Loading attachments`. Is it perhaps running a query for each image?

Adding/removing ratings from 500 images works. Noticed an anomaly unrelated to this PR https://github.com/ome/omero-web/issues/230

Several issues here, I'm not sure which are due to this PR and which are existing issues that have only become apparent due to this PR letting you select more images than previously, so feel free to indicate what is/isn't in scope.

jburel:
Screenshot 2021-01-07 at 16 19 07
I was able to create KVP

pwalczysko:
http://idr.openmicroscopy.org/webclient/?show=dataset-3558 is a good candidate for testing - filtering by anything else except by name does not work

pwalczysko:
Bug:  Error when annotating many images with KVP:


 - go to https://merge-ci.openmicroscopy.org/web/webclient/?show=dataset-13651 and select all the images in the first page
 - expand the KVP hamonica and start creating a new KVP
 - observe an error dialog after you hit Enter on entering the Value
 - the error dialog is empty (hard to get more  info, but @will-moore is able to reproduce)
 -

pwalczysko:
If a KVP (or, I suspect, any annotation) is added to many objects, the tooltip is trying to list all of them, but it does not go very far, instead listing just listing one image 1 repeatedly - it is very hard to screenshot a tooltip, sorry. But this can be seen on the KVP "home" (there is no value to it), which was added to all images of https://merge-ci.openmicroscopy.org/web/webclient/?show=dataset-13651

will-moore:
The error reported in https://github.com/ome/omero-web/pull/225#issuecomment-762230533 is not due to this PR.
It's just another issue revealed when trying to add annotations to many images (which we've never really tested before).
I can try to fix it, but wouldn't like it to be a blocker on merging this PR, which is primarily to fix the IDR use-case above.

jburel:
Screenshot 2021-02-03 at 14 58 49

Tag dialog did not load for me when I selected all the displayed images and clicked on the ``+`` button

jburel:
@will-moore have you checked with parts need to be ported to mapr?

jburel:
Since this PR aims at reducing the size of the url, the url generated when clicking on link icon (right-hand pane) could be reviewed. This will require changes in app like figure handling this type of url. Note that figure managed to  handle the url.

jburel:
When I tried to edit the KVP, the "Send Feedback" dialog popped up but there was no visible message

will-moore:
Re: https://github.com/ome/omero-web/pull/225#issuecomment-772576162 
There is a PR in `omero-mapr` with the corresponding changes to 1 file above https://github.com/ome/omero-mapr/pull/64

jburel:
* Open 5269
* Select the images
* Select tag.
* Tag dialog pops up, selects a tag
* Click Save
Screenshot 2021-03-01 at 10 52 33

will-moore:
@jburel Trying to reproduce...

 - What is `5269`?
 - When you say 'Select tag`, are you adding a Tag by clicking the `[+]` button?
 - Then choosing an existing tag in the left hand column, clicking `>` to move it into the right column and clicking Save?

This just worked for me when selecting the 500 images in https://merge-ci.openmicroscopy.org/web/webclient/?show=dataset-13651 and adding an existing Tag.

jburel:
I have updated the description.
The same problem happened before.
The second time it seems to work

will-moore:
I tried a bunch of adding and removing tags to images in that Dataset and it seems to be working. It's a bit slow. I wonder if it's just timing-out sometimes. Can you look in the dev-tools Console or Network tabs for any errors? thx

will-moore:
@jburel - I'd love to try and get this merged. Could you see if you're still getting the errors you reported above for the Tag dialog? Thanks.

jburel:
The tag dialog loads but it takes time and b/c there is no feedback e.g. spinner I end up having a blank dialog for a good amount of time.
user-1 has a limited number of tags so I will need to create more tags to see the impact. 
The data loading without feedback is also an issue on the right-hand panel. the General tab is blank for several seconds when numerous objects are selected

jburel:
--exclude

will-moore:
@jburel Another example of users needing to filter larger numbers of images at https://forum.image.sc/t/omeroweb-visualising-images-on-filtered-sets-for-labeling/71558/2 which will expose the issue that this PR addresses...

Could you remove the `exclude` flag (or OK for me to do it)?
I'll try to reproduce the issues described above... I could/should try to address the "show loader for slow responses" issues in a separate PR, but could do that before this one if that's a blocker for testing.

will-moore:
With this PR and https://github.com/ome/omero-mapr/pull/64 deployed on `merge-ci` again, I've tested tagging, filtering, rating, file attachments and key-value pairs.
The only issue I saw was in the *creation* of Key-Value pairs for a large number of Images. If the auto-save triggers to update KVP while the initial call is still creating them, you get an error (not sure why).
I can look into that, but I would argue that it's a separate bug (due to slow saving) from the one that is being fixed in this PR (URL request strings too long).

will-moore:
@pwalczysko @jburel OK for another round of testing, to see if there's any reproducible issues that need fixing?

jburel:
* https://github.com/ome/omero-web/issues/400
* https://github.com/ome/omero-web/issues/401
*

omeroweb/webclient/controller/container.py
```diff
@@ -239,8 +239,14 @@ def getAnnotationCounts(self):
 
     def getBatchAnnotationCounts(self, objDict):
         """
-        Get the annotion counts for the given objects
+        Get the annotion counts for the given objects.
```
jburel, line 242:
```suggestion
        Get the annotation counts for the given objects.
```

will-moore:
@jburel Thanks for the testing. I'm not sure which of those issues need fixing before this PR is good to merge?

Related: I revived a PR (even older than this one!) at https://github.com/ome/omero-web/pull/17 and tried to address performance of right panel...

pwalczysko:
Went through the RHP entirely (Tags, attachments, Comments, Ratings, download menu, link, scripts, Open With) whilst having 500 images selected today.
All was fine (on UoD network) except:

1. no downloads menu resulted in a valid window (only blank window)
2. 500 images do not open in Figure
3. the url harvested successfully (?) from the top right corner icon, when pasted into a new tab, does not bring me to the 500 images, insead it just does not bring up anything
4. Split View Figure script has a blank window (but the Thumbnail figure works as expected)

will-moore:
- #400 is fixed by #423
- ... more to come...

jburel:
Excluding this PR ahead of the dependencies upgrade work
--exclude

PR #232, opened 2020-10-02
Openwith fixes

will-moore:
Fixes some issues with Open-with, noticed when using https://github.com/will-moore/omero-openwith-url
When the openwith function to getUrlProvider() is called by the right panel, it includes 'name' info for selected image(s).
Previously, this was only available when the open-with was used on the jsTree.
Also, the right-panel open-with getUrl() is only called when the Open-with menu is clicked on. This means we don't call those methods unless needed, which is important if e.g. the open-with is doing an ajax call to check on enabled status.

Both the right-panel and jsTree support the passing of a callback as the second argument to `isEnabled(selected, callback)` allowing plugins to do async look-up of the enabled state and to call the callback.

To test:
 - Test https://github.com/will-moore/omero-openwith-url/pull/1 as described.
 - Also check that other open-with items are still functioning as before - Enabled/disabled as expected depending on what's selected. And open-with opens Image as expected.

PR #233, opened 2020-10-05
Add a URL Link as output of OMERO.script in Activities tab

emilroz:
The changes here add additional output option ("URL") to the OMERO.script engine.

Example script utilizing new feature: https://gist.github.com/emilroz/e82d9b8f53f759c23235e2f38256d413

Expected result:
image

New "Open external link" button should open a new tab with the specified HTTP address.

We utilize this feature when the output of the script creates URLs to OMERO viewers/plugins (PathViewer, Figure, PathViewer Grid, etc.) with settings encoded in the URL based on user input.


manics:
Functionally this works 👍 

Things to consider:
- It looks like `v.type` is currently an OMERO type, this adds `URL` to the supported types
- Should we document this somewhere?
- Related to the previous point, the URL is not escaped (not necessarily a problem), but script developers must be aware of this

emilroz:
Thanks @manics. I'll add the URL encoding/decoding where appropriate. Where do you think would be the best place to document it?

manics:
Not sure on where we should document it, maybe wait for @will-moore to take a look? I'd hold off on any encoding/decoding until we have his input as it can be tricky to get right, it may be best to rely on the script developer.

will-moore:
I think the place to add to docs is under https://github.com/ome/ome-documentation/blob/develop/omero/developers/scripts/style-guide.rst#script-outputs

I'm not sure that we need to escape the URL. Scripts should be trusted etc. Do we need to support URLs with " etc? I guess it might improve usability for devs to handle this, but at least if this is documented then I don't see it as an essential feature.

emilroz:
@will-moore, @manics docs PR opened here: https://github.com/ome/ome-documentation/pull/2165, please let me know if there's anything missing.

emilroz:
@will-moore, @manics code updated to remove the `id` requirement. I have also added changes proposed by @will-moore to the test script here: https://gist.github.com/emilroz/e82d9b8f53f759c23235e2f38256d413

will-moore:
Works fine, thanks. LGTM. 👍

emilroz:
@will-moore anything else we need to add here before the PR is approved?

will-moore:
I think it's good to merge.

sbesson:
Is it worth scheduling a quick omero-web release including these changes?

emilroz:
I'm all for it of course :)

manics:
If we do it would be nice if we could get https://github.com/ome/omero-web/pull/232 in too so we can start linking to vizarr for testing (https://github.com/IDR/deployment/pull/297)

PR #235, opened 2020-10-06
Allow a custom javascript tree key function

manics:
The key used to sort items in the JSTree can be customised by defining a javascript function that returns the key used for sorting. This function takes the container name as a parameter, and can do anything that javascript can do.

This is an implementation of https://github.com/ome/design/issues/108#issuecomment-704141276

For example
```
config set omero.client.ui.tree.type_order false
config set omero.web.ui.tree.sort_key_function "function(s){var a=/(\w+)-.+\/\w+(\w)/.exec(s); if (a) {return a[2]+'-'+a[1];} else {return s;}}"
```
This will order first by ABCDE index (ie list all `screenA/experimentA` from any study, then list all `screenB/ExperimentB` from any study), and within `screenA/experimentA` order by idrNNNN:

image

In case it's not obvious this is just an example to illustrate the sort of thing you can do since the dataset driving this work isn't yet live, in practice we'd want a key along the lines of `idrNNNN-[ABCDEF]`:
`"function(s){var a=/(\w+)-.+\/\w+(\w)/.exec(s); if (a) {return a[1]+'-'+a[2];} else {return s;}}"`

The `if(a)` ensures that if a container name such as `idr0038-held-kidneylightsheet` doesn't match the regex it is used as is.

### DIFF
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -1001,6 +1001,18 @@ CUSTOM_SETTINGS_MAPPINGS = {
             "The javascript loads data into ``$('#div_id')``."
         ),
     ],
+    "omero.web.ui.tree.sort_key_function": [
+        "TREE_SORT_KEY_FUNCTION",
+        "null",
+        str,
+        (
+            "An inline javascript function that returns a key used for "
+            "sorting the tree. Must be of the form (string) -> string, e.g. "
+            "``function(name) {return name.split('-')[1];} ``. "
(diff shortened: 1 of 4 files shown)


manics:
I've rebased this, but I haven't retested this

will-moore:
This is conflicting with https://github.com/ome/omero-web/pull/270 and is not actively being worked on, so I will close it for now.

PR #240, opened 2020-10-20
Migrate CI to GitHub actions

sbesson:
Largely inspired by the work in omero-py this migrates the CI to using GitHub actions and also adds coverage for Python 3.7 and 3.8

- create GitHub workflow with 3 steps: install from source, test and publish
- remove Travis-specific environment from tox.ini and use markers like omero-py
- expand envlist to run tests under Python 3.7 and 3.8
- remove .travis.yml
- update CI badge and add badge for PyPI


joshmoore:
In general, big :thumbsup:. I'd still like to see some for of re-use, but I'm failing to see an easy path forward where changes can be propagated:

 - https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/sharing-workflows-with-your-organization#creating-a-workflow-template -- allows copying for the first instance, but updates don't get propagated

manics:
To reuse this you'd probably want to split it into several github actions which do one thing, e.g. install standard dependencies, run standard scripts, a bit like https://github.com/ome/omero-test-infra but with one action per stage instead of bundling everything together.

sbesson:
Folllowing the preliminary work on omero-cli-zarr and https://github.com/ome/.github/pull/4, I went ahead and split the single workflow into 3 workflows. The PyPI publication should match the proposed template and I would assume the Tox one could also be turned into a template

sbesson:
Relisting for a last round of comments and objections to merging this in the current state? 
Proposed next step as a few of our Python libraries are using `tox` is to test the tox workflow  and eventually convert it into a template.

joshmoore (approved):
In general, big :+1:. Some of the badges are hard to confirm until after this is merged, but certainly no objections to iterating through this. Thanks.

PR #243, opened 2020-11-20
Get table rows batch

kkoz:
Modify `table_query` to get the rows in a single batch using `slice` instead of `read`. Also, add a `table_metadata` endpoint to fetch just column information and the total number of rows in the table without any data.

### DIFF
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -2945,7 +2945,11 @@ def _table_query(request, fileid, conn=None, query=None, lazy=False, **kwargs):
         if not offset:
             offset = int(request.GET.get("offset", 0))
         if not limit:
-            limit = int(request.GET.get("limit")) if request.GET.get("limit") is not None else None
+            limit = (
+                int(request.GET.get("limit"))
+                if request.GET.get("limit") is not None
+                else None
+            )
         range_start = offset
         range_size = kwargs.get("limit", rows)
         range_end = min(rows, range_start + range_size)
@@ -2966,7 +2970,6 @@ def _table_query(request, fileid, conn=None, query=None, lazy=False, **kwargs):
             except Exception:
                 return dict(error="Error executing query: %s" % query)
 
-        logger.info("Retrieving {} rows".format(len(hits)))
         def row_generator(table, h):
             # hits are all consecutive rows - can load them in batches
             idx = 0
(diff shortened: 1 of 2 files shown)


joshmoore:
@kkoz : can you run `pre-commit run -a` and push the formatting changes?

omeroweb/webgateway/views.py
```diff
@@ -2962,29 +2971,23 @@ def _table_query(request, fileid, conn=None, query=None, lazy=False, **kwargs):
...
-                        for col in table.read(range(len(cols)), hit, hit + 1).columns
+            # hits are all consecutive rows - can load them in batches
+            idx = 0
+            batch = 1000
+            while idx < len(h):
+                batch = min(batch, len(h) - idx)
+                col_indices = range(len(cols))
+                if col_names:
+                    col_indices = [
+                        i for (i, j) in enumerate(cols) if j.name in col_names
```
will-moore, line 2982: Tiny point - It looks like the col_names / col_indices logic could be moved out of the `while` loop, which would make it slightly easier to understand?

will-moore:
Functional testing and code looks good. Minor code comment.
Looks like this got temporarily excluded from the merge build: https://github.com/snoopycrimecop/omero-web/commit/dfb12b2f6bb16a87693a82068e53fd30c89d6bc5 so can't confirm tests etc.

kkoz:
Thanks for the suggestion @will-moore - is there something I need to do to get scc to pick up this change?

sbesson:
@kkoz nothing more than you have already done so far I am afraid (thanks for fixing pre-commit). The core issue here is that we are migrating away from travis-ci.org, which is getting shut down, to GitHub actions/workflows. Unfortunately, the latter uses a different API to report its status and the onus is on usto update scc to support this endpoint and fix the daily testing for all repositories which have been migrated.

This upgrade is very much on my todo list and I would hope to have it resolved in the upcoming days. Do you have a timeline for integrating these changes?

kkoz:
@sbesson Clarification question - are you asking when we would like these omero-web changes merged or when we can implement the necessary scc and testing changes for the new endpoint?

sbesson:
@kkoz I managed to get `scc 0.15.0` released <2h ago with the necessary endpoints. This PR (and others) should now been included in the daily OME CI builds. So we'll put it on our to-review list this week and/or notify you if it destabilizes tests.

kkoz:
@jburel Are there any documentation updates you would like me to do with respect to this PR?

jburel:
The example https://github.com/ome/openmicroscopy/blob/develop/examples/Training/python/Tables.py and the corresponding page https://github.com/ome/ome-documentation/blob/develop/omero/developers/Python.rst could be updated

sbesson:
Now https://github.com/ome/openmicroscopy/pull/6260 has been opened and the examples are tested daily, should we get this reviewed and merged @jburel ?

omeroweb/webgateway/views.py
```diff
@@ -2994,7 +2995,7 @@ def row_generator(table, h):
                 "columns": [col.name for col in cols],
             },
             "meta": {
-                "rowCount": rows,
+                "rowCount": len(hits),
```
will-moore, line 2998: The `rowCount` is supposed to be the total number of rows in the table, not the number of rows returned by the query, which is the `totalCount` below (maybe the naming is confusing).

will-moore:
https://merge-ci.openmicroscopy.org/web/webgateway/table/189802/metadata/ (user-3) looking good.

However, the 'rowCount' issue (see https://merge-ci.openmicroscopy.org/web/webclient/omero_table/479011/json/) where `rowCount:500` instead of `60480` means that the total number of rows is not shown at https://merge-ci.openmicroscopy.org/web/webclient/omero_table/479011/

kkoz:
@will-moore Do you need anything else from me on this?

will-moore:
Looks good, thanks. 👍

PR #250, opened 2021-01-19
filter KV show auto-complete

### LINKED ISSUE #247: Filter by Key-Value: show auto-complete immediately
There is auto-complete feature for Values when filtering the centre panel by Key-Value pairs,
but it only pops up once you have started typing, so you need to know the first letter, which can be tricky.

Much better to show the auto-complete list of Values as soon as the user picks a Key.

See https://forum.image.sc/t/a-couple-of-queries-about-omero-web-iviewer-and-search-by-key-value-pairs/47334/3?u=will-moore

will-moore:
Fixes #247

This improves the usability of filtering Images by Key-Value pairs.
To test:

 - Filter by Key-Value and choose a Key...
 - This should immediately show the auto-complete options, without having to start typing
 - When you choose an option, filtering should happen as before.
 - Then you can click again on the auto-complete input, ALL the options should be shown, allowing you to pick another without having to delete the text from the one you chose before.
 - So the input acts more like a drop-down chooser, as well as allowing you to enter any text (e.g. partial text to filter for many different values)

### DIFF
--- a/omeroweb/webclient/static/webclient/javascript/ome.center_plugin_filter.js
+++ b/omeroweb/webclient/static/webclient/javascript/ome.center_plugin_filter.js
@@ -65,6 +65,7 @@ function MapAnnFilter(image_ids, $element, callback, filterObjects) {
             var self = this;
             $(".filter_map_value", $filter).autocomplete({
                 disabled: false,
+                minLength: 0,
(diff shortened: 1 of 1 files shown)


pwalczysko:
Works as expected. Ready to merge fmpov

PR #180, opened 2020-06-24
Update jquery

will-moore:
Started following jQuery upgrade guide:
https://jquery.com/upgrade-guide/3.0/

Completed running of all robot tests (see https://github.com/ome/openmicroscopy/pull/6291) against merge-ci branch (containing this branch) and only 1 minor bug found, fixed in ed7f2f0 below.

PR #256, opened 2021-02-02
annotations api limit

will-moore:
This fixes an issue noticed during testing of #225.

With a large number of images selected e.g.
https://merge-ci.openmicroscopy.org/web/webclient/?show=dataset-13651
it is possible to get a large number of annotations:

Screenshot 2021-02-02 at 12 23 34

However, if the `Key-Value Pairs` tab is expanded here, we won't load all 2520 annotations, because the default limit is much lower at 500. So even if all the images have a particular Key-Value pair, you won't see them all.

Another example from IDR:
http://idr.openmicroscopy.org/webclient/?show=dataset-12101

Here, 231 images are selected (with 696 Key-Value pairs), but only 500 are loaded, so only 38 Gene annotations are loaded:
Screenshot 2021-02-02 at 12 32 09

With this higher limit, this bug should be much less likely.

### DIFF
--- a/omeroweb/webclient/views.py
+++ b/omeroweb/webclient/views.py
@@ -1307,7 +1307,7 @@ def api_annotations(request, conn=None, **kwargs):
     run_ids = get_list(request, "acquisition")
     well_ids = get_list(request, "well")
     page = get_long_or_default(request, "page", 1)
-    limit = get_long_or_default(request, "limit", settings.PAGE)
+    limit = get_long_or_default(request, "limit", 10000)
 
     ann_type = r.get("type", None)
     ns = r.get("ns", None)
--- a/omeroweb/webclient/views.py
+++ b/omeroweb/webclient/views.py
@@ -133,6 +133,10 @@ logger = logging.getLogger(__name__)
 
 logger.info("INIT '%s'" % os.getpid())
 
+# We want to allow a higher default limit for annotations so we can load
+# all the annotations expected for a PAGE of images
+ANNOTATIONS_LIMIT = settings.PAGE * 100
+
 
 def get_long_or_default(request, name, default):
     """
@@ -1307,7 +1311,7 @@ def api_annotations(request, conn=None, **kwargs):
     run_ids = get_list(request, "acquisition")
     well_ids = get_list(request, "well")
     page = get_long_or_default(request, "page", 1)
-    limit = get_long_or_default(request, "limit", 10000)
+    limit = get_long_or_default(request, "limit", ANNOTATIONS_LIMIT)
 
     ann_type = r.get("type", None)
     ns = r.get("ns", None)
@@ -2626,7 +2630,7 @@ def annotate_tags(request, conn=None, **kwargs):
         well_ids=selected["wells"],
         ann_type="tag",
         # If we reach this limit we'll get some tags not removed
-        limit=100000,
+        limit=ANNOTATIONS_LIMIT,
     )
 
     userMap = {}
--- a/omeroweb/feedback/views.py
+++ b/omeroweb/feedback/views.py
@@ -200,8 +200,7 @@ def handler404(request, exception=None):
         "Not Found: %s" % request.path, extra={"status_code": 404, "request": request}
     )
     if request.is_ajax():
-        msg = traceback.format_exception(*sys.exc_info())[-1]
-        return HttpResponseNotFound(msg)
+        return HttpResponseNotFound()
 
     return page_not_found(request, "404.html")
 
--- a/omeroweb/webclient/static/webclient/javascript/ome.right_panel_tags_pane.js
+++ b/omeroweb/webclient/static/webclient/javascript/ome.right_panel_tags_pane.js
@@ -125,8 +125,6 @@ var TagPane = function TagPane($element, opts) {
             request = request.join("&");
 
             var annsUrl = WEBCLIENT.URLS.webindex + "api/annotations/?type=tag&" + request
-            // set high limit on number of results (default is 200)
-            annsUrl += '&limit=10000'
             $.getJSON(annsUrl, function(data){
 
                 // manipulate data...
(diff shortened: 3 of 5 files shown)


joshmoore:
Less-likely, but possible, right? What would it take to loop through the various pages?

will-moore:
It's technically not too hard to load page 2 etc. 
But the UI usability would not be great with the current UI, so it would need some design work.
Simply pagination is not a great solution in most cases (e.g. centre panel of webclient). Only being able to see/filter the first n annotations, instead of a summary of all. It's even harder when you're trying to support editing.
It's probably best to wait until there is the need for it before building for a theoretical use-case.
Boosting the default limit is a quick fix that allows us to put that off for a bit.

jburel:
Maybe I am missing something
I selected the 500 images in the dataset mentioned above. I select KVP (3025) but I can only see 13.

will-moore:
The KVP are grouped together if they are identical. For example, the `foo: test` KVP has been added to all 500 images, resulting in 500 KVPs. But they are grouped together so that they can be edited together. 
If you mouse over the '500 objects' you'll see a tooltip that gives more details, including all the annotation IDs:

Screenshot 2021-02-03 at 17 35 00

jburel:
ok, so we need to add the value To XXX objects
A message under Add annotation to xx objects might help indicating that kvp are grouped
I missed it when expanding the tab

will-moore:
@jburel Any suggestions for what that might look like? I don't really want to make many/any UI changes in this PR as this was simply about fixing the limit bug.

jburel:
A message similar to the one indicating that the user cannot annotate.
Something like ``Key-value pairs are group together if they are identical. Check the value of objects annotated`` will already help

will-moore:
So, this message could be added to the "Add..." message at the top, where you're really saying what *will* happen? (and you don't need to say identical, since they will be identical:
```
Add annotations to 500 objects. 
Key-Value pairs will be grouped together.
```

Or does it go beside *each* grouped set of identical KVPs? The problem here is that there will be a lot of repetition in the UI. This might be better to have in the tooltip. 
NB: the tooltip above already has "You are editing 500 identical Key-Value pairs".

Maybe something more concise is possible. e.g:
```
Added by user 3
500 Key-Value Pairs linked to 500 objects
```
The user will know these are identical since we only show one Map Annotation (and the 'identical' is still in the tooltip).
However, we probably need to use the configured nomenclature for KVPs. In IDR they are called `Attributes`, so this would need to be:
```
Added by: public data
500 Attributes linked to 500 objects
```
However, if you see that written above:
```
Dilution factor                1: 5000
Retrieval method               HIER pH6
Human protein atlas version    v19
```
You might be confused because that looks like 3 attributes, not 500.

jburel:
I like the short message: People are already confused by the 2 terms we use i.e. KVP in UI, mapannotations in API and adding attributes in the mix does not really clarify things

pwalczysko:
> People are already confused by the 2 terms we use i.e. KVP in UI, mapannotations in API and adding attributes 

But unlike in vanilla OMERO, the IDR users know nothing about KVPs. The harmonica there is called "Attributes". I think the Attributes or Key-Value pairs word could be fetched as appropriate according to whether or not the KVP harmonica was renamed in that particular OMERO.web instance ?

jburel:
We could limit at ``500 annotations linked to 500 objects`` i.e. since we are in the KVP harmonica not need to indicate the type

will-moore:
That last commit uses that last suggestion `500 annotations linked to 500 objects`

Screenshot 2021-02-16 at 13 40 33

jburel:
The last commit matches what was discussed.
There is a check failure that needs to be looked at

jburel:
@will-moore could you have a look at the output of the ``Code scanning`` job?

omeroweb/webclient/views.py
```diff
@@ -1307,7 +1307,7 @@ def api_annotations(request, conn=None, **kwargs):
     run_ids = get_list(request, "acquisition")
     well_ids = get_list(request, "well")
     page = get_long_or_default(request, "page", 1)
-    limit = get_long_or_default(request, "limit", settings.PAGE)
+    limit = get_long_or_default(request, "limit", 10000)
```
joshmoore, line 1310: I'm a bit concerned about this [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants)

will-moore: @joshmoore I've made that 'magic number' a bit less magic (and replaced other similar numbers) in 766a269.
I *could* add another setting for this but am a bit reluctant to proliferate settings unless there's a real need to?

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#612](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/612/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/612/consoleText) for more details.
Possible conflicts:
    - omeroweb/webclient/static/webclient/javascript/ome.right_panel_tags_pane.js
  - PR #225 will-moore 'Query string ids'
    - omeroweb/webclient/static/webclient/javascript/ome.right_panel_tags_pane.js

--conflicts

joshmoore:
The refactoring definitely looks more findable. Thanks. Note this PR wasn't tested today due to conflicts.

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#613](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/613/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/613/consoleText) for more details.
Possible conflicts:
    - omeroweb/webclient/static/webclient/javascript/ome.right_panel_tags_pane.js
  - PR #225 will-moore 'Query string ids'
    - omeroweb/webclient/static/webclient/javascript/ome.right_panel_tags_pane.js

--conflicts

jburel:
Retested this morning.
Easier to read the annotations. When many objects are annotated, it is not really possible to view everything in the tooltip.
This strategy has its limits. Something to think about and outside the scope of this PR

PR #237, opened 2020-10-08
Pad the bottom of jstree

manics:
Closes https://github.com/ome/omero-web/issues/234, see issue for testing instructions.

Issue #234, opened 2020-10-06
Bottom of left-hard tree can be obsured by mouse-over link status bar


manics:
If you have a long list of containers in the tree it reaches all the way to the bottom of the browser view.
If you mouseover a container higher in the tree in Firefox the link target is shown in the bottom left of the browser. This may obscure the last study in the list.

In the following screenshot I've scrolled to the bottom of the LH tree, my mouse is hovering over `idr0082-pennycuick-lesions/experimentA` and the last study `idr0094-ellinger-sarscov2/screenB` is obscured.

<img width="369" alt="Screen Shot 2020-10-06 at 10 24 51" src="https://user-images.githubusercontent.com/1644105/95183887-83b83400-07be-11eb-8ad6-56cddc4f6724.png">

In order to get to the last study it's natural for the mouse to move over the items higher in the tree therefore displaying the link status, so this is a usability issue.

pwalczysko:
thanks @manics , I hit this problem very commonly in IDR too

manics:
Possible fix: Add `padding-bottom: 2em;` to the `.jstree-children` CSS class

PR #257, opened 2021-02-03
Jstree padding bottom

will-moore:
This replaces https://github.com/ome/omero-web/pull/237

EDIT: Instead of padding under the tree, we now use a `<span>` element instead of `<a>` for the jstree nodes, as discussed at https://github.com/vakata/jstree/pull/1616

The tree should look and behave the same, but nodes are no-longer links so we don't get the link popup when we mouse over.

### DIFF
--- a/omeroweb/webclient/static/webclient/css/layout.css
+++ b/omeroweb/webclient/static/webclient/css/layout.css
@@ -24,7 +24,7 @@
 .left_panel_content {
 	position: relative;
 	width: 100%;
-	height: 100%;
+	height: calc(100% - 20px);
 	overflow: hidden;
 }
 
--- a/omeroweb/webclient/static/webclient/css/layout.css
+++ b/omeroweb/webclient/static/webclient/css/layout.css
@@ -24,7 +24,7 @@
 .left_panel_content {
 	position: relative;
 	width: 100%;
-	height: calc(100% - 20px);
+	height: 100%;
 	overflow: hidden;
 }
 
--- a/omeroweb/webclient/static/webclient/javascript/jquery.jstree.childcount_plugin.js
+++ b/omeroweb/webclient/static/webclient/javascript/jquery.jstree.childcount_plugin.js
@@ -12,7 +12,7 @@
 
             if(obj) {
                 var node = inst.get_node(obj);
-                var anchor = $(obj).children('a.jstree-anchor');
+                var anchor = $(obj).children('.jstree-anchor');
 
                 // Add child count html
                 if (node.data !== undefined && node.data.obj.childCount > 0) {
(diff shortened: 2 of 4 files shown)


jburel:
I can see the difference between merge and release version.
Small improvement, maybe other users could be used to make the benefit of the change more obvious
Screenshot 2021-02-15 at 12 20 29
Screenshot 2021-02-15 at 12 20 49

jburel:
Discussed with @will-moore closing

joshmoore:
What was discussed? Wouldn't some improvement be better than none? (I regularly lose the last new IDR study in the UI)

will-moore:
This PR currently makes the UI look a bit broken:
It's not really clear where the bottom of the panel is or why that space is not being used or why items disappear before the bottom of the page. So as it stands that change generally makes the UI worse.

Screenshot 2021-02-16 at 08 47 50

So, it needs something else:
 - Put something else in that gap (different colour and blank, or some other placeholder)?
 - Fix jsTree so that it doesn't have ```<a>``` tags - some discussion of people doing this (see links from previous PR) but haven't looked at how much work or other issues with this.

We felt that this wasn't worth the time to fix a relatively minor issue.

will-moore:
NB: I think IDR is a slightly odd use-case where we are almost always most interested in the last item of the tree. But for most users, that is the "Orphaned Images" and so might not be so bad.
The fact that this issue has arisen "internally" and not as feedback from users was also a factor in the discussion.

joshmoore:
Is there an option to have this fix then at least when orphaned images is turned off.  cf. https://github.com/ome/omero-web/issues/234#issuecomment-704181932 For the IDR use case, this is quite significant. (But I do wonder if there are users who don't know there's an orphaned images folder hidden at the bottom)

jburel:
Padding was an option but when it is applied "blindly" is not great for the general case, it certainly improves the IDR case.
Since it was a change in the css, having the conf coming from the server will probably be tricky to apply so alternative needs to be looked into (jsTree as @will-moore mentioned for example)

joshmoore:
Happy to leave you guys to recommend a solution that either is IDR-only or works across the board. But currently we have three closed items (#257, #237, #234) and nothing on the radar for a fix.

jburel:
I will re-open so it is included in the build and people can look at it further and discuss option

jburel:
* url obstruction the view has now been removed. That url was not clickable/copiable.
* Cut/paste: count updated without refresh
* D&D: count updated without refresh
* Bug: I cannot D&D an image from a dataset into an empty dataset. This works on outreach
* Bug: I cannot D&D a dataset (empty or not) into a project. This works on outreach

jburel:
I have not noticed any problems with the tree
but it will be good to have a second pair of eyes
@pwalczysko could you have a look at it?

pwalczysko:
> I have not noticed any problems with the tree
> but it will be good to have a second pair of eyes
> @pwalczysko could you have a look at it?

Tested the whole right-click context menu in the LHP tree, checked the image/dataset counters, tried drag and drop, and also all the icons above the LHP.

All worked as expected. Ready to merge fmpov.

jburel:
Thanks @pwalczysko

PR #264, opened 2021-02-24
omero table link to ROI

will-moore:
Since `populate metadata` now supports ROI columns, and they will be used in idr0101, it is useful to be able to link from an ROI in omero_table to view the ROI in iviewer.

To test:
 - Need to create an OMERO.table with RoiColumn - see https://merge-ci.openmicroscopy.org/web/webclient/?show=image-39715
 - With iviewer installed, ROI IDs will become links to `/iviewer/?roi=ID` (as they already do for Images and Wells)

### DIFF
--- a/omeroweb/webclient/templates/webclient/annotations/omero_table.html
+++ b/omeroweb/webclient/templates/webclient/annotations/omero_table.html
@@ -99,6 +99,8 @@
                 <td><a target="_blank" href="{% url 'webindex' %}?show=image-{{ col }}">{{ col }}</a></td>
                 {% elif well_column_index == forloop.counter0 %}
                 <td><a target="_blank" href="{% url 'webindex' %}?show=well-{{ col }}">{{ col }}</a></td>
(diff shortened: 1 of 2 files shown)


sbesson:
Functionally, the proposed change works as expected. Tested with https://idr.openmicroscopy.org/search/?query=Name:idr0097-reicher-proteintag/experimentB which also includes some ROI tabular data. Clicking on a given ROI opens the region surrounding the detected tagged protein which is a nice workflow. Once this is merged and released, this is a feature worth deploying in IDR.

Screenshot 2021-02-25 at 16 04 00
Screenshot 2021-02-25 at 16 04 06

Only question is related to the `Potentially unsafe external link` warning raised by the CodeQL analysis - see https://github.com/ome/omero-web/pull/264/checks?check_run_id=1971135145. Is there some mechanism to whitelist the iviewer relative URL or it is worth adding the suggested meta tags?

will-moore:
I think we can just click on the `Dismiss > Won't fix` for those since they are known, internal URLs.

sbesson (approved):

PR #268, opened 2021-03-03
Getting range_size from kwargs negates the limit URL parameter

kkoz:
Bug fix for #243

### DIFF
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -2963,7 +2963,7 @@ def _table_query(request, fileid, conn=None, query=None, lazy=False, **kwargs):
                 else None
             )
         range_start = offset
-        range_size = kwargs.get("limit", rows)
+        range_size = limit if limit is not None else rows
         range_end = min(rows, range_start + range_size)
 
         if query == "*":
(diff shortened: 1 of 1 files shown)


will-moore:
Hi @kkoz, can you explain what bug is fixed by this and how to test? Thanks

omeroweb/webgateway/views.py
```diff
@@ -2963,7 +2963,7 @@ def _table_query(request, fileid, conn=None, query=None, lazy=False, **kwargs):
                 else None
             )
         range_start = offset
-        range_size = kwargs.get("limit", rows)
+        range_size = limit if limit is not None else rows
```
joshmoore, line 2966: Can this become `range_size = limit` and the setting to rows all go in the `if` block above?

kkoz:
@will-moore To test, simply query a table with the `limit` URL parameter set to several different things less than the total number of rows, e.g. `http://localhost/webgateway/table/3164/query/?query=*&limit=10`. In the current code it will have no effect. With the fix it should limit the number of rows returned.

jburel:
@will-moore did you have a chance to test this PR? so we can potentially include in the coming release

will-moore:
Working for me on merge-ci (and bug confirmed locally).
Code looks good.
Good to merge 👍

jburel:
Thanks @will-moore 
Could you add an entry in the changelog?

will-moore:
Done

jburel:
Thanks

Issue #167, opened 2020-04-29
Auto scroll to first well


dominikl:
Problem: If you open a plate with lots of empty wells around the edges, the plate can appear 'empty', especially on smaller displays. For example IDR0064, where wells start at C8:

![Screen Shot 2020-04-29 at 09 48 51](https://user-images.githubusercontent.com/6575139/80581182-b33f6500-8a04-11ea-818d-76d15afd0b01.png)

http://idr.openmicroscopy.org/webclient/?show=plate-7401

Maybe it would be possible to add some automatic scrolling to the first visible well in plate?

/cc @francesw @jburel

will-moore:
After the table html is generated and added to the page (https://github.com/ome/omero-web/blob/e5b30a74bab82dffe26a3df4cf6f57a104e9e723/omeroweb/webgateway/static/webgateway/js/ome.plateview.js#L163), should be possible to calculate the coordinates of the first Well and scroll the scrollable container https://github.com/ome/omero-web/blob/e5b30a74bab82dffe26a3df4cf6f57a104e9e723/omeroweb/webclient/templates/webclient/data/plate.html#L316

will-moore:
Will be fixed in https://github.com/ome/omero-web/pull/270

PR #260, opened 2021-02-15
Custom response headers

will-moore:
This replaces https://github.com/ome/omero-web/pull/143 (having merged in origin/master).

Add property to set custom headers in Django OMERO.web response

# What this PR does

This allows custom headers to be added to the HTTP response sent to the client by OMERO.web.
These header(s) will only be added if they're not already present in the response.

# Testing this PR

Set one or more custom headers, e.g.

```
omero config append omero.web.http_response_headers '["X-custom-header", "my value"]'
omero web start
```
Check the headers sent by OMERO.web, e.g. `curl -I localhost:4080`. You should see the custom headers you configured.

PR #270, opened 2021-03-04
constrain plate grid

### LINKED ISSUE #263: Hide empty wells
Request: Option to hide empty wells in the Plate view. See [Hide empty placeholders in Plate View in Omero.web](https://forum.image.sc/t/hide-empty-placeholders-in-plate-view-in-omero-web/49208) .

will-moore:
This PR has 2 types of changes:

# Layout changes
- This fixes #263 and is configurable - but the default behaviour has *not* changed.
- Use the new config `omero.web.plate_layout` to specify whether you want:
  - `shrink` Plate only extends from first to last Wells. Other rows/columns not shown
  - `trim` Plate extends from A1 to the last Well.
  - `expand` (default) retains the existing behaviour of expanding the grid to a factor of 12 x 8

# Other UI fixes
 - These are not configurable since they are really bug-fixes
 - When viewing a big plate (with scrollbars), when you scroll right and down, the row and column labels will remain stuck to top/left (always visible). 
 - Fixed scrolling to selected Well on loading. To test, select a Well towards bottom/right of Big plate. Copy the URL link to it and visit link in browser. When plate loads it should scroll to the selected Well
 - Thicker border for selected Well since it is currently very difficult to identify which Well is selected.
 - When zooming thumbnails, the plate should now stay centred on the same point. So if you're looking at a Well towards the bottom-right of the plate and you zoom in, it shouldn't fly out of the viewport. 

Screenshot shows:
 - Layout hasn't changed. The plate is expanded to a 384-well plate as before.
 - The URL is to a selected image (in the selected Well). This scrolls the plate to show the Well
 - The selected Well has a thicker border
 - The plate row and column names are still visible, even though we have scrolled the plate.

Screenshot 2021-04-30 at 09 53 07

EDIT: #167 not fixed by this PR.


snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#621](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/621/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/621/consoleText) for more details.
Possible conflicts:
  - PR #235 manics 'Allow a custom javascript tree key function'
    - omeroweb/settings.py
  - PR #260 will-moore 'Custom response headers'
    - omeroweb/settings.py

--conflicts

pwalczysko:
conflicts - not deployed on merge-ci

sbesson:
This PR seems to break the expectations from https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/700/testReport/junit/OmeroWeb.test.integration.test_plategrid/TestPlateGrid/test_get_plate_grid_metadata/. Should the previous default be preserved?

will-moore:
That's the question I was asking in standup last week. Do we want to change the default behaviour?
I would vote for yes - I don't see any advantage to expanding the plate, and it generally gives a poorer UI.
But happy to hear other opinions.
If we decide to change the default, the tests will need updating.

jburel:
To be clear, this PR won't be considered for the upcoming release.

sbesson:
Immediate consideration while looking at the original issue (6 wells in 384 well plates), what is the expected behavior if the wells are nota starting at A1 (e.g. consider [J-L][9-10] for the sake of discussion)? What is also the current behavior of Insight in terms of reducing the plate size grid on the left and top?

Is it worth getting back to the forum post and ask for screenshots/details of their use case? We definitely have examples with sparse plates in IDR where we could do some testing.

will-moore:
Yes, I think Insight has different handling of layout if you don't start at A1: truncating rows & columns *before* the first Well.
Whereas here we still show all the empty rows and columns before the first Well.
Having a lot of empty rows & columns at the start is a problem because we don't auto-scroll to the first Well. I think we have a feature request for that somewhere. One or other will be fine - I don't have a strong preference yet.

sbesson:
I'd definitely be inclined to loop other people in this discussion incl. @pwalczysko and @francesw who is dealing with this type of scenarios (e.g. https://idr.openmicroscopy.org/webclient/?show=screen-2603).

My 2p:

- in general, making this layout configurable certainly sounds like the easiest way forward also in terms of backwards-compatibility
- if we are supporting both plate layouts: `(0,0,round(well_x_max/12) * 12, round (well_y_max/8) * 8)` (current) and `(0, 0, well_x_max, well_y_max)`, adding a third layout `(well_x_min, well_y_min, well_x_max, well_y_max)` is easy enough in terms of code. It would mean use enums or values rather than a boolean.
- if trying to unify the behavior with Insight is in scope, would we consider turning this into a more generic `omero.client`  property?

will-moore:
Normal (new proposed default layout):

Screenshot 2021-03-09 at 11 46 35

With config:
```
$ omero config set omero.web.plate_layout "expand"
```

Screenshot 2021-03-09 at 11 50 58

```
$ omero config set omero.web.plate_layout "shrink"
```

Screenshot 2021-03-09 at 11 51 45

NB: I have also improved the scrolling of the plate so that the row and column labels stay stuck to the top/left.
And fixed the scroll on initial load to scroll the first Well into view.

I don't think we should try to replicate this in Insight.

joshmoore:
I like the "expand" / "shrink" terms. I'd just suggest coming up with a fitting term for the default in case we ever _change_ the default.

jburel:
Even if we decide not to propagate at this stage to insight, I will be more in favour of ``omero.client.xx`` properties instead of ``omero.web.xx`` We made the mistake with some properties containing "insight" in the name and use in web for example.

will-moore:
I'd prefer not to add more settings than necessary. Seems like counting in binary!
I don't think we need to support expanding bottom-right and shrinking top-left, but we *might* want to add other options.
We could rename `normal` to `regular` or `standard` etc?

Using `omero.client.xx` properties on the server makes the development a bit harder - They have to be added to a different repo - I can't even remember where now. And then you can't set different values on different `omero-web` installs for e.g. public and non-public.
Also, users might have the expectation that they would be supported in both clients.
If we don't have the intention to support in Insight now, I don't think we will in the future. No-one has requested this.

jburel:
I was thinking of other clients not necessarily developed by us so limiting it to wait to omero.web might to restrictive

will-moore:
It seems like a very edge-case use case where someone would:
 - Create an OMERO client that attempts to duplicate the webclient UI
 - Want to exactly follow the `plate_layout` setting to show the same layout

The extra dev effort doesn't seem worth it for a feature that no one has requested.
It also ties `omero-web` to the server release.

jburel:
a napari extension to display plate not only zarr plate?

sbesson:
> I'd prefer not to add more settings than necessary. Seems like counting in binary!

Fair enough, I can certainly live with a list of accepted terms if we decide on a shortlist.

> We could rename normal to regular or standard etc?

My primary issue with this terminology is that this PR actually implements a change in the OMERO.web standard layout for plates. So calling the new behavior `standard` would feel fairly confusing to me. Trying to come up with alternatives: `fit`, `tight`, `minimal`

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#626](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/626/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/626/consoleText) for more details.
Possible conflicts:
    - omeroweb/webclient/templates/webclient/data/plate.html

--conflicts

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#627](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/627/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/627/consoleText) for more details.
Possible conflicts:
    - omeroweb/webclient/templates/webclient/data/plate.html

--conflicts

will-moore:
@sbesson - Trying to think of better names, but struggling.
I think `shrink` and `expand` are good (I see that Josh agrees). So what comes in-between in terms of size, and doesn't involve "shrinking", "expanding" or really changing the size in any way - since if you've got a plate that has `C4` in it, then I think it will also have `A1` in it (even if the well is empty), without having to "expand".
So, if `standard` is no good, then I think `regular` is OK, since `shrink, regular, expand` is like `small, regular, large`.
I don't know that the naming *has* to reflect that there was a change. I wouldn't want the naming to be less clear forever into the future, just because the original behaviour was was once to `expand`. And if we changed the default to `shrink` or `expand` in the future, I think that `shrink, regular, expand` would still be valid?

ctrueden:
More words: trimmed; cropped to content.

Alternately, you could sidestep the naming conundrum here by calling the properties `omero.client.plate_layout.expand_top_left` and `omero.client.plate_layout.expand_bottom_right` with values `true` and `false` (or `yes` and `no`—whatever is consistent with other boolean properties).

will-moore:
NB: If this is going into a point release and we don't want to change the behaviour, we can leave the default to be "expand", then switch the default to e.g. "regular" at a later Major release if agreed?

joshmoore:
Guess I don't see anything "regular" about the current behavior. I could see "cropped" but often that's an "remove-from-all-sides" kind of thing thing, more like "shrink".

"trim" has a nice "remove-the-least-stuff" feel. (Thanks, @ctrueden) After just this discussion, I can see needing more properties in the future (:wink:) but I assume we don't want that complexity yet. Perhaps if/when that becomes, we could add `plate_layout=advanced` with related properties.

My vote would be: `shrink, trim (default), expand`

sbesson:
I'll leave others to assess whether the change in the default UI behavior is worth a major version increment. It would feel unfortunate to have to bump to omero.web 6 for this type of feature.

Looking forward to seeing this change in an upcoming release as I could see testing the `shrink` configuration in the context of the upcoming `idr0072` dataset, maybe via  a pre-release?

sbesson:
Tested in the context of IDR. The `idr0064` study e.g. https://idr.openmicroscopy.org/webclient/?show=screen-2751 are good examples of plates where the data appears far enough from  A1 that it can look missing depending on your screen resolution

Screenshot 2021-03-22 at 11 43 08

With this PR deployed, without any configuration change i.e. selecting `trim` by default, clicking on the plate first displays a grid starting at A1 with loading thumbnails

Screenshot 2021-03-22 at 11 42 08

then updates the view to have C8 in the top-left corner.

Screenshot 2021-03-22 at 11 42 10

Scrolling to the bottom right reveals, that G17 is now the last accessible well

Screenshot 2021-03-22 at 11 43 27

I will test the `shrink` layout as a follow-up but overall I think this provides an UI improvement for some of the sparse HCS datasets. Two immediate questions:

- the 2-step update of the grid (triggered by the loaded thumbnails?) feels a bit strange. Would it be possible to compute the first well position before drawing the grid and completely skip the first step?
- the horizontal and vertical lines delimiting the row and column names have disappeared. Is this by design (or related to another open PR)?

sbesson:
With `omero.web.plate_layout` set to `shrink`, I see the same UI behavior as for `trim`

Screenshot 2021-03-22 at 13 45 00

will-moore:
> the 2-step update of the grid (triggered by the loaded thumbnails?) feels a bit strange.

Yes - unfortunately we don't know the coordinates of the first well until the thumbnails have loaded because the thumbnail aspect ratio is not known.
I guess if we assume that all thumbnails are the same shape, then we just need to add that info to the plate metadata we're loading so that the JavaScript can access it. I'll see what this looks like... 

> the horizontal and vertical lines delimiting the row and column names have disappeared.

I'm not sure what you mean. For comparison, this is a current screenshot from IDR:
http://idr.openmicroscopy.org/webclient/?show=plate-7815

Screenshot 2021-03-22 at 14 37 36

will-moore:
Is there some similar data that I can test this with?
`shrink` is working for me locally. I can create some test data on `merge-ci` so we can see it working (or not) there too.
But there might be some difference between my test data and yours?

will-moore:
Test Plate created at https://merge-ci.openmicroscopy.org/web/webclient/?show=plate-33805

will-moore:
I've set `omero config set omero.web.plate_layout shrink` on merge-ci web deploy, so we can check after next deploy.

sbesson:
Retested with some other IDR data and I can now see the `shrink` effect

Screenshot 2021-03-22 at 14 57 56

I suspect the difference will appear when you load plates containing empty wells e.g. the synthetic wells created by https://github.com/IDR/idr0064-goglia-erkdynamics/blob/master/screens/1.1.screen for idr0064.

For sparse HCS data like single columns, this is what `shrink` looks like

Screenshot 2021-03-22 at 14 58 51

sbesson:
> I'm not sure what you mean. For comparison, this is a current screenshot from IDR:
http://idr.openmicroscopy.org/webclient/?show=plate-7815 

Taking this example plate without this PR:

Screenshot 2021-03-22 at 15 04 12

vs

Screenshot 2021-03-22 at 15 04 14

I am wondering about the border line left of column 1 and on top of row A. Very minor and I like the extra space between the column name and the wells.

will-moore:
@sbesson After more than an hour, I figured out how to preserve the borders of the row/column labels
The tricky was https://developer.mozilla.org/en-US/docs/Web/CSS/border-collapse

Screenshot 2021-03-22 at 17 54 07

However, that change has a small affect on the space around the 'selected' Well, so that it changes the size of the Well when it is selected, which causes all the Wells to the bottom/right of it to shift. This isn't very nice, so if I can't find a fix we might have to roll back.
But I also want to try and increase the width of the blue selected border, since I find it very hard to identify the selected Well. I gave up on this earlier: with rectangular Wells of unknown aspect ratio, it's very hard to make them shrink (to show a thicker border) without changing overall size. 
But, if I load the actual aspect ratio as mentioned above, this may help here too...

will-moore:
@sbesson I fixed the '2-step' update on loading. Now it should just instantly load with the correct layout and scroll position without waiting for thumbnails (see updated description).

sbesson:
@will-moore thanks. With the last set of commits, I confirm that the view immediately scrolls to the first populated well before starting the thumbnail loading.

While dealing with sparse plates using the default webclient view, this feels like a huge improvement to me as you maximize the usage of the center panel to display real data wherever possible instead of stub wells.  Taking `idr0064` as a representative example, this is the workflow in current IDR

recording

and with this PR included

recording (1)

As a side note, while reviewing the different plate layouts with @francesw in the context of the IDR screens, we found that the auto-scrolling was the key functionality here and the `expand` plate  layout still felt like the best UI option, as it reflects the fact the data comes from a subset of a 96, 384, 1536 wells plate.
There might be a case for keeping `expand` as the default value but I can see the unification with the default OMERO.insight UI as the driver so I am not strongly opinionated either way.

Including the Glencoe team in the loop of this proposed change to confirm there is no impact on their workflow as they are routinely dealing with HCS data including sparse layouts  @callan, @erindiel, @emilroz, @kkoz and @knabar

callan:
@sbesson perhaps you meant to ping @chris-allan? 😄

sbesson:
> @sbesson perhaps you meant to ping @chris-allan? 😄

Yes. Sorry about the spam. 🤦

will-moore:
With that last commit, I've found a way to switch back to using CSS to resize thumbnails when using the slider, so there shouldn't be any performance issues with the change. But a few pieces were changed (selection etc) so needs a good test

chris-allan:
I really like the options available here.

The only thing I'd be very careful of is making assumptions that instant image visibility is necessarily the most important workflow for all people or that a universal client wide setting is going to be sufficient. I certainly appreciate that the driver for this PR was an instance of direct user feedback. However, at the many deployments we have performed over many years and the 10s of thousands of plates that are imported into these systems we have not once heard a similar feedback.

Obviously, there are lots of different types of consumers of HCS data. Some are going to care deeply about and have workflows that depend on the physical plate layout, others are not going to care at all.

Before we go making changes to the defaults let's consider the implications on the potential silent majority.

/cc @emilroz, @erindiel

will-moore:
Apologies for dropping this. Based on the feedback from @sbesson and @chris-allan, I'm happy to revert the default plate layout so as to retain the current behaviour.
I'll re-list for testing since there's a couple of commits since it was last tested.
To test the latest changes, need to zoom thumbnails with the slider.

pwalczysko:
@will-moore 
Maybe there is a typo in the header of this PR ?
Screenshot 2021-04-27 at 10 53 18
Not so impotant about the typo, but is there a second issue which this PR is addressing ? If yes, which one ? Atm, the second link points to this very PR (circular behaviour)

will-moore:
Thanks @pwalczysko. Fixed - I meant #167.

pwalczysko:
The zooming in and scrolling and panning, as well as the well selection marker and the visibility of the column and row headers are improved and work as expected on merge-ci. Nevertheless, as discussed with @will-moore , there is a need to summarize the state of the PR, expected behaviour and whether the testing of the last commits is sufficient as done here by myself. This could go into the header of the PR, which is at present very outdated.

will-moore:
@sbesson - I think this is good to merge now. Let me know if you need anything else from me...

sbesson:
Apologies, I am afraid the merging of #284 creates some conflicts (probably on the docstrings), could you merge `origin/master` into this branch?

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#684](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/684/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/684/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - omeroweb/webgateway/views.py

--conflicts

sbesson:
Quickly retested with and without this PR using a plate with wells B6, B7, B8, E7, E8, E9.

- the default layout behavior remains to expand the layout to a 96 wells plate
- well/field of view have a thicker border when selected
- the column/row names are visible when scrolling through the plates
- when zooming in/out, the context is not lost when a well is selected.
- when selecting the plate, the scrollbar are automatically adjusted to have B6 in the top-left.

Regarding the last point, the ability to make the first well instantly visible had been discussed in the context of IDR sparse plates like https://idr.openmicroscopy.org/webclient/?show=plate-9462 and would certainly be beneficial to enable in an upcoming deployment.
Given the  concerns expressed by @chris-allan in https://github.com/ome/omero-web/pull/270#issuecomment-810083240 about client assumptions including instant image visibility, I would leave him to comment on whether such behavior change is acceptable of whether it should also be made configurable.

will-moore:
Just checking the scroll-to-first-Well behaviour on a few more Plates, I noticed a bug if the first Well isn't the left-most Well e.g. at https://merge-ci.openmicroscopy.org/web/webclient/?show=plate-23055 (user-3). Fixed in that last commit to make sure the horizontal scroll goes to the left-most Well, not the first Well in the Plate.

will-moore:
Re: scrolling to the first Wells when Plate loads - Currently this PR will scroll to the put the first Well in the top-left of the viewport, even if the Well wasn't originally outside the viewport. An alternative is to ONLY scroll to the first Well if it's not within the viewport (if NO Wells are visible when the Plate is loaded). 
That would mean that scrolling would happen less frequently, but would still fix #167 (which I think is a bug).
This might reduce any need to make the scrolling behaviour configurable? cc @chris-allan ?

chris-allan:
The change in 15b0af88bd5630cfdf6e9ff3c6cc07efade6d167 is definitely an improvement to the design in my opinion. Focusing on the first well vertically definitely has the problems you experienced when also worrying about the first well horizontally. The reverse is also true. Personally, I don't consider #167 to be a bug but rather a deliberate long standing design decision: to focus on display of the physical Plate layout. To say one Well is more important than another to display first just because it happens to have an Image present at the currently selected field seems like reaching to me.

As I outlined previously, I have real problems with changing the default behaviour. It's my opinion that we are just as likely to annoy people by scrolling the viewport as we are to appease them.

will-moore:
So, is the next step to make the scrolling configurable? If so, what is the default (to auto-scroll to show a Well or not)?
Part of my reluctance to add a config for everything is that they are not tested very well, but if we could configure IDR with `scroll = true` then at least we'd see any issues there.

will-moore:
@chris-allan I have removed the auto-scroll to first Well when plate loads (#167 is not "fixed" by this PR).
PR description updated.

Hopefully that was the last outstanding issue here and the rest of the fixes and improvements can be merged.

will-moore:
@chris-allan You OK with this PR now? It now *only* contains bug fixes. The change in scrolling was removed and plate layout is configurable but unchanged by default.

will-moore:
@chris-allan If you are OK with the changes here now, are you happy for others on the OME team to review, approve etc?

chris-allan:
Thanks, @will-moore. Great work on all the bug fixes and the new functionality. I have no concerns functionality wise with the PR as it stands.

sbesson (approved):
Retested the three proposed layouts

`expand` (default)
Screenshot 2021-09-02 at 16 09 11

`shrink`

Screenshot 2021-09-02 at 15 44 16

`trim`

Screenshot 2021-09-02 at 16 02 05

All works as expected. Also tested the usability fixes when selecting a well, zooming in and out and scrolling. All works as expected. Merging for inclusion in the pre-release

PR #276, opened 2021-03-10
Handle show object in invalid group

### LINKED ISSUE #236: Bug: webclient gets into endless loop because of group context
- Have a user (username "free") with some images in a group A. 
- the admin removes the "free" user from group A
- the user is not able to access their images in webclient anymore (expected)
- the user is able to find their images in group A still using the CLI command ``omero fs images``
- the user harvests the ID of the image found using ``omero fs images`` and goes ahead in the webclient to search for that image
- the search gets into an endless loop, where webclient is loading something in 1-2 sec intervals, then falls back onto the empty central pane

See video of this behaviour in  https://openmicroscopy.slack.com/archives/C0K5EED3R/p1602066661046100

cc @will-moore @mtbc @joshmoore

will-moore:
Fixes #236 

To test:
 - Remove a user from a group where they have some data e.g. a Project (remember it's ID)
 - Log them out, then log them in again, and try to load the project: `/webclient/?show=project-1`
 - The webclient should load showing their default group (but not finding any Project) - instead of recursively re-loading the page.

### DIFF
--- a/omeroweb/webclient/views.py
+++ b/omeroweb/webclient/views.py
@@ -446,7 +446,11 @@ def _load_template(request, menu, conn=None, url=None, **kwargs):
 
     # need to be sure that tree will be correct omero.group
     if first_sel is not None:
-        switch_active_group(request, first_sel.details.group.id.val)
+        group_id = first_sel.details.group.id.val
+        if conn.isValidGroup(group_id):
+            switch_active_group(request, group_id)
+        else:
+            first_sel = None
 
     # search support
     init = {}


pwalczysko:
I have the behaviour repeated now without this PR on https://merge-ci.openmicroscopy.org/web/webclient/?show=project-33957, login as import_test_user.

Will check when this PR is in the build.

pwalczysko:
this works on merge-ci now as expected. Logging in and out as different users, going via Search or just clicking on the link had the expected effect of landing in the default group of the user in question. No bliniking, the public user could be navigated too, and the login button worked also for other users.

Ready to merge fmpov

pwalczysko (approved):

PR #277, opened 2021-03-17
Table query case insensitive

will-moore:
See discussion at https://github.com/ome/omero-metadata/pull/56#issuecomment-801030871
Currently, we don't know whether an OMERO.table column will be named 'Image' or 'image'.
So we have to query for BOTH (in series, since parallel queries lead to table.close() race conditions).

As @sbesson suggested, this logic can be simplified if we handle case mismatches in column names.
In this PR, if you query a table for `Image, Roi, Plate or Well` and the table doesn't have a matching
column name but *does* contain the column name in lower case, e.g. `image`, then we
simply use the lower case for the query.

This means we don't need duplicate queries from the UI (see the JavaScript code is much simpler).

To test:
 - See https://merge-ci.openmicroscopy.org/web/webclient/?show=project-857 (user-3)
 - The most recent table on this Project (which is always picked for queries) has the Image column named 'image': https://merge-ci.openmicroscopy.org/web/webclient/omero_table/634804/
 - Images named `img-xx.png` in that Project should display the Tables data correctly in the right-panel
 - The URL https://merge-ci.openmicroscopy.org/web/webgateway/table/Project.datasetLinks.child.imageLinks.child/36638/query/?query=Image-36638 (using query `Image-36638`) should return results for the table, even though the column names in the JSON data will be "image".

### DIFF
--- a/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
+++ b/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
@@ -409,17 +409,11 @@
                                 loadBulkAnnotations(projectQuery, query, function(data) {
                                     if (data && data.id) {
                                         showBulkAnnTooltip(data);
-                                    } else {
-                                        // If query1 didn't work, try query2 on the same OMERO.table
-                                        loadBulkAnnotations(projectQuery, query2, showBulkAnnTooltip);
                                     }
                                 });
                                 loadBulkAnnotations(datasetQuery, query, function(data) {
                                     if (data && data.id) {
                                         showBulkAnnTooltip(data);
-                                    } else {
-                                        // If query1 didn't work, try query2 on the same OMERO.table
-                                        loadBulkAnnotations(datasetQuery, query2, showBulkAnnTooltip);
                                     }
                                 });
                             {% endif %}
--- a/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
+++ b/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
@@ -352,7 +352,6 @@
                     var projectQuery = "{% url 'webgateway_object_table_query' 'Project.datasetLinks.child.imageLinks.child' manager.image.id %}";
                     var datasetQuery = "{% url 'webgateway_object_table_query' 'Dataset.imageLinks.child' manager.image.id %}";
                     var query = "Image-{{ manager.image.id }}";
-                    var query2 = "image-{{ manager.image.id }}";
                     {% endif %}
 
                     var showBulkAnnTooltip = function(data) {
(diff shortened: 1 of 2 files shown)


will-moore:
Re: https://github.com/ome/omero-metadata/pull/56#issuecomment-804980465
I *could* also support `?query=ROI-123` here, but the logic would be slightly different from the current logic:
`if "ROI" is not in column_names, then check if "Roi" is in column_names, if not then check if "roi" is in column_names`.
Since we don't query by `ROI` anywhere, this doesn't help simplify the code and rather makes the behaviour less predictable. E.g. if `ROI` is supported, then why not `IMAGE`?

We could also make queries case insensitive for ALL column names (not just `Image, Well, Roi, Plate`). This wouldn't hurt performance, but again, I would prefer if the `?query=...` API is as close as possible to the underlying table query API.

omeroweb/webclient/templates/webclient/annotations/metadata_general.html
```diff
@@ -392,24 +391,8 @@
...
-                                    } else {
-                                        // If query1 didn't work, try query2 on the same OMERO.table
-                                        loadBulkAnnotations(datasetQuery, query2, showBulkAnnTooltip);
-                                    }
-                                });
+                                loadBulkAnnotations(projectQuery, query, showBulkAnnTooltip);
+                                loadBulkAnnotations(datasetQuery, query, showBulkAnnTooltip);
                             {% endif %}
                         }
                     });
```
sbesson, line 398: Should the block below be updated accordingly to remove the conditional statements?

omeroweb/webgateway/views.py
```diff
@@ -2972,7 +2973,12 @@ def _table_query(request, fileid, conn=None, query=None, lazy=False, **kwargs):
         else:
             match = re.match(r"^(\w+)-(\d+)", query)
             if match:
-                query = "(%s==%s)" % (match.group(1), match.group(2))
+                c_name = match.group(1)
+                if c_name in ("Image", "Roi", "Plate", "Well"):
```
sbesson, line 2977: If we wanted to handle all case-insensitive variations of these reserved column names, this could become 

```
if c_name not in columns_names and c_name.lower() in ("image", "roi", "plate", "well") and c_name.lower() in column_names:
```

I don't know of a concrete use case for now so unless you feel strongly about it, happy to start with the variants for our own tools i.e. `image/Image` for now and adjust as new use cases arise.

sbesson (commented):
Tested the query API change using https://idr.openmicroscopy.org/webclient/omero_table/15780187 (`idr0041`) which is a reasonably large OMERO.table with the `image` column names. Using production IDR,

```
[sbesson@prod95-omeroreadonly-1 ~]$ time curl http://localhost/webclient/omero_table/15780187/?query=image-3431104
...
                <td><a target="_blank" href="/webclient/?show=image-3431104">3431104</a></td>
                
            
        </tr>
        
    </body>
</html>

real	0m0.158s
user	0m0.000s
sys	0m0.010s
[sbesson@prod95-omeroreadonly-1 ~]$ time curl http://localhost/webclient/omero_table/15780187/?query=Image-3431104
{"error": "Error executing query: (Image==3431104)"}
real	0m0.111s
user	0m0.004s
sys	0m0.006s
```

With this PR included in the pilot server

```
[sbesson@pilot-idr0072-omeroreadwrite ~]$ time curl http://localhost/webclient/omero_table/15780187/?query=image-3431104
...
                <td><a target="_blank" href="/webclient/?show=image-3431104">3431104</a></td>
                
            
        </tr>
        
    </body>
</html>

real	0m0.154s
user	0m0.003s
sys	0m0.007s
[sbesson@pilot-idr0072-omeroreadwrite ~]$ time curl http://localhost/webclient/omero_table/15780187/?query=image-3431104
...
                <td><a target="_blank" href="/webclient/?show=image-3431104">3431104</a></td>
                
            
        </tr>
        
    </body>
</html>

real	0m0.152s
user	0m0.006s
sys	0m0.007s
```

In terms of code changes, most of the logic happens during the translation of our  `<name>-<value>` into a case sensitive `<column_name>==<value>` query so there should be minimal performance impact. This also greatly reduces the complexity of the table loading in `metadata_general.html`.

In terms of documentation, it might make sense to update the docstring of `_table_query` to capture the new contract for this specific query variant. Is there another reference documentation place that describes the expectation for `?query=<name>-<value>`?
In terms of test coverage, I could imagine a handful of test additions to https://github.com/ome/openmicroscopy/blob/develop/components/tools/OmeroWeb/test/integration/test_table.py covering the various scenarios would be useful to catch potential regressions.

Generally, given the existing variations in the column names of OMERO.tables, having some API allowing to handle synonyms for some of the reserved columns while preserving the underlying case-sensitive query API is a good balance.

omeroweb/webclient/templates/webclient/annotations/metadata_general.html
will-moore: Good catch. Removed 'query2' calls below.

omeroweb/webgateway/views.py
will-moore: I think it's simpler to stick with `image/Image` instead of supporting all upper/lower case variants.

will-moore:
To test last commit, you need to open the dev tools and check the Console when opening the `Tables` tab for an Image that doesn't have any Tables data. Shouldn't see any errors in the Console.

PR #282, opened 2021-04-01
mapr object id fix

will-moore:
This fixes 2 OMERO.mapr bugs reported as the "second issue" at https://forum.image.sc/t/omero-mapr-questions/50843

The first bug is a long-standing issue with Mapr: https://github.com/ome/omero-mapr/issues/67

To test:
 - Click on the top-level item in the Tree for a mapr search (as in the screenshot at https://github.com/ome/omero-mapr/issues/67). 
 - In the right panel, the annotations should load (instead of permanently saying "Loading key value annotations"

The other was caused by https://github.com/ome/omero-web/pull/232 (omero-web 5.9.0) since it assumes that Object IDs are numbers and can be rendered in Django to a valid JavaScript value.
But when you click on a mapr value in the tree (e.g. `AB_528479` in the screenshot below) then this String is the `obj_id` and causes a JavaScript error when rendered in the Django template.

This JS error shows up when you browse down to an Image and find that the Map Annotations tab won't load (as reported above).

Screenshot 2021-04-01 at 10 29 41
the right panel omapr 

To test:
 - Search in mapr for a Value, browse down the Tree to Project/Dataset/Image etc and check that the Map Annotations load OK
 - If you check the JS console, you won't see the `Uncaught Reference Error` (first error in the console screenshot reported above) but you will still see another error that has been highlighted by that report - created https://github.com/ome/omero-mapr/issues/67
 - Also confirm that the right-panel Open-with menu is working with the various apps/viewers

### DIFF
--- a/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
+++ b/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
@@ -58,7 +58,7 @@
                 // If single object selected
               {% if manager.obj_type %}
                 selectedObjs = ["{{ manager.obj_type }}-{{ manager.obj_id }}"];
-                selJson = [{'id': {{ manager.obj_id }},
+                selJson = [{'id': '{{ manager.obj_id }}',
                            'name': '{{ manager.obj_name }}',
                            'type': '{{ manager.obj_type }}' }]
               {% endif %}
--- a/omeroweb/webclient/templates/webclient/annotations/mapanns_underscore.html
+++ b/omeroweb/webclient/templates/webclient/annotations/mapanns_underscore.html
@@ -37,7 +37,7 @@
                       <% print (ann.permissions.canEdit && clientMapAnn ? 'editing' : 'viewing') %>
                       <b><%= ann.parentNames.length %></b> identical Key-Value annotations:
                     <% } %>
-                    <% if (!ann.parentNames) { %>
+                    <% if (!ann.parentNames && ann.link) { %>
                         <!-- If single object show e.g. Image ID: (slice ImageI -> Image) -->
                         <b><%= ann.link.parent.class.slice(0, ann.link.parent.class.length-1) %>
                             ID:</b> <%= ann.link.parent.id %><br />


joshmoore:
Certainly don't object to the safety check, but it'd be good to understand why the parent is unloaded.

sbesson:
Tested on a pilot IDR resource. The first issue can easily be reproduce  in current production IDR and is fixed by this PR with the map annotations loaded on the right-hand panel.

Screenshot 2021-04-07 at 17 32 40

> Certainly don't object to the safety check, but it'd be good to understand why the parent is unloaded.

My understanding is that the main issue is  the absence of a `link` rather than unloaded object. Is it correct to assume that `ann.link` is context-dependent i.e. in an object-centric navigation a parent object should be expected but in the mapr UI, map annotations are loaded without any parent context?

I wasn't able to reproduce the second JS error so I might need some additional guidance here to confirm the PR fixes the additional issue.

I was

will-moore:
@sbesson Apologies for dropping this...
If you go to https://latest-ci.openmicroscopy.org/web/mapr/anyvalue/?value=HSE (user-3) and click on the `HSE` in the tree, you'll see in the Console `Uncaught ReferenceError: HSE is not defined`.
Then, when you browse the hierarchy down to an Image, the Map Annotations won't be shown.

will-moore:
Actually, this bug can be seen in IDR. E.g.
http://idr.openmicroscopy.org/mapr/gene/?value=INCENP
Click on INCENP in the tree - see error in console.
Then browse down to an Image and the Map Annotations won't load.

sbesson (approved):
Without this PR, using production IDR with the test the link above, I can reproduce the two errors from the description in the Developer Console Tools:

Screenshot 2021-05-14 at 09 07 24

With this PR deployed in a pilot instance, none of the errors above are raised with the same workflow

Screenshot 2021-05-14 at 09 07 27

Good to merge from my point of view.
Only question is whether something else needs to happen in order to address the concerns in https://github.com/ome/omero-web/pull/282#issuecomment-814776615.

will-moore:
Re: https://github.com/ome/omero-web/pull/282#issuecomment-814776615 I think this is mapr re-using the webclient's display of Map-Annotations linked to the selected Object, instead displaying a list of Map-Annotations corresponding to the selected tree node (not linked to any specified objects). So mapr doesn't have any `links` which the webclient is expecting.

We could fix the issue in mapr itself, by passing placeholder `link` with e.g. empty strings. But this seems like a nicer solution.

jburel:
@will-moore do you want to open an issue for mapr? so we can go back to it if required.

will-moore:
@jburel I created an issue for this earlier at https://github.com/ome/omero-mapr/issues/67

jburel:
Thanks for linking the 2.

PR #283, opened 2021-04-13
Add user table metadata to metadata endpoint

kkoz:
It is possible to add custom user metadata to an Omero table using the `setMetadata` or `setAllMetadata` methods (see https://docs.openmicroscopy.org/omero/5.6.1/developers/Tables.html#omero.grid.Table.setMetadata). We would like this to be added to the response of the get table metadata endpoint. To test, add some metadata to an existing table and then request the metadata for that table (`/webgateway/table/<File ID>/metadata/`). The response should contain the metadata added under `user_metadata`.

### DIFF
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -3042,7 +3042,18 @@ def _table_metadata(request, fileid, conn=None, query=None, lazy=False, **kwargs
     try:
         cols = t.getHeaders()
         rows = t.getNumberOfRows()
+        allmeta = t.getAllMetadata()
 
+        user_metadata = {}
+        for k in allmeta:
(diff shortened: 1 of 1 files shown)


will-moore:
Apologies for not testing this earlier...
When adding metadata with:
```
table.setMetadata("test", omero.rtypes.rstring("value"))
table.setMetadata("my_number", omero.rtypes.rint(100))
```
I can confirm that I see JSON at `/webgateway/table/<File ID>/metadata/` like:
```
"user_metadata": {
"__version": 2,
"test": "value",
"my_number": 100,
"__initialized": 1620922240
}
```

will-moore:
I added a test for this URL in https://github.com/ome/openmicroscopy/pull/6279

PR #291, opened 2021-05-21
Add support for /shape/ID/

### LINKED ISSUE #290: JSON API Shapes support
Add support for  `/shapes/ID/` (and probably `/shapes/` too)

I wanted this in order to be able to get the ROI ID from a Shape ID.

will-moore:
Fixes #290

This adds support for `/api/v0/m/shapes/[ID]/`
and `/api/v0/m/shapes/`.

To test:

 - Go to `/api/v0/m/shapes/`. Check JSON returned OK.
 - Check `/api/v0/m/shapes/ID/` for a particular Shape. This should include `"url:roi"` with a URL to the parent ROI.
 
Tests added in https://github.com/ome/openmicroscopy/pull/6281/

### DIFF
--- a/omeroweb/api/urls.py
+++ b/omeroweb/api/urls.py
@@ -316,12 +316,21 @@ ROI url to GET or DELETE a single ROI
 """
 
 api_image_rois = url(
-    r"^v(?P<api_version>%s)/m/images/" "(?P<image_id>[0-9]+)/rois/$" % versions,
+    r"^v(?P<api_version>%s)/m/images/(?P<image_id>[0-9]+)/rois/$" % versions,
     views.RoisView.as_view(),
     name="api_image_rois",
 )
 """
 GET ROIs that belong to an Image, using omero-marshal to generate json
+"""
+
+api_shape = url(
+    r"^v(?P<api_version>%s)/m/shapes/(?P<object_id>[0-9]+)/$" % versions,
+    views.ShapeView.as_view(),
+    name="api_shape",
+)
+"""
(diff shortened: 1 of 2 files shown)


jburel:
I am a bit concerned by the call ``api/v0/m/shapes/`` this will load all the shapes. This could a huge amount of elements.

jburel:
This PR also modifies ``api_image_rois`` call. This is not mentioned in the description

will-moore:
The `api_image_rois` edit is simply cleaning up extra whitespace from code formatting. It has no effect on the code.
E.g. https://merge-ci.openmicroscopy.org/web/api/v0/m/images/83221/rois/ (user-3) is unchanged.

For all types supported by the JSON api, we have URLs that list 'all' objects but this is paginated to avoid loading too many objects. By default (e.g. https://merge-ci.openmicroscopy.org/web/api/v0/m/shapes/) this loads 200 items.
The number can be controlled by `?limit=500` but the `maxLimit` is defined by config (default is 500) so even https://merge-ci.openmicroscopy.org/web/api/v0/m/shapes/?limit=1000 will only load 500 items.

will-moore:
Not sure if this has had manual testing yet?
Test added in PR: https://github.com/ome/openmicroscopy/pull/6281

jburel:
Tested the loading of ROIs and shapes via the JSON API.
Having the ability to load a given shape will allow to for example to link a value from OMERO.table to iviewer using ``iviewer/?shape=99741``

PR #301, opened 2021-06-25
table_errors status_codes

will-moore:
See discussion at https://github.com/ome/omero-web/pull/300#discussion_r657133704

This aims to return appropriate `status` codes for Json Responses for the `_table` end-points:

`/webclient/omero_table/50534/json/`
`/webgateway/table/Screen.plateLinks.child.wells.wellSamples.image/806/query/?query=Image-806`
`/webgateway/table/Dataset.imageLinks.child/806/query/?query=*`
`/webgateway/table/5/metadata/`
`/webgateway/table/50534/query/?query=Image>1211`

However, this breaks the webclient, since the [global ajaxError handling](https://github.com/ome/omero-web/blob/b25ce9e19a6f55f6b7f71b582bf211b15eb5abd6/omeroweb/webgateway/static/webgateway/js/ome.popup.js#L517) shows an error message for any 404s, 500s etc instead of simply ignoring them as before:

Screenshot 2021-06-25 at 10 25 56

cc @ chris-allan @kkoz 

--exclude

### DIFF
--- a/omeroweb/webclient/views.py
+++ b/omeroweb/webclient/views.py
@@ -3191,7 +3191,7 @@ def omero_table(request, file_id, mtype=None, conn=None, **kwargs):
     )
 
     if context.get("error") or not context.get("data"):
-        return JsonResponse(context)
+        return JsonResponse(context, status=context.get("status", 500))
 
     # OR, return as csv or html
     if mtype == "csv":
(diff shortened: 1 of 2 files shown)


sbesson:
As mentioned in https://github.com/ome/omero-py/pull/292#issuecomment-869501106, this PR causes failures in the integration tests. Also trying to access the endpoint `webgateway/table/<table_id>/metadata` endpoint directly leads to the following error message `"'dict' object has no att…'createServiceOptsDict'"`.

PR #310, opened 2021-08-10
omero table link to shape

will-moore:
Follow-up of #264, this PR links Shape IDs in an OMERO.table to `iviewer?shape=ID`.

However, in this case we don't have a `ShapeColumn` to check for, so I've just looked for a column where `name.lower() == "shape"` and type is `LongColumn`.

This is useful for cases like `idr0101` where you have many Shapes per ROI, so that linking to the ROI is not sufficient.

To test - see example Image with bulk_annotations table that has `shape`: https://merge-ci.openmicroscopy.org/web/webclient/?show=image-94481
NB: that image doesn't have correct dimensions for the ROI - it's a fake image with ROIs from idr0101 applied to it, so selection of the expected Shape may not happen if the z-index is greater than sizeZ for the image.

 - Table values in the `shape` column should be links to open iviewer?shape=ID` in new tab

cc @sbesson

### DIFF
--- a/omeroweb/webclient/templates/webclient/annotations/omero_table.html
+++ b/omeroweb/webclient/templates/webclient/annotations/omero_table.html
@@ -101,6 +101,8 @@
                 <td><a target="_blank" href="{% url 'webindex' %}?show=well-{{ col }}">{{ col }}</a></td>
                 {% elif roi_column_index == forloop.counter0 and iviewer_url != None %}
                 <td><a target="_blank" href="{{ iviewer_url }}?roi={{ col }}">{{ col }}</a></td>
+                {% elif shape_column_index == forloop.counter0 and iviewer_url != None %}
+                <td><a target="_blank" href="{{ iviewer_url }}?shape={{ col }}">{{ col }}</a></td>
                 {% else %}
                 <td>{{ col }}</td>
                 {% endif %}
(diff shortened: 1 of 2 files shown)


jburel:
For ``idr0103``, I had multiple columns with shape Ids/roi ids (3 +RoiColumn) so the strategy of naming one column "shape" does not really scale. The proposed change is highly tailored to a specific need.

sbesson:
I'd also agree that if there was a ShapeColumn in the OMERO.tables, the natural support in the `omero-metadata` plugin would be to map  columns either named `shape` or annotated with the `shape` header. A difference though is that such name-based assumptions are only used when translating CSV into OMERO.tables while the OMERO.web table API relies on the column type and this PR effectively creates a precedent. Will the special handling (name='shape' & type='long') be supported even if we add a `ShapeColumn` type in the future?

will-moore:
If all existing tables with `shape` (long) columns were updated to `ShapeColumn` then we wouldn't need to continue to support it here. But otherwise, yes.

Although in this case, the `RoiColumn`, `WellColumn` or `ImageColumn` are used when creating links, elsewhere the webclient relies on column naming. E.g. when loading table data for an `Image` in the right panel, this uses column name `query=Image-123`, not column type.

jburel:
The functionality works as expected.
Discussed with @will-moore this morning about the current limitation of the ROIs support in iviewer.
this PR highlights a loading issue that @will-moore previously attempted to fix (see https://github.com/ome/omero-iviewer/pull/351)

PR #311, opened 2021-08-24
Add omero.web config option to set custom favicon

emilroz:
This PR introduces a config option to overwrite `webgateway/img/ome.ico` with a custom icon.

Favicon can be overwritten by setting `favicon_url` option with:

```
omero config set omero.web.favicon_url 'omero_web_module_name/subdir/new.ico'
```

The URL to the favicon needs to be specified relative to the Django's static directory.

### DIFF
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -834,9 +834,7 @@ CUSTOM_SETTINGS_MAPPINGS = {
         "FAVICON_URL",
         '"webgateway/img/ome.ico"',
         json.loads,
-        (
-            "Favicon URL, specifies the path relative to django's static file dirs."
-        ),
+        ("Favicon URL, specifies the path relative to django's static file dirs."),
     ],
     "omero.web.staticfile_dirs": [
         "STATICFILES_DIRS",
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -833,7 +833,7 @@ CUSTOM_SETTINGS_MAPPINGS = {
     "omero.web.favicon_url": [
         "FAVICON_URL",
         '"webgateway/img/ome.ico"',
-        json.loads,
+        str,
         ("Favicon URL, specifies the path relative to django's static file dirs."),
     ],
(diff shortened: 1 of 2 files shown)


omeroweb/urls.py
```diff
@@ -118,7 +118,7 @@ def redirect_urlpatterns():
 urlpatterns += [
     url(
         r"^favicon\.ico$",
-        lambda request: redirect("%swebgateway/img/ome.ico" % settings.STATIC_URL),
+        lambda request: redirect("%s%s" % (settings.STATIC_URL, settings.FAVICON_URL)),
```
joshmoore, line 121: Would an absolute FAVICON_URL ever be in scope?

emilroz: That would require additional nginx/webserver config if you want to do it yourself right?

will-moore: @joshmoore You mean something equivalent to the `top_links` logic where we handle local vv absolute links:
```
try:
    link["link"] = reverse(link_id)
except NoReverseMatch:
    # assume we've been passed a url
    link["link"] = link_id
```

I guess for `favicon_url` you could test `url.startswith("http")` or something like that?

omeroweb/settings.py
```diff
@@ -830,6 +830,12 @@ def leave_none_unset_int(s):
             "(OME team by default)."
         ),
     ],
+    "omero.web.favicon_url": [
+        "FAVICON_URL",
+        '"webgateway/img/ome.ico"',
+        json.loads,
```
chris-allan, line 836: Should this not just be `str`? Is there a reason for this to be JSON encoded?

emilroz: Fixed.

chris-allan: You need to loose the quotes in your example in the PR description and the default now I believe.

emilroz: Thanks, done.

will-moore:
This works fine for me testing locally and code looks good. 👍 
Good to merge.

PR #312, opened 2021-08-24
Respect login view for public login button

emilroz:
The changes in this PR pass the Login View setting to the OMERO.web "Login" button when public user is enabled.

image


will-moore:
Testing locally (just picking "history" as a view name that needs no parameters):
```
omero config set omero.web.login_view history
```
and with public user enabled the `Login` button has the `history` URL as expected.

Looks good. 👍

PR #317, opened 2021-09-06
Upgrade Underscore to 1.13.1

chris-allan:
Upgrade Underscore to the latest 1.13.1 version. Current use is unaffected by the changelog and isolated to the following functions:

 * _.each
 * _.template
 * _.uniq
 * _.escape
 * _.debounce


will-moore:
Testing on merge-ci as well as merged locally, testing all the right-hand annotation panels where this is used extensively reveals no issues. LGTM 👍

jburel:
Thanks, Merging so it is included in the 5.10.0 rc

PR #328, opened 2021-11-08
saveToAll includes Z and T

will-moore:
See https://forum.image.sc/t/updating-plate-thumbnails-in-omero-web/59729

To test:

 - Need a Plate or a Dataset with multi-T or multi-Z images (probably needs images with the same sizeT or sizeZ).
 - View one Image in the Preview Tab. Adjust settings, including Z or T slider.
 - Click "SaveToAll"
 - Check that the Z or T index is saved to that Image AND applied to all the other Images in the Dataset or Plate.


jburel (approved):
Tested with a dataset and a plate (user-1, test_pr_328)
Everything works as expected

PR #331, opened 2021-11-12
group stats filter

will-moore:
Fixes bug reported at https://forum.image.sc/t/few-users-mistakenly-appear-in-the-statistics-of-all-groups/59932

To test:
 - Log in as a group owner. Under User Settings, look at the data usage for a group.
 - Shouldn't see any users listed who are not members of the group *unless* they have more than 10MB
 - Try removing someone from the group who has more than 10MB data, will see:

Screenshot 2021-11-12 at 12 52 10

The query loads all Pixels and Files for *every* Experimenter in the DB, filtering by the current group context. This includes the User's Avatar photo which is in the "user" group (not filtered out by group context).
A much faster query is to only query the Experimenters who are members of the Group, but then we'd miss any user who has a lot of data in the group (but is not a member) e.g. previous member.
NB: the filtering of 10MB limit is done in JavaScript. All the query results are available in the JSON data.

Also update the flot library as I was getting poor layout with long user names.


joshmoore:
> I can't think how these log files seem to belong to a group that the owner never belonged to?

Can you point me at an example?

will-moore:
@joshmoore A bunch of examples locally are from the User Photo (avatar) uploaded with `adminService.uploadMyUserPhoto()` which apparently sets the group as "user" group so as to be available across all groups. So I guess these aren't filtered out with a group context.

joshmoore:
Yes, those would be accessible to everyone across groups.

pwalczysko:
> Yes, those would be accessible to everyone across groups.

Discarding such files based on the abovementioned quality would seem to me logical. The workaround here (discarding later, and only based on overall filesizes) is much less granular and targeted than seems reasonable imho, at least on the first naive look of course.

will-moore:
Thanks @pwalczysko - Now filtering explicitly for data that is in the group (not across all groups).

pwalczysko:
Thank you , I have relisted it for tomorrow, can have a look then.

omeroweb/webadmin/templates/webadmin/myaccount.html
```diff
@@ -39,11 +39,9 @@
     <script type="text/javascript" src="{% static "3rdparty/jquery.tablesorter-2.0.3/jquery.tablesorter.js" %}"></script>
     <script type="text/javascript" src="{% static "3rdparty/jquery.quicksearch-1.0.js" %}"></script>
     <script type="text/javascript" src="{% static "3rdparty/jquery.form-4.3.0.js" %}"></script>
-    <script type="text/javascript" src="{% static "3rdparty/jquery.flot-0.7/jquery.flot.js" %}"></script>
-    <script type="text/javascript" src="{% static "3rdparty/jquery.flot-0.7/jquery.flot.pie.js" %}"></script>
-    
-    <!--[if lte IE 8]><script language="javascript" type="text/javascript" src="{% static "3rdparty/jquery.flot-0.7/excanvas.js" %}"></script><![endif]-->
```
joshmoore, line 45: Does this need testing on IE then?

omeroweb/webadmin/views.py
```diff
@@ -256,18 +256,28 @@ def getBytes(ctx, eid=None):
             "from Pixels p join p.pixelsType as pt join p.image i "
             "left outer join i.fileset f "
             "join p.details.owner as owner "
+            "join p.details.group as group "
             "where f is null"
         )
 
         filesQuery = (
             "select sum(origFile.size) from OriginalFile as origFile "
+            "join origFile.details.group as group "
```
joshmoore, line 265: Are you using the whole object or could just the group id be returned?

joshmoore (commented):
Few minor code related comments.

pwalczysko:
Tested on merge-ci

1. Remove a user with a lot (GB) data from group
2. Add the user back to group
3. Remove a user with a little (< 10MB) from group
4. Add the user back to group
5. incrementally import images for a user with little data, observe how the usage rises
6. Run a script as a user, then delete all the images of that user -> observe that the user usage is still listed, but very small (440 B)

All the tests above made sense as checked in the report listings of the usage for that group when logged in as a group owner.

Looks good fmpov.

pwalczysko (approved):

omeroweb/webadmin/views.py
will-moore: This query isn't selecting (returning) the group. It's just in the join for filtering. Would that be more efficient to use
```"join origFile.details.group.id as group "```
Actually, that gives me an Illegal query exception.

omeroweb/webadmin/templates/webadmin/myaccount.html
will-moore: I don't think we support any versions of IE now, and that line was only for handling IE 8 (which we certainly don't support). I haven't tested on omero-web on IE for a long time now.

pwalczysko: I tried on Maxquant on IE9, and the webclient does not render correctly, basically not usable at all.
Screenshot 2021-11-19 at 14 38 40

I think we do not even have an approximate capacity to test on IE at all, the VB Win10 I have is only containing Edge.
Maybe we should remove mentions of IE from this code as well as the supported browsers list ?

omeroweb/webadmin/views.py
joshmoore: Sorry for the noise. Your `sum()` takes care of reducing the output size.

PR #336, opened 2021-11-23
Acquisition metadata json

will-moore:
See https://github.com/IDR/idr.openmicroscopy.org/issues/149

This adds Acquisition metadata URL to webclient. Also adds ObjectiveSettings support to JSON api, which requires https://github.com/ome/omero-marshal/pull/72

To test, go to:
 - /webclient/api/metadata_acquisition/IMAGE_ID/ 
 - /api/v0/m/images/IMAGE_ID/objective_settings/
 

e.g. https://merge-ci.openmicroscopy.org/web/webclient/api/metadata_acquisition/140015/ (user-3)
and https://merge-ci.openmicroscopy.org/web/webclient/api/metadata_acquisition/200515/ 
and depending on the metadata available, you should get JSON like:

```
{
    "channels": [
        {
            "id": 2641,
            "excitationWave": null,
            "emissionWave": 528,
            "logicalChannel": {
                "id": 2636
            }
        }
    ],
    "objectiveSettings": {
        "id": 1072,
        "objective": {
            "id": 1068,
            "model": "1-UB932",
            "manufacturer": "Olympus",
            "nominalMagnification": 60,
            "lensNA": 1.4,
            "workingDistance": 0.1,
            "immersion": "Oil",
            "correction": "PlanApo"
        }
    }
}
```

cc @gwaygenomics

### DIFF
--- a/omeroweb/api/urls.py
+++ b/omeroweb/api/urls.py
@@ -147,8 +147,15 @@ api_image = url(
 Image url to GET or DELETE a single Image
 """
 
+api_image_objective_settings = url(
+    r"^v(?P<api_version>%s)/m/images/(?P<image_id>[0-9]+)/objective_settings/$"
+    % versions,
+    views.image_objective_settings,
+    name="api_image_objective_settings",
+)
+
 api_image_datasets = url(
-    r"^v(?P<api_version>%s)/m/images/" "(?P<image_id>[0-9]+)/datasets/$" % versions,
+    r"^v(?P<api_version>%s)/m/images/(?P<image_id>[0-9]+)/datasets/$" % versions,
     views.DatasetsView.as_view(),
     name="api_image_datasets",
 )
@@ -414,6 +421,7 @@ urlpatterns = [
     api_dataset_images,
     api_dataset_projects,
     api_image,
+    api_image_objective_settings,
     api_image_datasets,
     api_screen,
     api_screens,
--- a/omeroweb/api/views.py
+++ b/omeroweb/api/views.py
@@ -873,6 +873,30 @@ class ExperimenterGroupsView(ObjectsView):
         return opts
 
 
+@json_response()
+@login_required(useragent="OMERO.webapi")
+def image_objective_settings(request, image_id, conn=None, **kwargs):
+
+    image = conn.getObject("Image", image_id)
+    if image is None:
+        raise NotFoundError("Image not found")
+
+    settings = image.getObjectiveSettings()
+    objective = settings.getObjective()
+    correction = objective.getCorrection()
+    immersion = objective.getImmersion()
+
+    settings_object = settings._obj
+    objective_object = objective._obj
+    # replace unloaded objects with loaded ones
+    objective_object.correction = correction._obj
+    objective_object.immersion = immersion._obj
+    settings_object.objective = objective_object
+
+    encoder = get_encoder(settings_object.__class__)
+    return encoder.encode(settings_object)
+
+
 class SaveView(View):
     """
     This view provides 'Save' functionality for all types of objects.
--- a/omeroweb/webclient/urls.py
+++ b/omeroweb/webclient/urls.py
@@ -112,6 +112,11 @@ urlpatterns = [
         name="load_metadata_acquisition",
     ),
     url(
+        r"^api/metadata_acquisition/(?P<image_id>[0-9]+)/$",
+        views.api_metadata_acquisition,
+        name="api_metadata_acquisition",
+    ),
+    url(
         r"^metadata_preview/(?P<c_type>((?i)image|well))/"
         r"(?P<c_id>[0-9]+)/(?:(?P<share_id>[0-9]+)/)?$",
         views.load_metadata_preview,
--- a/omeroweb/webclient/views.py
+++ b/omeroweb/webclient/views.py
@@ -1841,20 +1841,24 @@ def api_metadata_acquisition(request, image_id, conn=None, **kwargs):
 
     json_data = {"channels": []}
     for ch in channels:
-        channel_json = {"id": ch.id}
+        channel_json = {
+            "id": ch.id,
+            "excitationWave": ch.getExcitationWave(),
+            "emissionWave": ch.getEmissionWave(),
+        }
         logicalChannel = ch.getLogicalChannel()
         if logicalChannel is not None:
-            channel_json = {
-                "exWave": ch.getExcitationWave(),
-                "emWave": ch.getEmissionWave(),
+            channel_json["logicalChannel"] = {
+                "id": logicalChannel.id,
             }
             lightPath = logicalChannel.getLightPath()
             if lightPath is not None:
-                lightPathDichroic = lightPath.getDichroic()
-                if lightPathDichroic and lightPathDichroic._obj is not None:
-                    channel_json["dichroic"] = {
-                        "manufacturer": lightPathDichroic.manufacturer,
-                        "model": lightPathDichroic.model,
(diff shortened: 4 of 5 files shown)


gwaygenomics:
Wow! Thanks @will-moore - exciting to see such quick progress on a request. A new record I think

sbesson:
Since these objects are part of the OME(RO) model,  what is the rationale for not exposing this metadata via `omero-marshal` and the JSON API? https://github.com/ome/omero-marshal/blob/master/omero_marshal/decode/decoders/channel.py already includes support for a few channel properties?

will-moore:
@sbesson No objection to exposing this via omero-marshal and the JSON API. Just quite a bit of work since the stack is much bigger. It would require significant code changes and releases of omero-marshal, omero-py and omero-web, as well as tests for each.
NB: Just started adding tests for this PR in https://github.com/ome/openmicroscopy/pull/6295

sbesson:
> It would require significant code changes and releases of omero-marshal, omero-py and omero-web, as well as tests for each.

Why `omero-py` ? I thought `omero-marshal` and `omero-web` only unless there are related changes in the BlitzGateway to be added?

will-moore:
omero-web uses the BlitzGateway objects to generate the query: https://github.com/ome/omero-web/blob/394e75170f5d6e9997edf12f6906c489186d5f22/omeroweb/api/api_query.py#L142

This approach has been used in the past because BlitzGateway already has all the logic for generating the HQL queries for loading objects, pagination etc and it allows all the end-points to use the same pattern.
But I guess It's possible to use a different approach for this case where we don't need that, just using the code that's in this PR.

chris-allan:
Yes it's a big job but it's also the correct way to do this as it builds on our existing tooling, provides value outside this limited use case and is consistent with our overarching strategy for handling OME model derived metadata in OMERO.web.

You also might also consider whether putting Objective metadata under `/api/image/ID/objective_settings` is conceptually inconsistent with the OME  model and way the rest of this metadata is handled. It probably should instead be a combination of both `/api/image/ID/objective_settings` and `/api/image/instrument/objective/ID`.

I can for example see `/api/image/instrument` being of incredible value and would be a critical starting point to enabling the modernization of areas of OMERO.web that display and interact with instrument metadata; for example, the right-hand panel.

sbesson:
In addition to the suggestions expressed in https://github.com/ome/omero-web/pull/336#issuecomment-979911259 regarding the handling of `ObjectiveSetting` (and more generally `Instrument`), this PR also exposes some channel metadata namely the emission and excitation wavelength.

Two comments on this:
-  there is some level of overlap with the metadata exposed in the `imgData` endpoint eg.g https://idr.openmicroscopy.org/webclient/imgData/9844380/. Assuming we wanting to go with another `webclient` URL, how is the decision made regarding what gets exposed where?
- seconding @chris-allan's comment, there is definitely value in exposing metadata derived from the OME model via the JSON API. In the case of channels, the encoding/deconding functionality is already available in `omero-marshal` but there is no API endpoint. Would it be worth looking into exposing this e.g. via `/api/channels/ID` and/or `/api/images/ID/channels`?

will-moore:
@chris-allan I think the graph traversal from `/api/images/ID/objective_settings/` to include the `objective` itself is consistent with the way we traverse from `/roi/` to include Shapes or /image/` to include Pixels? Otherwise from an Image ID, you have to make 2 calls to get the Objective ID from the settings, then to get the Objective itself.

`api/images/ID/instrument/` would be nice too, but the priority with this PR was to provide the Objective used for an Image, so I would prefer to leave that for another PR.

@sbesson The /webclient/api/metadata_acquisition/IMAGE_ID/ in this PR was really just a partial JSON representation of the html in: webclient/metadata_acquisition/image/11514301/ used for the right panel, which displays the channel wavelength etc. So these are historical decisions based on the webclient UI vv channel rendering settings etc are loaded separately for Image viewers. I can also see `/api/images/ID/channels` being useful (although not useful for Image viewers as it wouldn't include rendering settings).

chris-allan:
> I think the graph traversal from /api/images/ID/objective_settings/ to include the objective itself is consistent with the way we traverse from /roi/ to include Shapes or /image/` to include Pixels?

I disagree. The OME model hierarchy is Image --> Instrument --> Objective, these elements host each other. ObjectiveSettings has a reference to an Objective but it is not the container for the element. This is a "backreference" similar to the discussions we had on ome/omero-marshal#71 surrounding ROI --> Shape <-- ROI.

> Otherwise from an Image ID, you have to make 2 calls to get the Objective ID from the settings, then to get the Objective itself.

Correct, unless you also want to reopen ome/omero-marshal#71 and get a consensus there as well in order to address that deficiency.

> api/images/ID/instrument/ would be nice too, but the priority with this PR was to provide the Objective used for an Image, so I would prefer to leave that for another PR.

I don't think that convenience use case is sufficient to justify not doing this thoroughly and correctly from the beginning. We will have to retrofit it later otherwise as well as document and support an exception. I understand it's a significant amount of work; that's why I left it out of the initial implementation of omero-marshal. We didn't need it at the time and doing it partially is dangerous, that doesn't mean it's not incredibly valuable to do.

will-moore:
Maybe ObjectiveSettings -> Objective is more of a "reference" than a "backreference" since Objective isn't a container of ObjectiveSettings (in the way that an ROI is a container of Shapes).
When marshalling a Shape, we would never want to marshal the parent ROI (as that would lead to a recursive marshalling) but I don't see that happening with ObjectiveSettings marshalling Objective. Maybe my thinking is due to coming from OMERO API, instead of OME model.

I'm happy to add `/api/objectives/ID/` for directly getting an Objective based on it's ID. Then I guess I could NOT marshal the Objective within the ObjectiveSettings and let users do 2 calls. Or to use the other /webclient/api/metadata_acquisition/IMAGE_ID/ endpoint added in this PR.
In fact, it probably makes sense to separate the `/webclient/api/` changes and JSON `/api/` changes into 2 different PRs, since they are separate functionality, and it could take a while to get the JSON api changes done.

chris-allan:
Firstly, omero-marshal tries to follow the OME-XML model object and reference semantics as closely as possible with a few notable exceptions for OMERO objects containing attributes not present in OME-XML. This is why, for example, we marshal OMERO LogicalChannel attributes with Channel, use OME-XML case for attributes, and the consistent, predictable OME-XML element naming style. The desired result being someone is able to read the OME-XML documentation and not have to guess where data will be when marshalled and we don't have to re-document it.

All the "Settings" objects (DetectorSettings, LightSourceSettings, ObjectiveSettings) are special references (*Ref in OME-XML) and have the class hierarchy *Settings --> Settings --> Reference within the OME-XML model.  The concept of a reference as well as a back-reference within OME-XML, and by extension omero-marshal, is not exclusively about avoiding cyclical marshalling although that is certainly one of the key concerns. The XML DOM has the concept of element containment that a graph based system like OMERO does not have. Consequently, as far as omero-marshal is concerned an element should not be in two places at once and should not be marshalled twice. We've knowingly made some exceptions to this, most notably WellSample --> PlateAcquisition (ome/omero-marshal#31). You could make an argument for ObjectiveSettings/Objective here but you'll have to make it and we'll all have to agree to that exception; Instrument (Image --> Instrument --> Objective, Image --> ObjectiveSettings --> Objective) is much more likely to be marshalled with an Image.

I would focus on getting what's under `/api` right and expanding omero-marshal to handle the Instrument hierarchy correctly. Such improvements have significant utility beyond this use case and tremendous benefit across the entirety of omero-web. Adding sanctioned use case specific crutches under `/webclient/api`, especially ones that do not follow omero-marshal semantics, seems to me to be a really bad idea. We already have enough problems with `imgData` as it is.

PR #342, opened 2021-12-15
table csv writer

### LINKED ISSUE #338: OMERO.table download csv with commas
E.g. download http://idr.openmicroscopy.org/webclient/omero_table/14209193/ as csv
The columns are broken because some cells have commas, e.g:

`Gene identifiers and symbols from GRCh38.p5, Ensembl release 84, Mar 2016. Added by IDR curators.`

will-moore:
Fixes #338 
This uses the built-in `csv` module to write CSV strings when downloading an OMERO.table as CSV. This handles table values that contain commas.

To test:
 - Create an OMERO.table with column names or data values contain commas. (already did one for https://merge-ci.openmicroscopy.org/web/webclient/?show=plate-4351)
 - View table (click eye icon) and download as CSV
 - Check that the commas are correctly escaped (don't split a value into two cells)

 - Test updated in https://github.com/ome/openmicroscopy/pull/6299 to check for `,` in CSV values should be green.

### DIFF
--- a/omeroweb/webclient/views.py
+++ b/omeroweb/webclient/views.py
@@ -23,8 +23,10 @@ or a redirect, or the 404 and 500 error, or an XML document, or an image...
 or anything."""
 
 import copy
+import csv
 import os
 import datetime
+from io import StringIO
 import Ice
 from Ice import Exception as IceException
 import logging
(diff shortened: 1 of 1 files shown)


will-moore:
cc @erindiel If you want to test this on some 'real' data?

will-moore:
Causing test to fail at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/976/testReport/junit/OmeroWeb.test.integration.test_table/TestOmeroTables/test_table_html/
Looks like adding white-space or something between csv lines...

joshmoore:
In general, :+1: for use of standard libraries for writing CSV due to its gnarliness. Did you happen to check how much overhead it introduced? I assume not a ton, but checking briefly I saw that (at least in newer django versions) it can write directly to the response:

https://docs.djangoproject.com/en/4.0/howto/outputting-csv/

Is that worth trying out?

will-moore:
Thanks @joshmoore, I'd forgotten about those docs.
But actually, their "Streaming large csv files" example is pretty much what we're doing here, since we're using a Streaming response and the csv writer.

joshmoore:
Fair enough. There are enough redirections that I just can't tell if there's (significant) overhead or not at a glance.

PR #343, opened 2021-12-15
Omero table json link

### LINKED ISSUE #340: Add link to view JSON table data
During the 4DN demo, Will showed how to add "/json/" to the url in order to see the raw data. The header of the HTML page should include a link to that representation.

will-moore:
Fixes #340.

Adds a link to JSON view of the page (with same pagination and filtering). Also moves the similar link to CSV page from a Button to a link (bad practice to use a button for a link!).

When there is no pagination, the Download as CSV button and the CSV link do the same thing, but I think it's still useful to have the link available.

To test:
 - Click the JSON link - it should open the JSON version of the page in the browser.
 - The CSV link will download the CSV
 - Both links should include the current pagination and filtering queries
 
cc @erindiel 

Screenshot 2021-12-15 at 12 07 55

### DIFF
--- a/omeroweb/webclient/templates/webclient/annotations/omero_table.html
+++ b/omeroweb/webclient/templates/webclient/annotations/omero_table.html
@@ -55,7 +55,7 @@
             </button>
         {% endif %}
         <br>
-        Show page as:
+        Show current page as:
         <a href="{% url 'omero_table' data.id 'csv' %}?limit={{ meta.limit }}&offset={{ meta.offset }}{% if meta.query != '*' %}&query={{ meta.query|urlencode }}{% endif %}">CSV</a> |
         <a href="{% url 'omero_table' data.id 'json' %}?limit={{ meta.limit }}&offset={{ meta.offset }}{% if meta.query != '*' %}&query={{ meta.query|urlencode }}{% endif %}">JSON</a>
     </div>
@@ -69,7 +69,7 @@
 
     {% if meta.query != '*' %}
         <div class="back">
-            <a href="{{ meta.url }}">< Remove query</a>
+            <a href="{{ meta.url }}">< Remove filter</a>
         </div>
     {% endif %}
     <h2>{{ data.name }}</h2>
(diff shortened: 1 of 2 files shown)


pwalczysko:
This works as described in the PR header.

Couple of RFEs:

1. Maybe rename the "Remove query" button to "Remove filter" ? This would then match the wording with the wording you are using for the new top-right-corner button ?

Screenshot 2021-12-16 at 13 04 52

2. The pagination works, but in case of JSON, I do wonder how would I "glue" the three pages together ? Even the ordinal number of the row on the second page starts with "0" and not "201" on page 2. Can maybe a json for the non-paginated, whole table be offered, as it is offered for the CSV ?

3. "Show page" -> "Show current page" - will make much clearer the difference between the two rows of buttons in top-right imho.
Screenshot 2021-12-16 at 13 08 35

will-moore:
@pwalczysko Thanks, renamed as suggested in 1) and 3) above.
For the pagination, the rows will always start with `0` as this is a list.

For example, this is all the JSON returned by `/webclient/omero_table/72405/json/?limit=2&offset=6`.
We get 2 rows, and we can simply concatenate them with the rows we get from other pages e.g. `?limit=2&offset=8`

```
{
    "data": {
        "column_types": [
            "WellColumn",
            "StringColumn",
            "DoubleColumn",
            "LongColumn",
            "DoubleColumn",
            "StringColumn"
        ],
        "columns": [
            "Well",
            "Drug",
            "Concentration",
            "Cell_Count",
            "Percent_Mitotic",
            "Well Name"
        ],
        "rows": [
            [
                857,
                "DMSO",
                5.5,
                150,
                4,
                "b3"
            ],
            [
                858,
                "DMSO",
                75.5,
                550,
                4,
                "b4"
            ]
        ],
        "name": "bulk_annotations",
        "path": "bulk_annotations",
        "id": 72405
    },
    "meta": {
        "rowCount": 12,
        "totalCount": 12,
        "limit": 2,
        "offset": 6,
        "query": "*",
        "page": 4,
        "url": "/webclient/omero_table/72405/",
        "next": "/webclient/omero_table/72405/?limit=2&offset=8",
        "prev": "/webclient/omero_table/72405/?limit=2&offset=4"
    }
}
```

PR #345, opened 2021-12-16
Fix link to referrer-policy

sbesson:
The inclusion of this new property in the generated configuration documentation (https://github.com/ome/omero-documentation/pull/2203/files#diff-f9698d6a6ee93a6e33a579f394fd06094e079981b7fb54f8b8a70a8f7469e2cbR2260) highlighted that the anchor does not longer exists but redirects to a new URL

### DIFF
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -1101,7 +1101,7 @@ CUSTOM_SETTINGS_MAPPINGS = {
         (
             "Default content for the HTML Meta referrer tag. "
             "See https://www.w3.org/TR/referrer-policy/#referrer-policies for "
-            "allowed values and https://caniuse.com/#feat=referrer-policy for "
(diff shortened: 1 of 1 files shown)


jburel (approved):
A new release will be required to have the variable updated in the doc.
To avoid that (similar for omero-py) we may want to install ``master`` for the latest doc instead of the release version but that will make the documentation synch a bit more complex

sbesson:
I would rather avoid opening documentation PRs for unreleased code. Let's imagine a new API PR introducing a new `omero.web.*` configuration property is merged into `master` but the minor release is pending other changes under review as it happens frequently. In this scenario, I think an `omero-documentation` PR documenting the property would be erroneous.

In the scope of this PR, my intention was to create a patch release of `omero-web`.

jburel:
I agree. I don't want that either. I was just indicating that it won't be "automatic" to have it in the doc.

PR #347, opened 2021-12-22
Set keepalive response content type to text/plain

knabar:
We ran into a setup where an intermediary proxy server modified the response of the `keepalive_ping` API call by appending extra whitespace.  This did not affect OMERO.web itself, which does not check the response content, but broke another client that expects the exact two-character response `OK`.

While this is easy to work around, it might be useful to change the response content type to `text/plain` to match the actual response, and to discourage intermediaries from processing the response as HTML.


will-moore:
👍

will-moore (approved):
Looks good. thx

PR #349, opened 2022-01-13
Make 'remove all filters' properly clear map filters.

### LINKED ISSUE #348: Key-Value filters are not cleared by the 'remove all filters' option
In the center panel, the 'Add Filter' dropdown allows the user to create different filter types to restrict which images are displayed.

That dropdown includes a 'Remove all filters' option which clears any filters that the user has set.

However, while running this option removes Key-Value filters from the display, those filters are still applied. I don't think you can remove them without refreshing the page. Removing map filters individually still works.

This problem appears to be exclusive to Key-Value filters, other filter types are removed successfully.

DavidStirling:
Fixes #348

I believe that the problem with the 'remove all filters' option was that it didn't wipe the Key-Value filters list. Adding code to do this (if any filters are present) seems to fix the issue.

### DIFF
--- a/omeroweb/webclient/templates/webclient/data/includes/center_plugin.thumbs.js.html
+++ b/omeroweb/webclient/templates/webclient/data/includes/center_plugin.thumbs.js.html
@@ -448,6 +448,11 @@ $(document).ready(function() {
                 $("#dataIcons li.row").removeClass('ratingFilter_hidden');
                 // reset Tag filtering
                 filter_tag_ids = [];
+                // clear filter objects if any exist
+                if (filterObjects.length) {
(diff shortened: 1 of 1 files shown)


joshmoore:
Launched the workflows for testing. @DavidStirling : do you happen to see an invite to the org?

DavidStirling:
Unfortunately I'm not seeing an invite, sorry!

joshmoore:
No worries. Resent.

DavidStirling:
Got it, thanks!

will-moore:
Thanks for that. Tested locally. Looks good.

PR #352, opened 2022-02-07
Multiple omero tables

### LINKED ISSUE #351: Handle multiple OMERO.tables on Image containers
Currently, the `Tables` tab for an Image attempts to load OMERO.table data (selecting for the row where `?Image=ID`) from Tables attached to parents: We make 4 calls for Project, Dataset, Screen and Plate, of the form: `/webgateway/table/Screen.plateLinks.child.wells.wellSamples.image/2/query/?query=Image-2`

The first one that returns some valid data gets displayed and the others are ignored.
Also, if there is more than 1 table on any object, only the most recent is considered.

This means that a user is only shown data from 1 "~randomly" chosen table.
We need to allow a user to see ALL of the tables that are linked to parent objects and to choose which to show.
Use-case in `idr0123` with table on Project AND Dataset.

Need to first query the 4 parents to get list of tables to show this info for each (some could be in tooltip):
 - Annotation ID & File ID
 - File Name
 - Date
 - Owner
 - link owner & date?
 - namespace
 - Column names ?? (this would mean the table needs to be opened in this initial query, which already happens for up to 4 tables, but not more - e.g. if there are lots of tables on 1 object - Probably best not to do this just yet)

When the user expands a chosen Table, they get to see the current `Column: value` data for the image.
If we find only 1 table, this could be automatically expanded (to be similar to current behaviour).

cc @stick @muhanadz @erindiel @francesw @pwalczysko

will-moore:
Fixes #351

This adds support for multiple OMERO.tables on the containers of an Image or Well. See issue above.

To test:

 - For the cases where we only have a single table on a image container (Screen, Plate, Project or Dataset) this table will be loaded immediately: e.g. https://merge-ci.openmicroscopy.org/web/webclient/?show=well-10584 (user-3)

Screenshot 2022-02-07 at 17 14 17

 - When we have multiple tables, they are listed and can each be loaded by clicking on them. e.g. https://merge-ci.openmicroscopy.org/web/webclient/?show=image-18221 (user-3):

Screenshot 2022-02-07 at 17 18 23.

In both cases, the tooltip on the table gives extra info as before.

cc @sbesson @francesw

### DIFF
--- a/omeroweb/webclient/controller/container.py
+++ b/omeroweb/webclient/controller/container.py
@@ -29,6 +29,7 @@ from django.utils.encoding import smart_str
 import logging
 
 from omeroweb.webclient.controller import BaseController
+from omeroweb.webgateway.views import _bulk_file_annotations
 
 logger = logging.getLogger(__name__)
 
@@ -241,6 +242,31 @@ class BaseContainer(BaseController):
         Get the annotion counts for the given objects
         """
         return self.conn.getAnnotationCounts(objDict)
+
+    def countTablesOnParents(self):
+        """
+        For Wells or Images, query parents for any OMERO.tables
+        """
+        img_queries = [
+            "Screen.plateLinks.child.wells.wellSamples.image",
+            "Plate.wells.wellSamples.image",
+            "Project.datasetLinks.child.imageLinks.child",
+            "Dataset.imageLinks.child",
+        ]
+        well_queries = ["Screen.plateLinks.child.wells", "Plate.wells"]
+        total = 0
+        if self.image is not None:
+            for query in img_queries:
+                result = _bulk_file_annotations(None, query, self.image.id, self.conn)
+                if "data" in result:
+                    total += len(result["data"])
+        elif self.well is not None:
+            for query in well_queries:
+                result = _bulk_file_annotations(None, query, self.well.id, self.conn)
+                if "data" in result:
+                    total += len(result["data"])
+
+        return total
 
     def canExportAsJpg(self, request, objDict=None):
         """
--- a/omeroweb/webclient/static/webclient/css/layout.css
+++ b/omeroweb/webclient/static/webclient/css/layout.css
@@ -1243,9 +1243,14 @@
 	
 	
     /*   Acquisition Details */
+
+    h1.can-collapse {
+        border: 1px solid #bbb;
+        margin:5px 0;
+        padding: 7px;
+    }
     
     .can-collapse {
-        border: 1px solid #bbb;
         -webkit-box-shadow:0 1px 0 rgba(255,255,255,.7);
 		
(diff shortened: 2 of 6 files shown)


sbesson (commented):
The IDR use case behind this PR certainly points the current UI behavior as not fit for purpose for dealing with multiple tabular attachments.

Overall I find the new UI element (replacing the `Info` element) to be a natural and straighforward way to expose multiple tables. It is consistent with the UI used for `Attachments`. The auto-expansion feature in the case of single tables does not break the user workflow and for multi-tables, it allows to expand/shrink/navigate between the available tables.

One inline question about the tooltip. Otherwise, two general comments:

- the `(linked to: )` text is interesting as it provides some provenance metadata. I wonder whether this information should necessarily be visible immediately to the user. Naively, I was almost expecting clicking on `Project: xxx` to bring me to this project. An alternative would be to move it under the tooltip 
- with more use cases, I can foresee some RFEs associated with grouping these annotations by namespaces

Including also the Glencoe team to make sure this UI addition would not break existing workflow @knabar @erindiel (anyone else? 😄 )

will-moore:
I fixed the tooltip: "Linked by".
I also updated the `(Linked to Project-1)` to actually be a link.
But I think that it is useful to have that at the top-level on the object itself instead of in the tooltip as it helps give a better understanding of where these are coming from: to understand that these are not linked to the Image itself. E.g. see confusion at https://forum.image.sc/t/omero-saving-as-omero-table/54360/3

I also added sorting by name then date, so if you have a bunch of tables, they are sorted consistently:
Screenshot 2022-02-08 at 23 02 46

sbesson:
Thanks, the tooltip looks good now. The only downside of the latest version is that clicking on the triangle icon  does not expand/shrink the table. Instead you need to click on the table name. Is this expected?

Screenshot 2022-02-09 at 09 16 52

will-moore:
Hmmm, In order to make the link clickable as a link, I had to remove the `[click the whole row to expand]` behaviour.
But I'll try and fix it so that clicking on the name OR the `>` does the expanding

will-moore:
Now showing Tables counts for Wells and Images:

Screenshot 2022-02-09 at 16 09 33

sbesson (approved):
With the last commit, the clickable issue raised above is now fixed. The addition of `Tables` count is also a great asset for end users to be able to know that one or several tables are associated to a particular image and facilitate the discovery of this metadata

I assume the extra queries performed to compute the tables count will have a minimal performance impact but this is something we might need to check on an IDR scale deployment. 

Otherwise, I find the proposed changes brings the handling of tabular metadata in the right-hand panel to parity with image annotations and goes in the sense of increased usability. Happy to see this included in the next minor release of OMERO.web.

PR #359, opened 2022-02-21
Add extra

jburel:
Add ``django-cors`` and ``django-redis`` as extra
Background https://github.com/ome/omero-documentation/pull/2216


setup.py
```diff
@@ -61,4 +61,7 @@ def read(fname):
     ],
     include_package_data=True,
     tests_require=["pytest"],
+    extras_require={
+        "default": ["django-cors-headers==3.7.0", "django-redis==5.0.0"],
```
sbesson, line 65: Looking at our internal documentationa and consumptions, it looks like these two dependencies are typically activated independently of each other - see https://github.com/search?q=org%3Aome+django-cors-headers&type=code and https://github.com/search?q=org%3Aome++django-redis&type=Code.

Balancing the pros and the cons of the two approaches:
- a single `default` key has the benefit of installing all the extra dependencies in one go. I find `default` fairly uninformative
- declaring two separate keys e.g. `cors` and `redis` would have the benefit of explicitness and finer granularity. Consumers needing to install both would require the slightly more verbose `pip install omero-web[cors,redis]`

jburel: We could offer multiple option: ``all, cors, redis``

jburel:
Discussed this morning
* ``cors-headers`` is lightweight and required configuration to be used so this could probably be included by default
* ``django-redis`` is a binding, ``redis`` needs to be installed. Also apps like ``mapr`` require certain version so that could create a coupling between the 2.

setup.py
```diff
@@ -54,11 +54,14 @@ def read(fname):
         "omero-py>=5.7.0",
         # minimum requirements for `omero web start`
         "Django>=1.11,<2.0",
-        "django-pipeline==1.6.14",
+        "django-cors-headers==3.7.0" "django-pipeline==1.6.14",
```
will-moore, line 57: I think you're missing a comma there?

jburel: thanks

sbesson (approved):

will-moore:
Looks good 👍

PR #360, opened 2022-02-22
update build year

jburel:
Nothing to test


sbesson (approved):
This matches what has been done previously e.g. in https://github.com/ome/omero-web/commit/4174edec3a64d5f967b17c2fe0b85bea7f3d1627. 

Unlike the version which is managed at release time, we do not have a good mechanism for bumping this date. I wonder if it can be sourced from the packaging information. That being said, it's fairly a minor change and only need to happen once a year. Maybe an issue?

PR #365, opened 2022-03-23
Added versions and logos to feedback page

kyleBalisWest:
Added versions and logos to feedback page,
image
Once ome/omero-server#126 is merged we should get bio-formats versions included as well. Right now omero-web and omero-server's versions are expected.


sbesson (commented):
Thanks for your (first?) contribution @kyleBalisWest. Barring one typo, the proposed changes look very straightforward 

Tested in the context of an OMERO.web deployment talking to an OMERO.server that has support for exposing the underlying versions

Screenshot 2022-03-24 at 09 17 45

I'll try and run some tests against a server without the Bio-Formats version capability confirm the  logic works as expected.

The primary concern from my side is that the metadata is added to the feedback UI which can be selectively disabled - see https://docs.openmicroscopy.org/omero/5.6.3/sysadmins/config.html#omero-web-feedback-comment-enabled.

To maximize the exposure, a potential question is whether this addition should be moved to its own about-like endpoint. For reference, the UI that comes from ImageJ/Fiji when going through `Help > About Plugins > Bio-Formats Plugins...`

Screenshot 2022-03-24 at 09 20 55

chris-allan:
Thanks for the feedback, @sbesson.

We wanted to keep this light touch and reuse as much infrastructure as possible in the first instance but I definitely agree that an "About" page is something we just need to do. Since actually doing that is a pretty big change and would require a number of design decisions to be made I didn't want to completely overwhelm @kyleBalisWest with those for his first PR.

Definitely happy to have a conversation on some of those bigger design decisions. Some that cross my mind right away:

* Where in the UI does an "About" button go?
* What parts of the UI have access to the "About" button?
* Is "About" a modal dialog, using jQuery-UI, like we use for tag management? A separate popup page like the feedback comment? Something else entirely?

/cc @will-moore

will-moore:
I guess we don't have a typical app menus in the webclient. The most obvious place to look for 'About' is probably in the `top_links`, where we have `Help`. But adding an option there is hard because you'd have to update everyone's `top_links` config when they upgrade, which is tricky. And also the space can get limited there if you've configures a bunch of links (e.g. with mapr) and the only option for links is to open a new tab (not a popup window).
Probably the best option is to add an `About` item to the user's menu (where you currently have the `Send Feedback` option) and have it open a popup window in the same way.

sbesson:
Adding a few extra thoughts.

Unlike the Bio-Formats plugin use case above, the content is much more stateful and rich in the sense that it provides information about the OMERO.web deployment itself but also the underlying server you are connecting to. A use case that might not be so common in production but heavily used in development/testing is the case where several servers are defined in the Web configuration and appear in the drop-down menu.

>  Probably the best option is to add an About item to the user's menu (where you currently have the Send Feedback option) and have it open a popup window in the same way.

Although I would tend to agree this is a natural place, this will suffer from similar limitations as the feedback UI above i.e. it can be hidden notably in read-only mode (see https://github.com/ome/openmicroscopy/pull/5892). 

> Is "About" a modal dialog, using jQuery-UI, like we use for tag management? A separate popup page like the feedback comment? Something else entirely?

At least, I find the modal About window like the ones used in several of the OMERO.web apps (iviewer, figure) fairly standard and not too invasive.

> Since actually doing that is a pretty big change and would require a number of design decisions to be made I didn't want to completely overwhelm @kyleBalisWest with those for his first PR

All understood and I hope I didn't manage to scare away @kyleBalisWest with my review 😄 . Happy to start capturing and moving all the considerations around a new About UI into a separate issue.

kyleBalisWest:
No @sbesson, you didn't scare me away. I appreciate all the feedback you can give me 😄

will-moore:
It's not ideal that this info would be hidden when you're a `public` user, but I don't consider that a blocker on this PR (or the addition of an "About" option to that menu) if that is a good balance of solving this issue for the majority without redesigning the UI too much.
Other options: you could add a 3rd button to:
Screenshot 2022-03-25 at 10 20 06
or add a permanent "About" button/link between the OMERO icon and the top-links?

joshmoore:
Especially for smaller windows, it might be generally useful to consider a hamburger menu to hold some of these items. That way we could have a single additional icon (`///`) and About could live under it.

chris-allan:
@kyleBalisWest: On my `logos-and-versions` branch you can now find an additional commit that changes the template variables so that we now have versions for all servers you have configured via `omero.web.server_list`. The new variable is `versions` and has the prototype:

```
[
  {
    "server": <server>,
    "bioformats_version": <bioformats_version>,
    "omero_version": <omero_version>,
    "omeroweb_version": <omeroweb_version>,
  },
  ...
]
```

I'll let you get creative on how you want to visually layout the versions for multiple servers.

kyleBalisWest:
Okay, I added them as ui chips because I thought that was a nice visual way to list them.

image (1)

sbesson:
Tested using our CI servers

Screenshot 2022-04-04 at 16 01 44

Both production servers (i.e. using released versions of OMERO) and development servers are properly recognized. Also the Bio-Formats version parsing shows that server with and without https://github.com/ome/omero-server/pull/126 are properly handled, falling back to `Unknown`.

A few comments:

- the OMERO.web version is duplicated for each server. This should always be the case as this exposes information about a single OMERO.web deployment can connect to multiple servers. An alternative to the latest proposal would be to update the dictionary and the UI to expose a single OMERO.web version following by an array of versions for each server?
- while testing this PR, I accidentally realised that one of the configuration in the servers list could not be reached which caused the feedback endpoint to time-out. This might be an edge case but it can be complex to debug especially if the server is not the default.
- testing the case of a stopped OMERO.server, the feedback endpoint returns with a version shown as `Unknown` which matches my expectation

Screenshot 2022-04-04 at 17 03 55

omeroweb/feedback/templates/comment.html
```diff
@@ -29,6 +29,26 @@
     <div class="bodyWrapper">
         <h1>{% trans "OMERO.web Help" %}</h1>
 
+        <ul class="logo_list">
```
sbesson, line 32: Custom deployments might want to override/modify this base template. My main suggestion would be to add the markup allowing to easily extend this template.
@will-moore is adding `{% block logos %}` here the correct way to achieve this?

omeroweb/feedback/templates/comment.html
```diff
@@ -29,6 +29,26 @@
...
+            </li>
+            <li class="middle_logo">
+                <img src="{% static "feedback/image/omero.svg"|add:url_suffix %}" alt="OMERO Logo" height="25">
+            </li>
+            <li>
+                <img src="{% static "feedback/image/bio-formats.svg"|add:url_suffix %}" alt="Bio-Formats Logo" height="25">
+            </li>
+        </ul>
+
+        <p>
```
sbesson, line 44: Same as above, we might want to encapsulate the version block using the `{% block versions %}` markup.

sbesson (changes requested):
Two suggestions to facilitate the extension of the comment template.

Otherwise, this works as expected and exposes versioning metadata that has been on the wish list for a long time. So I would propose to go ahead and make this available as a patch release once this is approved.

PR #368, opened 2022-04-01
Elasticsearch

will-moore:
This allows you to configure omeroweb search to use an elasticsearch endpoints provided by https://github.com/ome/omero_search_engine_client to replace regular search...


jburel:
Swapping the 2, means that search by Image ID for example is no longer possible.

omeroweb/settings.py
```diff
@@ -980,6 +980,12 @@ def leave_none_unset_int(s):
         int,
         ("ID of group to pre-select in search form."),
     ],
+    "omero.web.elasticsearch.url": [
+        "ELASTICSEARCH_URL",
+        "https://idr-testing.openmicroscopy.org/searchengine/",
```
joshmoore, line 985: Could this default to `/searchengine` to work on the same host?

omeroweb/webclient/templates/webclient/base/includes/search_field.html
```diff
@@ -22,6 +22,10 @@
 
 <form id="search" action="{% url 'load_template' 'search' %}" title="Search OMERO">
 
+    {% if elasticsearch_url %}
+      {% include "webclient/base/includes/search_field_elastic.html" %}
```
joshmoore, line 26: could we potentially extract the field below to it's own view or include, and then rather than "elasticsearch_url" the proper could be: "omero.web.searchbox" and the default is the legacy code that gets extracted?

pwalczysko:
@will-moore would it be possible to consider closing this PR ? The fact that search for IDs is not possible on merge-ci is complicating my testing workflows

will-moore:
Sure - It should be working (I even added support for search by ID) but I haven't tested recently and I'm not working on it just now so I'll close it until I can take a close look...

will-moore:
Initial phase of search-engine use in IDR will be via idr-gallery and not in webclient. Closing for now...

PR #372, opened 2022-05-18
Type error fix(?) for WSGI threads

barrettMCW:
When trying to set omero.web.wsgi_threads in docker it seems to pass a string instead of an integer, causing a type error (new to python3 I think?) I haven't pulled this or tested it but I don't see why this wouldn't fix it :p

### DIFF
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -1236,7 +1236,7 @@ def check_worker_class(c):
 
 
 def check_threading(t):
-    t=int(t)
+    t = int(t)
     if t > 1:
         try:
             import concurrent.futures  # NOQA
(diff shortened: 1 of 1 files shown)


will-moore (approved):
Looks good, thanks!

sbesson:
Although it's almost a one-line fix, @barrettMCW would you be able to fill and return the [OME CLA](https://ome-contributing.readthedocs.io/en/latest/cla.html)? Once done we can get this integrated into the mainline and likely released immediately as a patch release @will-moore ?

will-moore:
I'd like to get https://github.com/ome/omero-web/pull/369 included in the next release too.

barrettMCW:
No problem, just sent it to contributors@openmicroscopy

PR #385, opened 2022-07-22
Show fileset ID

will-moore:
See https://github.com/ome/omero-cli-zarr/pull/123#pullrequestreview-1032712957

It is useful for webclient users to see the Fileset ID for a selected Image.
We already load fileset info (file paths) in a file-paths info panel, so this seems a logical place to load and display fileset ID:

The tooltip for the button that launches this panel also mentions Fileset ID.

Screenshot 2022-07-22 at 12 04 40

### DIFF
--- a/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
+++ b/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
@@ -355,7 +355,7 @@
     {% if not share %}
     {% if image.showOriginalFilePaths %}
       <!-- show original file paths -->
-      <button id="show_fs_files_btn" class="btn silver btn_fspath" title="Show file paths on server">
+      <button id="show_fs_files_btn" class="btn silver btn_fspath" title="Show Fileset ID and file paths on server">
         <span></span>
       </button>
     {% endif %}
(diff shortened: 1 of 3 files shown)


sbesson:
Thanks, I am unaware of an alternate location in the web client which would expose Fileset information and would be more suitable.

Do we foresee adding more metadata about the fileset in the future? In this case, would it make sense to structure the top-level panel similarly to other elements i.e.

```
Fileset ID: <id>
Owner: <owner name>
File count: <count>
```

will-moore:
I don't think we need to show the Owner (can't imagine it's different from the Image owner very often) and the File count is already shown.

pwalczysko:
Works as expected with multi- and single-file filesets.
Maybe RFE would be to do one of the following:
1. de-capitalize the word "Image" in the "Image file", see screenshot (because the capital image is not proper imho as well as it suggests that the important bit here is an "Image", whereas the number shows number of **files**
2. Go partially for the suggestion of @sbesson in https://github.com/ome/omero-web/pull/385#issuecomment-1199269707 - I would also leave the "Owner" info out though, as it will add too much burden on the dialog unecessarily

I realize this is not a topic of this PR, but the whole header of the "Paths" dialog could do with some reformatting and unification
Screenshot 2022-08-02 at 13 20 16

pwalczysko (changes requested):
Please address https://github.com/ome/omero-web/pull/385#issuecomment-1202419052

will-moore:
Last commit looks like this:

Screenshot 2022-08-02 at 13 57 44

pwalczysko (approved):

Issue #379, opened 2022-06-30
Server is not used by default when authenticating via GET param


hpawe01:
Hello,

I want to login to the webclient using the `bsession` query param:
https://github.com/ome/omero-web/blob/d4f857d2f021d032734ac314c651454172a75369/omeroweb/decorators.py#L424-L440

This is not working because of the lines before:
https://github.com/ome/omero-web/blob/d4f857d2f021d032734ac314c651454172a75369/omeroweb/decorators.py#L409-L420

I need to explicitly provide `server=1` as query param to make it work, although @will-moore states [here](https://github.com/ome/omero-iviewer/issues/197#issuecomment-379192508) that we shouldn't need to set it for the default configuration. This is only true, if we already have a session - then the `server_id` is retrieved from the `connector` which comes from the `session`: https://github.com/ome/omero-web/blob/d4f857d2f021d032734ac314c651454172a75369/omeroweb/decorators.py#L406

As I said, it is working when I provide `server=1` but the problem is, that this same query param is used for something else in [OMERO.iviewer](https://github.com/ome/omero-iviewer/blob/f98323a4bf24e3474537bd79195283e909dac78c/src/app/context.js#L431-L445). So when I want to open an image in OMERO.iviewer with "auto-login" via `bsession`, I could
- add `server=1`, then the login works but the the viewer crashes
- omit `server=1`, then the login will not work but the viewer would

I think we should use the default `server_id` when the param `server` is missing but I don't really know how to get it. Otherwise I would create a pull request.

will-moore:
Hi, apologies for delayed response.
I think the best solution here is to update iviewer to use a different parameter instead of `server`, since this is really a developer option and changing this won't break any use-case.
I've opened a PR to rename it to `host` at https://github.com/ome/omero-iviewer/pull/414

will-moore:
"don't use the server by default" - you mean that we only check `?server=1` if `connector` is None?
You would prefer that we always check the GET request for `server` first?

will-moore:
The iviewer fix at https://github.com/ome/omero-iviewer/pull/414 is released. Omero-web fix will be in next release of that too...

will-moore:
The fix in https://github.com/ome/omero-web/pull/386 has now been released in omero-web.

PR #386, opened 2022-07-22
Use server=1 in login_required if only 1 server

will-moore:
See https://github.com/ome/omero-web/issues/379#issuecomment-1175236423

To test:

- Setup: you need an omero-web where only 1 server is configured. TODO: update merge-ci config temporarily to only show 1 server
- look up an Image ID on the server (e.g. via Insight or webclient in another brower)
- Get a session ID, e.g via cli to the server above:

```
$ omero login
$ omero sessions list
```

- Make sure you are not logged in to webclient
- Go to: https://merge-ci.openmicroscopy.org/web/webclient/img_detail/IMAGE_ID/?bsession=SESSION_ID

You should be able to see the image.

### DIFF
--- a/omeroweb/decorators.py
+++ b/omeroweb/decorators.py
@@ -37,7 +37,7 @@ from django.urls import reverse, resolve, NoReverseMatch
 from django.core.cache import cache
 
 from omeroweb.utils import reverse_with_params
-from omeroweb.connector import Connector
+from omeroweb.connector import Connector, Server
 from omero.gateway.utils import propertiesToDict
 from omero import ApiUsageException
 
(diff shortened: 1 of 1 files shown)


pwalczysko:
Works as expected, tested on two images on merge-ci.

pwalczysko (approved):

will-moore:
cc @chris-allan This look good to you?

omeroweb/decorators.py
```diff
@@ -416,8 +416,14 @@ def get_authenticated_connection(self, server_id, request):
                 try:
                     server_id = request["server"]
                 except Exception:
-                    logger.debug("No Server ID available.")
-                    return None
+                    # If only 1 server, use server=1
```
chris-allan, line 419: `Server` operates like an iterable so you can `list(Server)` if you wish and access the `id` via that attribute. For example:

```
servers = list(Server)
if len(servers) == 1:
    server_id = servers[0].id
else:
    return None
```

Honestly up to you. It achieves the same thing but has the advantage of not depending on the internals of `Server`. Not sure if we'd ever change that.

chris-allan (approved):
No conceptual issues from my side. One might even argue that using the first server as the default is always a good idea regardless of the server count. Though I've only ever seen an omero-web in the wild with more than one server configured a couple of times so it's probably not worth debating.

chris-allan (approved):

jburel:
Do we want to include the change in a point release?

will-moore:
Yes, I started a changelog already... will open a PR...

jburel:
Thanks

PR #392, opened 2022-08-24
omero table sticky header

will-moore:
Simple fix to a problem I recently had checking OMERO.table data in IDR: When you scroll the table, the header gets lost so you don't know which columns are which.
This makes the header "sticky" so that it isn't scrolled off the top of the page.

To test:
 - Open a large table in web, e.g (random data): https://merge-ci.openmicroscopy.org/web/webclient/omero_table/1508508/ (linked to Plate https://merge-ci.openmicroscopy.org/web/webclient/?show=plate-13855) - user-3.
 - Scroll the page, header should remain visible at the top of the page:

Screenshot 2022-08-24 at 12 55 31

cc @muhanadz

### DIFF
--- a/omeroweb/webclient/templates/webclient/annotations/omero_table.html
+++ b/omeroweb/webclient/templates/webclient/annotations/omero_table.html
@@ -13,6 +13,14 @@
                 border-collapse: collapse;
                 border-spacing: 0;
             }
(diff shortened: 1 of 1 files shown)


pwalczysko:
LGTM

Issue #304, opened 2021-07-07
edit channel names deadlock PSQLException


will-moore:
From e-mail today:

```
Internal Server Error: /webclient/edit_channel_names/4461817/

TryAgain at /webclient/edit_channel_names/4461817/
exception ::omero::TryAgain
{
    serverStackTrace = ome.conditions.TryAgain: Hibernate operation: could not update: [ome.model.core.LogicalChannel#9707608]; SQL [update logicalchannel set contrastMethod=?, external_id=?, group_id=?, owner_id=?, update_id=?, detectorSettings=?, emissionWaveUnit=?, emissionWave=?, excitationWaveUnit=?, excitationWave=?, filterSet=?, fluor=?, illumination=?, lightPath=?, lightSourceSettings=?, "mode"=?, name=?, ndFilter=?, otf=?, photometricInterpretation=?, pinHoleSizeUnit=?, pinHoleSize=?, pockelCellSetting=?, samplesPerPixel=?, version=? where id=?]; ERROR: deadlock detected
  Detail: Process 22838 waits for ShareLock on transaction 13888063; blocked by process 9207.
Process 9207 waits for ShareLock on transaction 13888062; blocked by process 22838.
  Hint: See server log for query details.
  Where: while updating tuple (23635,1) in relation "logicalchannel"; nested exception is org.postgresql.util.PSQLException: ERROR: deadlock detected
  Detail: Process 22838 waits for ShareLock on transaction 13888063; blocked by process 9207.
Process 9207 waits for ShareLock on transaction 13888062; blocked by process 22838.
  Hint: See server log for query details.
  Where: while updating tuple (23635,1) in relation "logicalchannel"
        at org.springframework.jdbc.support.SQLErrorCodeSQLExceptionTranslator.doTranslate(SQLErrorCodeSQLExceptionTranslator.java:263)
        at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:73)
      ...

 serverExceptionClass = ome.conditions.TryAgain
[...]

PR #394, opened 2022-08-25
Update channels in batches for Dataset, show progress

will-moore:
Fixes long-standing issues with updating channel names: https://github.com/ome/omero-web/issues/304

 - Hide the 'submit' button so users can't make synchronous calls (causes exceptions above)
 - Update Images in batches, instead of whole Plate/Dataset in one call (which often times-out for web)
 - Show a progress bar
 - Support Cancel during update

Screenshot 2022-09-01 at 23 07 12

To test:
 - Compare renaming of channels with / without this PR
 - Rename single Images, Dataset, Plate
 - Test that Cancel button interrupts renaming

### DIFF
--- a/omeroweb/webclient/templates/webclient/annotations/includes/core_metadata.html
+++ b/omeroweb/webclient/templates/webclient/annotations/includes/core_metadata.html
@@ -96,6 +96,8 @@
                             <button name="confirm_apply" type="submit" value="apply" title="Save and apply to all Images">Continue</button>
                         </span>
                         <button type="reset" name="cancel" title="Cancel">Cancel</button>
+                        <div id="channel_edit_progress_label" style="display:none">Renaming images...</div>
+                        <progress id="channel_edit_progress" max="100" value="20" style="width:100%; display:none">20%</progress>
                     </form>
                     {% if image.canEdit %}
                         <button id="editChannelNames" class="btn silver btn_edit" 
--- a/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
+++ b/omeroweb/webclient/templates/webclient/annotations/metadata_general.html
@@ -296,6 +296,8 @@
                     $chNameForm.find(".confirmButtons").hide();
                     $("#channel_names_display").show();
                     $("#editChannelNames").show();
+                    $("#channel_edit_progress").hide();
+                    $("#channel_edit_progress_label").hide();
                     $chNameForm.hide();
                 }
                 // Workflow starts by displaying the form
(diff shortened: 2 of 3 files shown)


pwalczysko:
Metrics:

Comparison:
Setup 1: Connect merge-ci webclient with this PR to workshop server. 
Setup 2: Connect workshop webclient without this PR to workshop server.

INMAC plate (16 x 24 wells, 1 image per well, 3 channels per image)
Setup 1 (with PR): ``1.37 min``
Setup 2 (no PR):   ``1.92 sec``

siRNA-HeLa dataset (33 images, 4 channels per image):
Setup 1 (with PR): ``3.84 sec``
Setup 2 (no PR):   ``1.02 sec``

It seems that for small Datasets the performance loss is acceptable, but large(r) plates have some 60x slower performance now.

pwalczysko:
@will-moore could you please explain the expected behaviour with this PR when

1. Select an image inside a plate, edit a channel
2. Click Apply to All
3. Before the process finishes, click on another well and select another image in that well
4. Observe that the progress bar is gone from RHP, but also that all the images have the channels renamed as expected

Not sure if by clicking onto an image the behaviour is reverted to the behaviour without this PR ?
It feels a bit odd...

will-moore:
This could be a case where all the Images in the Plate are linked to the same 2 or 3 channels in the DB, so after you've re-named the channels for the first Image, you've effectively renamed them for all other Images in the Plate.

pwalczysko:
> This could be a case where all the Images in the Plate are linked to the same 2 or 3 channels in the DB, so after you've re-named the channels for the first Image, you've effectively renamed them for all other Images in the Plate.

I do not think this is the case.
This is because I have a counter-example, in the form of the siRNA-HeLa dataset. Albeit much smaller, I can still rename and whilst the progress bar is still in progress, click on another image. I can see that the newly clicked image had the old channel name for a short while, but quickly got updated to a new one. But of course, on the newly clicked image, there was no progress bar in RHP.
It seems like the progress bar is a nuisance which can be gotten rid off by clicking on another image - the process will still finish successfully, and, it feels, even faster... 
I do not believe that the images inside the siRNA.. dataset are linked to the same channels ?

will-moore:
Ah - OK, that's actually better than I thought. It turns out that by migrating away to view another Image, the renaming calls are not cancelled. So they go to completion. I hadn't tested but just assumed they would stop.
Only hitting the `Cancel` button stops them.

will-moore:
Hmmm - it seems that renaming Plates without this PR is really fast. And even Datasets rename quick enough (on a non-production server).
So it may be that this PR is not really needed - or it *might* not be needed on production systems once https://github.com/ome/omero-py/pull/331 is released and deployed.
Let's wait for that to get on to nightshade. Then we can do some more realistic testing...

jburel:
Could you exclude this PR so we can do some tests with only omero-py?

will-moore:
`--exclude`

will-moore:
Having upgraded our production server with `omero-py` 5.12.0 including https://github.com/ome/omero-py/pull/331, the channel renaming for a Dataset is MUCH faster. So we really don't need all the batch update and progress bar in this PR...
Will open another for a much simpler spinner etc.

pwalczysko:
> the channel renaming for a Dataset is MUCH faster.

Can confirm that. For example a pre-test without https://github.com/ome/omero-py/pull/331 was completely failing to rename the channels on a very small Dataset logged in as admin because of 1 min timeout.
Now, I have 689 ms on the same Dataset with https://github.com/ome/omero-py/pull/331.
Congrats @will-moore

PR #397, opened 2022-09-07
add whitenoise as a dependency

jburel:
See https://github.com/ome/omero-documentation/pull/2270 for background discussion
If the proposed change is merged. The installation documentation will need to be adjusted

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,7 @@ setup(
         "Django>=3.2,<4.0",
         "django-pipeline==2.0.7",
         "django-cors-headers==3.7.0",
+        "whitenoise>=5.3.0",
         "gunicorn>=19.3",
         "omero-marshal>=0.7.0",
         "Pillow",


will-moore:
Just adding `whitenoise` to an existing `omero-web` dev environment... It is small and doesn't include any other depenencies, so I'm happy to add it as a dependency.

```
$ pip install whitenoise
Collecting whitenoise
  Downloading whitenoise-6.2.0-py3-none-any.whl (19 kB)
Installing collected packages: whitenoise
Successfully installed whitenoise-6.2.0
```

will-moore (approved):
LGTM

jburel:
see https://github.com/ome/omeroweb-install/pull/85

joshmoore:
:+1: Thanks, @jburel. Guess the only questions are if there are any other dangling deps like this and if we know why it had such a special status.

jburel:
The only one that could be included is ``django-redis==5.0.0``

joshmoore:
That seems more clearly like an add-on.

jburel:
> That seems more clearly like an add-on.

agree

jburel:
Suggesting to push a tag: 5.16.0 since we have a dependency injection
Since we avoid to update the upgrade installation instruction i.e. https://github.com/ome/omero-documentation/pull/2270
Any objections?

will-moore:
👍

sbesson:
Assuming we are injecting dependency, would https://github.com/ome/omero-web/pull/415 also be a candidate for 5.16.0?

jburel:
It makes sense

PR #411, opened 2022-09-30
Thumbnail zooming

will-moore:
This PR cleans-up the thumbnail zooming logic in webclient.

Instead of using the sizeX and sizeY values to calculate the aspect ratio for every thumbnail and JavaScript sets the width and height on the fly while dragging the thumbnail slider, we now use CSS to do achieve the same effect which is much simpler and might be faster/smoother.

To test:
 - Should see no change in behaviour for thumbnail zooming
 - The table layout icon has a small fix (see screenshots below)

NB: Previously this PR didn't load sizeXYZ values from OMERO until the user clicked on table layout to display them, but this was added complexity for negligible performance gain so it has been removed.

### DIFF
--- a/omeroweb/webclient/static/webclient/css/layout.css
+++ b/omeroweb/webclient/static/webclient/css/layout.css
@@ -1582,6 +1582,17 @@
         position: relative;
         width: 100%;
     }
+    #dataIcons li {
+        width: var(--iconSize);
+        height: var(--iconSize);
+    }
+    #dataIcons img{
+        width: 100%;
+        height: 100%;
+        /* fit the image to container, regardless of aspect ratio */
+        object-fit: contain;
+        object-position: center top;
+    }
 
     #dataIcons .ratingFilter_hidden {
         display: none;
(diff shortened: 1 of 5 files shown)


pwalczysko:
Tested on 
Screenshot 2022-12-16 at 11 17 10

As discussed with @will-moore ^^^, the necessity of the refresh is not a great user experience, and the performance enhancement we saw on merge-ci is less than 10%, with total times well under 1 sec, i.e. not humanly perceptible.
The message wording ``...the data...`` is confusing, I would remove this, but most probably, removing the whole necessity of the refresh is the better option.

pwalczysko (changes requested):
some suggestions in https://github.com/ome/omero-web/pull/411#issuecomment-1354609284

will-moore:
Thanks @pwalczysko: fixed and re-deployed (and updated description).
Now, the only visible change in behaviour is the tiny fix to layout buttons. When on table layout:

Before:

Screenshot 2022-12-16 at 13 15 37

After:

Screenshot 2022-12-16 at 13 19 48

jburel:
Excluding ahead of the dependencies upgrade work

--exclude

jburel:
``exclude`` label removed

pwalczysko:
lgtm

pwalczysko (approved):

PR #415, opened 2022-10-10
Update to a suitable version of the concurrent-log-handler

chris-allan:
We have a vendored version of ConcurrentLogHandler 0.9.1 [^1] from omero-py that we use.  This version is from 2013, has a few bugs, and doesn't work properly with the encoding changes [^2][^3] made in Python 3.10.

A new version [^4] whose lineage is that of the now unmaintained ConcurrentLogHandler is now available which resolves several of the aforementioned bugs, supports Python 3.10, and is drop in replacement API compatible with the previous version.  This commit adds a dependency for this new version and swaps the default logger class for it in configuration.

  [^1]: https://pypi.org/project/ConcurrentLogHandler/
  [^2]: https://peps.python.org/pep-0597/
  [^3]: https://github.com/python/cpython/pull/19481
  [^4]: https://pypi.org/project/concurrent-log-handler/

### DIFF
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -87,7 +87,7 @@ FULL_REQUEST_LOGFORMAT = (
     " HTTP %(status_code)d %(request)s"
 )
 
-LOGGING_CLASS = "omero_ext.cloghandler.ConcurrentRotatingFileHandler"
+LOGGING_CLASS = "concurrent_log_handler.ConcurrentRotatingFileHandler"
 LOGSIZE = 500000000
 
 
--- a/setup.py
+++ b/setup.py
@@ -53,6 +53,7 @@ setup(
         # requires Ice (use wheel for faster installs)
         "omero-py>=5.7.0",
         # minimum requirements for `omero web start`
+        "concurrent-log-handler>=0.9.20",
         "Django>=3.2,<4.0",
         "django-pipeline==2.0.7",
         "django-cors-headers==3.7.0",


chris-allan:
Assuming there are no issues with this we could probably go ahead with removing the vendored `cloghandler` from upstream `omero-py`; as far as can tell it's only used by `omero-web`. Might also loose the vendored `portalocker` for https://pypi.org/project/portalocker/ and then test the impact on the wider OMERO server codebase.

sbesson (commented):
At least conceptually, my position is that any like-for-like replacement of the forks which are currently shipped under [omero_ext](https://github.com/ome/omero-py/tree/master/src/omero_ext) by upstream libraries maintained by the community is advantageous. Adding a new dependency is extremely low-cost thanks to all the infrastructure work that went into extracting and packaging `omero-py/omero-web` and allows us to benefit from several features including security detection etc.

Overall no objection from my side but a few general comments:

- based on a similar attempt with [argparse](https://github.com/ome/omero-py/pull/115), I would not exclude we have been relying on some API feature that is only present in our vendored version so I would definitely give this PR a few runs in the nightly CI builds
- in terms of functional testing, we probably want to confirm the log rotation works as expected. This might be best achieved by temporary enabling `omero.web.debug=True` on a deployment as this has the best potential of generating GBs of logs
- while looking at this code path recently, I was wondering how easy it would to add extra configuration for the log rotation (typically log size and/or number of concurrent logs) Since all the `omero.web.*` properties are defined in `settings.py` after the logging initialization, this seems a bit more involved that just adding extra properties.
- I also vote for the long-term removal of `omero_ext.cloghandler` from `omero-py` but I would probably argue for a more traditional lifecycle of deprecation in 5.x.x and removal in 6.0.0 in case someone depends on it

chris-allan:
Diff between our vendored version and the original 0.9.1 for anyone concerned about specific semantics we've added that we may be relying upon:

```
callan@behemoth:~/tmp$ curl -J -O -L -s 'https://files.pythonhosted.org/packages/fd/e5/0dc4f256bcc6484d454006b02f33263b20f762a433741b29d53875e0d763/ConcurrentLogHandler-0.9.1.tar.gz'
callan@behemoth:~/tmp$ tar zxf ConcurrentLogHandler-0.9.1.tar.gz
callan@behemoth:~/tmp$ diff -Nau ConcurrentLogHandler-0.9.1/src/cloghandler.py \
  ~/code/omero-py/src/omero_ext/cloghandler.py
--- ConcurrentLogHandler-0.9.1/src/cloghandler.py       2013-07-11 02:23:26.000000000 +0000
+++ /home/callan/code/omero-py/src/omero_ext/cloghandler.py     2020-01-24 09:52:22.909358076 +0000
@@ -44,8 +44,10 @@
 This module supports Python 2.6 and later.

 """
+from __future__ import absolute_import


+from builtins import range
 __version__  = '0.9.1'
 __revision__  = 'lowell87@gmail.com-20130711022321-doutxl7zyzuwss5a 2013-07-10 22:23:21 -0400 [0]'
 __author__ = "Lowell Alleman"
@@ -73,7 +75,7 @@
 # clobbering that the builtin class allows.

 # sibling module than handles all the ugly platform-specific details of file locking
-from portalocker import lock, unlock, LOCK_EX, LOCK_NB, LockException
+from .portalocker import lock, unlock, LOCK_EX, LOCK_NB, LockException


 # Workaround for handleError() in Python 2.7+ where record is written to stderr
 ```

joshmoore (approved):
MSTM. Since there aren't tests specifically on log rolling with the production settings (for good reason), we might deploy somewhere and trigger a rollover. The merge-ci web doesn't have enough logging at the moment to do so, but then again neither does Processor or co. (Did you consider the same PR for omero-py, @chris-allan?)

chris-allan:
> Did you consider the same PR for omero-py, @chris-allan?

I wasn't even aware we're using it in `omero-py`. A quick `git grep` didn't bring up any imports of `cloghandler` or references to `ConcurrentRotatingFileHandler`. Am I missing something?

joshmoore:
> Am I missing something?

No, glancing quickly, you're probably right. I very much assumed we were using the same infra both the Python services as well.

jburel (approved):

jburel:
Changes included in 5.16.0

sbesson:
> Changes included in 5.16.0

Thanks @jburel

PR #421, opened 2022-12-07
Add Python 3.10 to the testing matrix

sbesson:
As OMERO.web is starting to be deployed on distributions like Ubuntu 22.04 where Python 3.10 is the default Python version,  this proposes to add this environment to  the testing matrix so that we have some minimal coverage.

Proposed changes:
- updates the Tox dependencies to consumes the artifacts of https://github.com/ome/zeroc-ice-py-github-ci/releases/tag/0.2.0
- adds Python 3.10 to the Tox configuration as well as the GitHub workflow configuration
- fixes an unit test failing on Python 3.10

Incidentally, this makes a few updates/deprecations:
- the base OS under which the workflow is executed is upgraded to Ubuntu 20.04 (Ubuntu 18.04 is going EOL in April 2023)
- Python 3.6 (EOL since December 2021) is removed from the Tox configuration

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -14,7 +14,8 @@ jobs:
           - '3.7'
           - '3.8'
           - '3.9'
-    runs-on: ubuntu-18.04
+          - '3.10'
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


jburel (approved):
Everything is green

PR #423, opened 2022-12-08
Right panel spinner

### LINKED ISSUE #400: Spinner right hand panel
* Select a large number of objects e.g. 500 Possible with https://github.com/ome/omero-web/pull/225
* Right-hand panel is blank for a few seconds. 
* A spinner should be added to indicate that the loading is happening

will-moore:
Fixes #400 

This adds a spinner to the "General" right hand panel and the "Preview" panel, and updates the one on "Acquisition" panel to look the same.

The spinner fades in over 2 seconds, so that if the panel content loads quickly you won't see it (avoids an ugly "flash" of spinner).

To test:

 - Select single objects/images or multiple objects/images and move between right tabs. Spinner should fade in when loading.

### DIFF
--- a/omeroweb/webclient/static/webclient/css/layout.css
+++ b/omeroweb/webclient/static/webclient/css/layout.css
@@ -319,7 +319,19 @@
 	        }
         
         
-
+    .right_panel_spinner {
+        text-align: center;
+        margin-top: 25px;
+        font-size: 18px;
+        /* transition: opacity 2s; */
+        display: none;
+    }
+    /* .right_panel_spinner.fade {
+        opacity: 1;
+    } */
+    .right_panel_spinner span {
+        opacity: 0.5;
+    }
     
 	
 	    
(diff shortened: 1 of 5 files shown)


pwalczysko:
I can see the new-style spinner in the Acquisition tab. It is more central than it used to be and it ``fades in``. But no spinner at all on General and Preview tabs. Even when the throttling in FF was switched on and cache cleared. The throttling caused the General to be loading very slow indeed, much longer than 2 secs, but still no spinner observed.

pwalczysko (changes requested):
Please investigate https://github.com/ome/omero-web/pull/423#issuecomment-1344143164

will-moore:
The lack of changes to the "General" tab is caused by `mapr` overwriting that code.
It should be fixed by https://github.com/ome/omero-web/pull/425 and https://github.com/ome/omero-mapr/pull/74

pwalczysko:
Tested on merge-ci, indeed, the spinner is now in all three tabs.
LGTM

pwalczysko (approved):

will-moore:
Thanks Petr.
Since this is conflicting with #425 now, I'll merge and resolve that...

will-moore:
Ah, OK - apologies for that.
Please go ahead and review the code and I'll open another PR to address any issues.

PR #424, opened 2022-12-09
Mark unused templates as deprecated

will-moore:
This PR deletes a bunch of old templates used for the old home page which showed a tag cloud and recent imports.

Not used in a LONG time. Also duplicates a lot of code. Good to clean up.

### DIFF
--- a/omeroweb/webclient/templates/webclient/base/base.html
+++ b/omeroweb/webclient/templates/webclient/base/base.html
@@ -26,6 +26,12 @@
 
 
 {% block script %}
+    <!--
+        This template is no-longer used by omero-web
+        Since it will be removed soon, we need to be sure that any users are warned:
+    -->
+    <script>alert("WARNING: Django template webclient/base/base.html has been deprecated")</script>
+
     {{ block.super }}
     <!-- required for the script_launch html below -->
     {% include "webclient/base/includes/script_launch_head.html" %}
--- a/omeroweb/webclient/templates/webclient/index/index.html
+++ b/omeroweb/webclient/templates/webclient/index/index.html
@@ -27,6 +27,13 @@
 {% endblock %}
 
 {% block script %}
+
+    <!--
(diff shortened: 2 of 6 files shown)


jburel:
We do not use those templates. Technically speaking, someone might have adjusted the webclient and may still be using it.
I think we should follow the same pattern as we do for the code. Mark the template as ``deprecated`` for this release, and remove them the next one.

jburel (changes requested):
We do no longer use those templates. Technically speaking, someone might have adjusted the webclient and may still be using it.
I think we should follow the same pattern as we do for the code. Mark the template as ``deprecated`` for this release, and remove them the next one.

will-moore:
I can't find an official way of marking Django templates as deprecated, but I could add some appropriate comments to the templates.
I guess I could add a `<script>` with something like:

```
console.warn("OMERO.web template at webclient/base/base.html is deprecated")
```

or even use `alert("OMERO.web template at webclient/base/base.html is deprecated")` to be *really* sure that it gets seen?

jburel (approved):
Thanks @will-moore for making the adjustments

jburel:
Is the last commit expected on this PR?

will-moore:
@jburel Thanks - This is the last PR for 5.17.0 so I updated the Changelog accordingly in this PR.
If it looks good, please go ahead and merge when ready - Thanks

sbesson:
Have you guys also evaluated https://github.com/ome/omero-web/pull/431 which could be a possible candidate for 5.17.0?

I assume the goal is to cut a release asap to start the testing of #433?

jburel:
@sbesson including https://github.com/ome/omero-web/pull/431 in the release makes sense
As discussed in previous meeting, the aim is to have a release before the upgrade work so we have a reference point.
@will-moore: Could you create a PR with the changelog entry only so it is clearer to review and add #431?

PR #425, opened 2022-12-09
Add hook for mapr or other app to extend webclient main page

will-moore:
As noted in https://github.com/ome/omero-web/pull/423#issuecomment-1344143164
changes to the webclient `templates/webclient/data/includes/mapr_extension.html` file are completely overwritten by `mapr`, so any changes to that file must previously be duplicated in the corresponding file in `omero-mapr`.
Mapr needed to overwrite `update_metadata_general_tab()` to provide custom URL loading when custom items are selected.

This PR provides a function `OME.getCustomRightPanelUrl()` which can be overwritten, instead of replacing the whole of `update_metadata_general_tab()`.

This change is used by mapr in https://github.com/ome/omero-mapr/pull/74/files

To test:
 - with mapr installed (no config needed) with the branch above, the right panel should work as before (including the creation of custom right-panel links) E.g. https://merge-ci.openmicroscopy.org/web/webclient/?show=image-18188
 - Right panel should show the fading-in loading spinner (previously overwritten by mapr)


omeroweb/webclient/templates/webclient/base/base_container.html
```diff
@@ -289,8 +289,9 @@
         {% include rt.include %}
     {% endfor %}
     
-    
-    
+    <!-- allow omero-mapr OR another app to include custom code -->
+    {% include "webclient/data/includes/mapr_extension.html" %}
```
sbesson, line 293: In general :+1: for adding extension points rather than having apps fully overwriting the templates (which quickly turns into a maintenance hell)

I am a bit concerned about the introduction of an extension point explicitly named after a specific app though as introducing this type of circular dependency will eventually cause issues. Is it not possible to override the templates using `extends` and  `block` as defined in https://docs.djangoproject.com/en/3.2/howto/overriding-templates/#extending-an-overridden-template?

will-moore: mapr is not extending an existing template to create it's own page (which is what those docs refer to).
It is "hijacking" the template that is included in the "host" webclient page, so change the "host" app behaviour.

When Django looks up a template, e.g. to `{% include "webclient/data/includes/right_plugin.general.js.html" %}` in the main webclient page, it goes through all the apps in it's installed apps list until it finds a match.

We deliberately install the omero-web apps *after* any user-installed apps (https://github.com/ome/omero-web/blob/93f89eddfe15884e776a854d6c8d44c2b587903b/omeroweb/settings.py#L1542) to allow user-installed apps to overwrite a template. mapr has a template at the matching path: `webclient/data/includes/right_plugin.general.js.html` so this is chosen instead of the webclient one.

I don't think that Django provides a mechanism for this kind of behaviour.

However, we could add a mechanism in webclient to allow other apps to inject their code with a configuration (similar to the addition of code for centre / right panels in webclient).
This would just mean that we need to add an additional setting for mapr when installing, but I guess this is easy enough.

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#1286](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/1286/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/1286/consoleText) for more details.
Possible conflicts:
  - PR #225 will-moore 'Query string ids'
    - omeroweb/webclient/templates/webclient/data/includes/right_plugin.general.js.html
  - PR #423 will-moore 'Right panel spinner'
    - omeroweb/webclient/templates/webclient/data/includes/right_plugin.general.js.html

--conflicts

omeroweb/webclient/templates/webclient/data/includes/mapr_extension.html
```diff
@@ -0,0 +1,11 @@
+
+<!-- 
+This is a hook for use by omero-mapr OR another omero-web app to add custom
+code to the <head> of the webclient main page.
```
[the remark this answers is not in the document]
will-moore: @joshmoore @sbesson Was the *naming* of `mapr_extension.html` main problem with this PR in it's original form, or was it the fact that this extension point could only be used by 1 app at a time?

If it's only the naming that's the issue, I'm happy to re-name that extension point (e.g. to `head_extension.html`) and we could avoid the config that's needed to allow multiple apps to extend (which @sbesson is concerned about the config at https://github.com/ome/omero-mapr/pull/74#issuecomment-1352605507).

Do we need to chat about this to discuss the various options?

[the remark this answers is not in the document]
joshmoore: I don't think the naming is the main problem as much as having a good name helps us to understand what the costs/benefits of this strategy actually are. Happy to jump on a call next week Monday or Tuesday if you're still around.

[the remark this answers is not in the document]
will-moore: I'm away next week I'm afraid. 
The name `mapr_extension.html` was meant to warn that this is being used by `mapr`, so you should only use it if you don't have `mapr` installed (which is explained in the `mapr_extension.html` file itself).

will-moore:
One other option that could allow a single template to be included from each app without configuration:

For each app in INSTALLED_APPS, look for e.g. `/app/templates/app/include/webclient_head.html` and if found, then include it in the webclient main page.

will-moore:
That option is now implemented in the last 2 commits above.

will-moore:
I updated the description above to match the current behaviour. cc @sbesson

omeroweb/webclient/templates/webclient/data/includes/mapr_extension.html
```diff
@@ -0,0 +1,11 @@
...
+
+Simply add a file with the same path under the app/templates directory:
+/app/templates/webclient/data/includes/mapr_extension.html
```
sbesson, line 7: Is this still correct with the latest version of this PR? This is what this file and the description suggest but from the `util.py` changes, I would expect the file should be called `webclient_head` ?

omeroweb/webgateway/util.py
```diff
@@ -238,3 +243,42 @@ def points_string_to_XY_list(string):
         x, y = xy.split(",")
         xyList.append((float(x.strip()), float(y.strip())))
     return xyList
+
+
+def get_app_labels():
+    """For each installed app, try to load AppConfig and return label"""
+
+    labels = []
+    for app in settings.INSTALLED_APPS:
```
sbesson, line 252: Out of curiosity, what is the contract for the order in which these apps are resolved? Alphabetical order? Order in the configuration?

And assuming several apps would try to overwrite the same variable, would the last app win?

omeroweb/webclient/templates/webclient/data/includes/mapr_extension.html
will-moore: Apologies, I forgot to delete that file. It's not needed any more.

jburel:
Excluding ahead of dependencies upgrade work

--exclude

knabar:
I might not have fully wrapped my head around all changes, but I'm not yet convinced we need an additional way to include custom plugin files, versus just utilizing the Django template inheritance mechanism.

For example, if `right_plugin.general.js.html` introduced a block in the new function, even multiple plugins could override/extend it:
```django
// This can be overwritten by plugins to provide custom URL for right panel
// NB: The returned html should have an element <div id="object-id">oid</div>
// where the oid matches the oid from this function
{% block custom-right-panel-url %}
OME.getCustomRightPanelUrl = function getCustomRightPanelUrl(oid) {
    return;
}
{% endblock %}
```
A plugin (or multiple plugins) would have its own `right_plugin.general.js.html`:
```django
{% extends "right_plugin.general.js.html" %}
{% block custom-right-panel-url %}
{{ block.super }}
(function () {
  var old = OME.getCustomRightPanelUrl;
  OME.getCustomRightPanelUrl = function getCustomRightPanelUrl(oid) {
      return old(oid) || some_new_value;
  }
})();
{% endblock %}
```

Similarly other template blocks in `right_plugin.general.js.html` could be added at the top level or at crucial points in the logic.

Again, there may be a limitation here that I'm not seeing, but right now a whole new way to include plugin functionality seems overkill to me.

pwalczysko:
tested on merge-ci, I can see both the fading-in spinner and a correct Mapr behaviour, with the custom external and internal links working.

lgtm

pwalczysko (approved):

will-moore (commented):
Just confirmed the expected behaviour of this PR on merge-ci *without* mapr change (excluded https://github.com/ome/omero-mapr/pull/74#issuecomment-1569810517).
Mapr works as expected but with the existing behaviour of overwriting the right panel (no loading spinner etc).

PR #429, opened 2022-12-19
Expand description of default_user and default_group search configurations

sbesson:
Clarify the expectations when setting the special values of 0 (default) and -1 for the `omero.web.search.default_user` and `omeor-web.search.default_group` properties introduced in #297

Additionally, this reveals a lack of symmetry between default user where `0` == `current user` and `-1` == `all users/members` and the default group where both `0` and `-1` are all groups. I assume in an upcoming breaking change, the logic could be reviewed so that `0` could select the current group.

### DIFF
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -972,13 +972,22 @@ CUSTOM_SETTINGS_MAPPINGS = {
         "SEARCH_DEFAULT_USER",
         0,
         int,
-        ("ID of user to pre-select in search form."),
+        (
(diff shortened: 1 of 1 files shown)


jburel:
It will make sense to also include this PR in the upcoming release. This is only doc for properties.
@pwalczysko are you okay with that? (you are marked as reviewer)

pwalczysko (changes requested):
One minor syntax suggestion, when accepted, ready to merge and release fmpov.

PR #431, opened 2022-12-22
Clean up favicon references

knabar:
* Removes duplicate `<link>` entry from `login.html` template
* Change link URL from `.../ome.ico` to the standard `/favicon.ico`, which then gets redirected to the configured icon (or default `ome.ico`)

Discussion on favicon links:
https://stackoverflow.com/questions/10218178/necessary-to-add-link-tag-for-favicon-ico

Potential remaining issues:
* Running OMERO.web below the web server root, since the `/favicon.ico` URL then no longer is at the root.  Supporting this would likely require changes to the nginx configuration.

@emilroz The previous PR #311 allowing customizing the icon did not change the `<link>` in the HTML templates and therefore relied on the browser default behavior of requesting `/favicon.ico` (unless I am missing something)

### DIFF
--- a/omeroweb/webclient/templates/webclient/login.html
+++ b/omeroweb/webclient/templates/webclient/login.html
@@ -69,8 +69,6 @@
 {% block head %}
     {{ block.super }}
 	
-    {% include "webgateway/base/includes/shortcut_icon.html" %}
-	
 	<script type="text/javascript" charset="utf-8">
 			$(function(){ $("label").inFieldLabels(); });
 		</script>
--- a/omeroweb/feedback/templates/base_error.html
+++ b/omeroweb/feedback/templates/base_error.html
@@ -10,11 +10,6 @@
     {% trans "OMERO.web - support" %}
 {% endblock %}
 
-{% block head %}
-    {{ block.super }}
-    {% include "webgateway/base/includes/shortcut_icon.html" %}
-{% endblock %}
-
 {% block body %}
 <div class="bodyWrapper">
     <div><h1>
--- a/omeroweb/feedback/templates/disabled.html
+++ b/omeroweb/feedback/templates/disabled.html
@@ -10,11 +10,6 @@
     {% trans "OMERO.web - support" %}
 {% endblock %}
 
-{% block head %}
-    {{ block.super }}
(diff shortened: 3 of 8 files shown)


omeroweb/webclient/templates/webclient/login.html
```diff
@@ -69,8 +69,6 @@
 {% block head %}
     {{ block.super }}
 	
-    {% include "webgateway/base/includes/shortcut_icon.html" %}
```
sbesson, line 72: I assume the duplicate `link` comes from the fact that this block is already included in https://github.com/ome/omero-web/blob/90dd8ad0e2d6b5ed83b9f76fde6c0d07a8b7e688/omeroweb/webgateway/templates/webgateway/core_html.html#L47-L50 from which this page is derived?

Looking at https://github.com/ome/omero-web/search?q=shortcut_icon, there are a few other places which seem to be re-including. Should all of these be cleaned up as part of this PR?

knabar: Good catch @sbesson, I had not been looking much past the login page itself.

sbesson (commented):
Thanks for the cleanup. Tested in a CI deployment including a custom web prefix (`/web`) after adding a new icon under the static and setting the favicon to via `omero.web.favicon_url`.

The following endpoints were tested and successfully displayed the new icon

- `/web/webclient/login`
- `/web/feedback/thanks`
- `/web/feedback/disabled`
- `/web/forgotten`
- `/web/webadmin/myaccount/edit/`
- `/web/webclient/script_ui/<id>/

After authentication, the regular `/web/webclient` does not display the custom favicon although the link to `/web/favicon.ico` can be found in the source page.

In all these pages, the duplicate `link` is now removed and the update of shortcut icon in favor of icon is inline with the recommendations in https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types.


The re

sbesson (approved):
Retested in a simpler environment than the merge-ci Web deployment with several OMERO.web apps and candidate PRs.

After installing the head of this branch and setting `omero.web.favicon_url`, confirmed that the `favicon.ico` is updated in all expected URLs.

Issue #333, opened 2021-11-15
Missing `jquery-ui` images


sbesson:
Since the upgrade of the OMERO.web 5.11.0 deployments to production, the nginx logs are regularly throwing warnings of style:

```
2021/11/15 13:48:47 [error] 3611#3611: *13742153 open() "/opt/omero/web/OMERO.web/var/static/3rdparty/jquery-ui-1.12.1/css/images/ui-icons_444444_256x240.png" failed (2: No such file or directory), client: 134.36.66.29, server: ome-ns-web, request: "GET /static/3rdparty/jquery-ui-1.12.1/css/images/ui-icons_444444_256x240.png HTTP/1.0", host: "nightshade.openmicroscopy.org", referrer: "https://nightshade.openmicroscopy.org/static/3rdparty/jquery-ui-1.12.1/css/jquery-ui.css"
```

Should these images have been included as part of https://github.com/ome/omero-web/pull/180/?

While trying to compare the bundle currently under `omeroweb/webgateway/static/3rdparty` with upstream, I also failed to reconcile the structure and the checksums of the `jquery-ui-1.12.1` folder

```
(base) sbesson@ls30630:3rdparty (master) $ find jquery-ui-1.12.1 -type f -exec shasum {} \;
9a32f41cc986942b283c75d887406480171d2d6b  jquery-ui-1.12.1/css/jquery-ui.css
86290501fa4a18b5149fe3edcc1829463e285bbe  jquery-ui-1.12.1/js/jquery-ui.js
```

with the bundle downloaded from https://jqueryui.com/download

```
(base) sbesson@ls30630:3rdparty (master) $ find /Users/sbesson/Downloads/jquery-ui-1.12.1.custom -type f -exec  shasum {} \;
1cb9e6e054cfa4bd1b5b06bfd8f84ad7dbb4ed05  /Users/sbesson/Downloads/jquery-ui-1.12.1.custom/jquery-ui.min.js
bf2aaba326e99a3941da8e00ea69a45b7ba7cf19  /Users/sbesson/Downloads/jquery-ui-1.12.1.custom/index.html
09a862bb4b51fcc8abafe0c74471b4a2b693f894  /Users/sbesson/Downloads/jquery-ui-1.12.1.custom/jquery-ui.min.css
ec1eba1d69ccc8335d62433f18640db1f14c12f5  /Users/sbesson/Downloads/jquery-ui-1.12.1.custom/jquery-ui.theme.min.css
[...]

PR #356, opened 2022-02-11
Update to Django 3.2.x

chris-allan:
## Introduction

This PR is a follow on from the Django 2.2.x upgrade WIP #200. The great initial work on that PR is unfortunately insufficient to bring us up to scratch with Django LTS versions; Django 2.2.x leaves LTS in just a few months in April 2022 [^1]. Django 3.2.x will remain in LTS until April 2024 [^2]:.

Furthermore, the strategy followed in this PR is slightly different: make the minimum number of changes required to maintain OMERO.web functionality and leave potential Django infrastructure modernisation to other follow up PRs.

Release notes for the major versions between our current 1.11.x and the 3.2.x target:

* https://docs.djangoproject.com/en/4.0/releases/2.0/
* https://docs.djangoproject.com/en/4.0/releases/2.1/
* https://docs.djangoproject.com/en/4.0/releases/2.2/
* https://docs.djangoproject.com/en/4.0/releases/3.0/
* https://docs.djangoproject.com/en/4.0/releases/3.1/
* https://docs.djangoproject.com/en/4.0/releases/3.2/

The `django-pipeline` dependency has been updated to nearly the latest version in order to support Django 3.2.x. All other dependencies have been left alone.

__NOTE:__ The supported Python versions for Django 3.2.x are currently 3.6, 3.7, 3.8, 3.9, and 3.10. This PR does not currently contain the requisite changes for Python 3.10 support.

## Proposed path to release

1. Decide on the new target version; I've selected 6.0.0 for the purposes of discussion
2. Release new versions of several OMERO.web plugins/applications with version pinning of `omero-web` to `<6.0.0`
3. Publicly announce that `omero-web` 6.0.0 __will__ include an upgrade to Django 3.2.x and publish the ideal timeline
4. Make whatever changes to this PR are required in order to have a test candidate; testing will need to be __extensive__ and __thorough__ as there's no telling what subtle bugs may be introduced
5. Deploy the test candidate and iterate until we are comfortable enough to publish a release candidate for community consumption
6. Continue to iterate, publishing new release candidates if required, until we are comfortable enough to make a release; this may include being comfortable with a subset of Django 3.2.x compliant OMERO.web plugins/applications
7. Make new releases of compliant OMERO.web plugins with version pinning of `omero-web` to `>=6.0.0`

### Django 2.0.x / 2.1.x

* `django.core.urlresolvers` removal (deprecated since 1.10) [^3]:
[...]

PR #433, opened 2023-01-05
Upgraded omero-web dependencies

kyleBalisWest:
Upgraded omero-web jquery dependencies. See [upgrade spreadsheet](https://docs.google.com/spreadsheets/d/1t4OS7DEjTtjPJVOZZkRlbtZtB_5F-krGILroVKD5pRc/edit?usp=sharing) for a detailed list of changes.

| Library               | Current Version | Latest Version | Link                                                                                                                               | Notes                                                                                                                                                                                             | Updated |
| --------------------- | --------------- | -------------- | ---------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------- |
| jquery-ui             | 1.12.1          | 1.13.2         | [https://jqueryui.com/download/](https://jqueryui.com/download/)                                                                   | Removed background and border from .ui-state-default<br>Removed background and color from .ui-widget-content<br>Removed border from .ui-state-active | Yes     |
| jquery                | 3.5.1           | 3.6.2          | [https://jquery.com/download/](https://jquery.com/download/)                                                                       | No changes needed because jQuery.htmlPrefilter isn't used                                                                                                                                         | Yes     |
| d3                    | 3.5.17          | 7.7.0          | [https://d3js.org/](https://d3js.org/)                                                                                             | No longer d3.scale.linear, now is d3.scaleLinear<br>No longer d3.svg.area, now is d3.area<br>No longer d3.svg.line, now is d3.line                                                                | Yes     |
| farbtastic            | 1.2             | 1.2            | [https://acko.net/blog/farbtastic-jquery-color-picker-plug-in/](https://acko.net/blog/farbtastic-jquery-color-picker-plug-in/)     |                                                                                                                                                                                                   | Yes     |
| hammer                | 2.0.2           | 2.0.8          | [https://hammerjs.github.io/](https://hammerjs.github.io/)                                                                         |                                                                                                                                                                                                   | Yes     |
| jquery.chosen         | 1.8.7           | 1.8.7          | [https://harvesthq.github.io/chosen/](https://harvesthq.github.io/chosen/)                                                         |                                                                                                                                                                                                   | Yes     |
| jquery.flot           | 0.8.3           | 0.8.3          | [https://www.flotcharts.org/](https://www.flotcharts.org/)                                                                         | Weirdness with npm version and github release version<br>Sticking to github release for now                                                                                                       | Yes     |
| jquery.jstree         | 3.3.10          | 3.3.12         | [https://www.jstree.com/](https://www.jstree.com/)                                                                                 |                                                                                                                                                                                                   | Yes     |
| jquery.tablesorter    | 2.0.3           | 2.31.3         | [https://mottie.github.io/tablesorter/docs/](https://mottie.github.io/tablesorter/docs/)                                           |                                                                                                                                                                                                   | No      |
| JQuerySpinBtn         | 1.3a            | ?              |                                                                                                                                    |                                                                                                                                                                                                   | No      |
| panojs                | 2.0.0           | 3.0.0          | [https://dimin.net/software/panojs/](https://dimin.net/software/panojs/)                                                           | Staying at 2.0.0 to allow for irregular tile sizes  | No     |
| raphael               | 2.1.0           | 2.3.0          | [https://dmitrybaranovskiy.github.io/raphael/](https://dmitrybaranovskiy.github.io/raphael/)                                       |                                                                                                                                                                                                   | Yes     |
| underscore            | 1.13.1          | 1.13.6         | [https://underscorejs.org/](https://underscorejs.org/)                                                                             |                                                                                                                                                                                                   | Yes     |
| aop                   | 1.3             | ?              |                                                                                                                                    |                                                                                                                                                                                                   | No      |
| jquery.blockUI        | 2.66.0          | 2.7            | [http://malsup.com/jquery/block/](http://malsup.com/jquery/block/)                                                                 |                                                                                                                                                                                                   | No      |
| jquery.form           | 4.3.0           | 4.3.0          | [https://github.com/jquery-form/form](https://malsup.com/jquery/form/)                                                             |                                                                                                                                                                                                   | No      |
| jquery.hotkeys        | 0.8             | ?              |                                                                                                                                    |                                                                                                                                                                                                   | No      |
| jquery.mousewheel     | 3.0.6           | 3.1.12         |                                                                                                                                    |                                                                                                                                                                                                   | Yes     |
| jquery.quicksearch    | 1.0             | ?              |                                                                                                                                    |                                                                                                                                                                                                   | No      |
| jquery.selectboxes    | 2.2.6           | ?              | [https://github.com/SamWM/jQuery-Plugins/tree/master/selectboxes](https://github.com/SamWM/jQuery-Plugins/tree/master/selectboxes) |                                                                                                                                                                                                   | No      |
| jquery.ui.touch-punch | 0.2.3           | ?              | [https://github.com/furf/jquery-ui-touch-punch](https://github.com/furf/jquery-ui-touch-punch)                                     |                                                                                                                                                                                                   | No      |


omeroweb/webgateway/static/3rdparty/jquery-ui-1.13.2/css/jquery-ui.css
```diff
@@ -1,13 +1,13 @@
-/* 
-	NB: manually edited to preserve webclient UI as it looked with jquery-ui 1.10.4
-	See https://github.com/ome/omero-web/pull/180
-*/
-/*! jQuery UI - v1.12.1 - 2016-09-14
+/*! jQuery UI - v1.13.2 - 2022-12-06
```
sbesson, line 1: As we are upgrading jquery-ui, 2 questions:

- can we use the opportunity to fix https://github.com/ome/omero-web/issues/333 which causes periodic errors in the nginx logs due to the lack of images
- as suggested in the comment above, the previous version included manual edits - see https://github.com/ome/omero-web/issues/333#issuecomment-969174368 for the relevant comments. Do we have an authoritative list of the places that were affected by these custom changes or would we expect this to be picked up as part of the OMERO.web testing?

omeroweb/settings.py
```diff
@@ -1585,20 +1585,20 @@ def report_settings(module):
                 "3rdparty/jquery-ui-1.10.4/js/jquery-ui.1.10.4.js",
```
sbesson, line 1585: Same comment as above

omeroweb/webgateway/static/3rdparty/jquery-ui-1.13.2/css/jquery-ui.css
kyleBalisWest: @sbesson I'm not sure about the first question but for the second one I forgot to write what was effected by the custom changes but I did put what the custom changes I made. Mostly they had to do with the tabs and background of the right hand panel if I remember correctly.

sbesson: For the first comment, I think the problem is that the jquery-ui CSS links to images which were not included in the former library upgrade e.g. in https://github.com/ome/omero-web/blob/v5.16.0/omeroweb/webgateway/static/3rdparty/jquery-ui-1.12.1/css/jquery-ui.css#L1094. This in turns leads to errors when these images are being requested.

Is this still the case for `jquery-ui 1.13.*` and if so, should the `images` be reincluded like panojs?

kyleBalisWest: Okay, I think I see what you mean. I added in the missing images to fix that problem.

omeroweb/settings.py
```diff
@@ -1570,7 +1570,7 @@ def report_settings(module):
                 "3rdparty/JQuerySpinBtn-1.3a/JQuerySpinBtn.css",
                 "3rdparty/jquery-ui-1.10.4/themes/base/jquery-ui.all.css",
```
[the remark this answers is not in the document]
kyleBalisWest: Yeah that is interesting. I tried fixing it to see if jquery-ui is loaded to the browser differently since it might be compressed with pipeline but I didn't see a difference.

chris-allan:
Not important right now while we're iterating but I would suggest that before this is merged we get all detail out of the Google Spreadsheet as Markdown and into the description of this PR. We will need the content in that form for a migration guide and/or changelog anyway.

omeroweb/webgateway/static/3rdparty/panojs-3.0.0/PanoJS.js
```diff
@@ -76,11 +76,8 @@ function PanoJS(viewer, options) {
                                                        options.tileExtension ? options.tileExtension : PanoJS.TILE_EXTENSION
                                                      );
 
-  this.xTileSize = (options.xTileSize ? options.xTileSize : PanoJS.TILE_SIZE);
-  this.yTileSize = (options.yTileSize ? options.yTileSize : PanoJS.TILE_SIZE);
-  this.realXTileSize = this.xTileSize;
-  this.realYTileSize = this.yTileSize;
-  
+  this.tileSize = (options.tileSize ? options.tileSize : PanoJS.TILE_SIZE);
```
will-moore, line 79: It looks like `PanoJs(viewer, options)` no-longer supports `options.xTileSize` and `options.yTileSize`?
So in our usage of `PanoJS` we'll need to use `options.tileSize` instead, which should be fine for when tile size x and y is the same, but I think there are cases when this isn't true. I don't have a feel for how often this occurs or if there's any way to support it (except maybe not upgrading)?

kyleBalisWest: Oh interesting, that's a good point. I don't know much about pyramidal data, so I was wondering how often does the xTileSize and yTileSize differ?

Though, if you don't want to upgrade it should be an easy switch back.

will-moore:
I'm afraid jQuery-UI auto-complete is still off (mapr):

Screenshot 2023-01-19 at 10 25 49

It looks like this is due to the previous jQuery-ui having a `display: block` on the `<a>`

Screenshot 2023-01-19 at 10 40 56

Whereas there is no such rule in the updated version:
Screenshot 2023-01-19 at 10 43 56

Probably easiest fix is to add `display: block` to that `dusty.css` under `.ui-menu-item a`?

jburel:
* Preview tab (right-hand panel)
   * when moving the right-hand side of the slider knob to the left, the knob always moves back to the full range after release

will-moore:
@jburel I think that sounds like expected behaviour if the right slider value is above the max intensity, the slider will update it's upper limit to be the max of the current slider value or the max intensity value.

omeroweb/webgateway/static/3rdparty/panojs-3.0.0/PanoJS.js
pwalczysko: @kyleBalisWest Often enough, see example in https://github.com/ome/omero-web/pull/433#issuecomment-1396786819 please :)

jburel:
> @jburel I think that sounds like expected behaviour if the right slider value is above the max intensity, the slider will update it's upper limit to be the max of the current slider value or the max intensity value.

The problem was noticed when using https://merge-ci.openmicroscopy.org/web/webclient/?show=image-236165

will-moore:
@pwalczysko For me, the old viewer with that image looks a bit different - Maybe you've got some tiles cached? Anyway, it's also due to Pano.js: In this case the tiles are square: ```"tile_size": {
"width": 1024,
"height": 1024
},```, so it could be fixed by passing Pano.js the correct (new) options, but I think we need to roll back the Pano.js upgrade anyway, to support non-square tiles of other images.

pwalczysko:
> @pwalczysko For me, the old viewer with that image looks a bit different - Maybe you've got some tiles cached? Anyway, it's also due to Pano.js: In this case the tiles are square: `"tile_size": { "width": 1024, "height": 1024 },`, so it could be fixed by passing Pano.js the correct (new) options, but I think we need to roll back the Pano.js upgrade anyway, to support non-square tiles of other images.

No, I tried to clear cache and also a new browser, the problem persists. The issue here might be that I have zoomed in a little, otherwise the problem does not manifest.

pwalczysko:
There are still some issues with icons on the Preview pane, see for example images such as https://merge-ci.openmicroscopy.org/web/webclient/?show=image-146365 

The icons on merge-ci

Screenshot 2023-01-23 at 12 08 47


The icons on lates-ci

Screenshot 2023-01-23 at 12 09 47


And errors in the Console

Screenshot 2023-01-23 at 12 11 32

will-moore:
@kyleBalisWest I think the zoom icons are missing (see Petr's screenshot) due to this diff in the reverting of Pano.js (last commit):

```
--- a/omeroweb/webgateway/static/3rdparty/panojs-3.0.0/controls.js
+++ b/omeroweb/webgateway/static/3rdparty/panojs-2.0.0/controls.js
@@ -10,15 +10,15 @@
 
 *******************************************************************************/
 
-PanoJS.CONTROL_IMAGE_ZOOMIN   = "images/32px_plus.png";
-PanoJS.CONTROL_IMAGE_ZOOM11   = "images/32px_11.png";
-PanoJS.CONTROL_IMAGE_ZOOMOUT  = "images/32px_minus.png";
-PanoJS.CONTROL_IMAGE_MAXIMIZE = "images/32px_show.png";
+PanoJS.CONTROL_IMAGE_ZOOMIN   = "32px_plus.png";
+PanoJS.CONTROL_IMAGE_ZOOM11   = "32px_11.png";
+PanoJS.CONTROL_IMAGE_ZOOMOUT  = "32px_minus.png";
+PanoJS.CONTROL_IMAGE_MAXIMIZE = "32px_show.png";
 
-PanoJS.CONTROL_IMAGE_ZOOMIN_TOUCH   = "images/64px_plus.png";
-PanoJS.CONTROL_IMAGE_ZOOM11_TOUCH   = "images/64px_11.png";
-PanoJS.CONTROL_IMAGE_ZOOMOUT_TOUCH  = "images/64px_minus.png";
-PanoJS.CONTROL_IMAGE_MAXIMIZE_TOUCH = "images/64px_show.png";
+PanoJS.CONTROL_IMAGE_ZOOMIN_TOUCH   = "64px_plus.png";
+PanoJS.CONTROL_IMAGE_ZOOM11_TOUCH   = "64px_11.png";
+PanoJS.CONTROL_IMAGE_ZOOMOUT_TOUCH  = "64px_minus.png";
+PanoJS.CONTROL_IMAGE_MAXIMIZE_TOUCH = "64px_show.png";
```

If I add the `images/` into panojs-2.0.0/controls.js in your branch, that fixes the issue and the icons load OK.
But I don't really understand why, since in the `master` branch, they don't exist there (you have reverted correctly)!
Maybe something else got changed and not reverted...?

omeroweb/webgateway/static/webgateway/js/ome.viewportImage.js
```diff
@@ -552,7 +552,7 @@ jQuery.fn.viewportImage = function(options) {
                 imageWidth      : myPyramid.width,
                 imageHeight     : myPyramid.height,
                 initialZoom     : init_zoom,
-                staticBaseURL   : mediaroot+'3rdparty/panojs-2.0.0/images/',
+                staticBaseURL   : mediaroot+'3rdparty/panojs-2.0.0/',
```
will-moore, line 555: I tracked it down to this change that didn't get reverted when you returned to `panojs-2.0.0` and that is breaking the icons.

kyleBalisWest: Oh yeah I missed that, thanks for finding it @will-moore. It should be fixed now!

pwalczysko (approved):

will-moore (changes requested):
All looks good - Just the auto-complete issues (and suggested fix) at https://github.com/ome/omero-web/pull/433#issuecomment-1396774881 to address

will-moore:
@sbesson That would mean that the web upgrade is a breaking change and you wouldn't want to upgrade omero-web without also upgrading mapr. But how to enforce that.
It adds complexity unnecessarily I think.
Just add `display: block` to `.ui-menu-item a` in dusty.css should fix it (going on my comment above)

chris-allan:
> That would mean that the web upgrade is a breaking change and you wouldn't want to upgrade omero-web without also upgrading mapr.

We are upgrading several core libraries in this PR and #435 is also likely to go in to the next version. These are both breaking changes in my opinion. Do you disagree?

Consequently, yes, I would expect you would not want to upgrade omero-web without also upgrading mapr, no? And you would want this new mapr version to depend on the at least the new minor version. This is the same pattern we followed for the Django 3.2 work in #356. Am I missing something here?

will-moore:
I guess I don't understand the objection to fixing the issue in this PR?

sbesson:
>  I guess I don't understand the objection to fixing the issue in this PR?

Having looked a bit the history, I found the root of this issue precedes this PR as turned out to also be a problem during the former jquery upgrade in #180. At the time, patches were introduced directly in the CSS to meet the expectations of the mapr plugin, more specifically in https://github.com/ome/omero-web/blob/efc042b7c836a28713835e1e0bc724cb8e40a2fe/omeroweb/webgateway/static/3rdparty/jquery-ui-1.12.1/css/jquery-ui.css#L133-L141 and https://github.com/ome/omero-web/blob/efc042b7c836a28713835e1e0bc724cb8e40a2fe/omeroweb/webgateway/static/3rdparty/jquery-ui-1.12.1/css/jquery-ui.css#L159-L163. In other terms, we have been maintaining a vendored fork of jquery-ui and more than the nature of the fix itself, my concern has to do with perpetuating this practice.

Let's assume another Web app developed by the community was relying on the internals of `d3` and using `d3.scale.linear` which has been dropped in the upgraded version. Would we consider introducing some form of compatibility layer in OMERO.web allowing them to continue using this API in the foreseeable future even as we keep upgrading the library?  In that case, I think we would all agree the maintenance cost is really high and the plugin should be responsible for upgrading its internals and making another release compatible with the new OMERO.web release. As far as I understand, the situation is no different for mapr which effectively relies on some jquery-ui styling which has been change upstream almost a decade ago.

Beyond this particular fix, paraphrasing @chris-allan, I also think we want to be transparent about the fact there are breaking changes associated with the library upgrades proposed here. I would be very careful about conveying the misleading idea that consumers/app developers can just upgrade OMERO.web without thinking about the nature of these changes and the implications on their own plugins. In many ways, having a handful of OMERO.web apps in our own ecosystem that also need to be updated and released is a healthy process as it forces us to think how to communicate the upgrade process to our community.

will-moore:
OK, I guess this PR is good to merge then

knabar:
Thanks everyone for the PR reviews!

sbesson (approved):

Issue #440, opened 2023-01-27
Consistent Behavior for Rendering Settings Query Parameters


kkoz:
This issue is to document some unexpected/inconsistent/undocumented behavior for rendering settings query parameters. 

- The tile parameter should require 5 values. If 3 or 4 are provided, and the second and third values are 0, the whole image will be returned. If they are not zero, the server experiences the following error: `serverStackTrace = ome.conditions.InternalException:  Wrapped Exception: (java.lang.RuntimeException):
Invalid Region, the width must be positive:0`
And the client gets a 404. If 2 or fewer values are provided, the client gets a 400 with the message “malformed tile argument, tile=0,0”. We suggest that the client get the 400 error with the message anytime fewer than 5 values are supplied.
- tile/region parameter are not specified at all in https://omero.readthedocs.io/en/stable/developers/Web/WebGateway.html. We should provide documentation specifying that x,y,w,h for region and resolution,tileX,tileY,tileWidth,tileHeight must be specified for region and tile respectively. Also should specify behavior when x,y lies outside the image bounds. We suggest returning a 400 with an error message if x,y is out of bounds and returning a cropped image if x,y is in bounds but x+w, y+h is out of bounds.
- If a window or color is not provided for a channel in “c”, the user’s rendering settings will be applied for that window. This is true whether other windows or colors are specified. This seems like it could be both confusing and buggy. The choice to use saved settings  might be unexpected behavior (the client won't really know what settings are being applied), and there may be issues with caching based on URLs.
- If the “m” query parameter is not supplied, it will default to the user’s saved rendering settings. The user’s saved settings will also be used if the “m” value supplied is not recognized by omero-web.
We suggest that m always be provided and that it be required to be a valid value otherwise a 400 will be returned with an informative error message

sbesson:
_Originally posted by @sbesson in https://github.com/ome/omero-web/pull/441#discussion_r1094287305_

As we review rendering settings query parameters across OMERO.web, we should be able to clarify the expectations for the API associated with saving rendering settings. As for the various render endpoints, a lot of the logic happens in `_get_prepared_image` with the following sequence of calls:
- windows, colors and inversions are applied to the channels
- the default Z/T are applied to the image
- img.saveDefaults() captures the rendering settings in the database

As mentioned in the discussion linked above, the current logic is confusing and potentially error-porne.
[...]

PR #443, opened 2023-02-03
Rdef query checks

kkoz:
Add a decorator `validate_rdef_query` which checks the validity of the query parameters supplied to a decorated endpoint. Decorated endpoints will be require the presence and (to a degree) correctness of rendering def-related query parameters or the client will receive a 400 error with a description of how to fix the request. Tries to manage some of the issues raised in #440. For backwards compatibility, two new endpoints `/webgateway/render_image_rdef` and `/webgateway/render_image_region_rdef` are created and the new decorator applied to them. Old endpoints have no behavior change.

### DIFF
--- a/omeroweb/webgateway/urls.py
+++ b/omeroweb/webgateway/urls.py
@@ -110,7 +110,7 @@ Params in render_image/<iid>/<z>/<t>/ are:
 render_image_rdef = url(
     r"^render_image_rdef/(?P<iid>[0-9]+)/(?:(?P<z>[0-9]+)/)?(?:(?P<t>[0-9]+)/)?$",
     views.render_image_rdef,
-    name="webgateway_render_image",
+    name="webgateway_render_image_rdef",
 )
 """
 Returns a jpeg of the OMERO image. See L{views.render_image}. Rendering
@@ -125,7 +125,7 @@ Params in render_image/<iid>/<z>/<t>/ are:
 render_image_region_rdef = url(
     r"^render_image_region_rdef/(?P<iid>[0-9]+)/(?P<z>[0-9]+)/(?P<t>[0-9]+)/$",
     views.render_image_region_rdef,
-    name="webgateway_render_image_region",
+    name="webgateway_render_image_region_rdef",
 )
 """
 Returns a jpeg of the OMERO image, rendering only a region specified in query
--- a/omeroweb/webgateway/urls.py
+++ b/omeroweb/webgateway/urls.py
@@ -107,6 +107,38 @@ Params in render_image/<iid>/<z>/<t>/ are:
     - t:    T index
 """
 
+render_image_rdef = url(
+    r"^render_image_rdef/(?P<iid>[0-9]+)/(?:(?P<z>[0-9]+)/)?(?:(?P<t>[0-9]+)/)?$",
+    views.render_image_rdef,
+    name="webgateway_render_image",
+)
+"""
+Returns a jpeg of the OMERO image. See L{views.render_image}. Rendering
+settings MUST be specified in the request parameters. See
+L{views.getImgDetailsFromReq} for details.
+Params in render_image/<iid>/<z>/<t>/ are:
+    - iid:  Image ID
+    - z:    Z index
+    - t:    T index
+"""
+
+render_image_region_rdef = url(
+    r"^render_image_region_rdef/(?P<iid>[0-9]+)/(?P<z>[0-9]+)/(?P<t>[0-9]+)/$",
+    views.render_image_region_rdef,
+    name="webgateway_render_image_region",
+)
+"""
+Returns a jpeg of the OMERO image, rendering only a region specified in query
+string as region=x,y,width,height. E.g. region=0,512,256,256 See
+L{views.render_image_region}.
+Rendering settings MUST be specified in the request parameters.
+Params in render_image/<iid>/<z>/<t>/ are:
+    - iid:  Image ID
+    - z:    Z index
+    - t:    T index
+"""
+
+
 render_split_channel = url(
     r"^render_split_channel/(?P<iid>[0-9]+)/(?P<z>[0-9]+)/(?P<t>[0-9]+)/$",
     views.render_split_channel,
@@ -568,6 +600,8 @@ urlpatterns = [
     webgateway,
     render_image,
     render_image_region,
+    render_image_rdef,
+    render_image_region_rdef,
     render_split_channel,
     render_row_plot,
     render_col_plot,
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -236,7 +236,7 @@ def validate_rdef_query(func):
 
         if "c" not in r:
             return HttpResponseBadRequest(
-                "Rendering settings must specify " + " channels as c"
+                "Rendering settings must specify channels as c"
             )
         channels, windows, colors = _split_channel_info(r["c"])
         # Need the same number of channels, windows, and colors
@@ -246,15 +246,15 @@ def validate_rdef_query(func):
             # Validation requires windows to be specified
             if window[0] is None or window[1] is None:
                 return HttpResponseBadRequest(
-                    "Must specify window for" + " each channel"
+                    "Must specify window for each channel"
                 )
             if colors[i] is None:
                 return HttpResponseBadRequest(
-                    "Must specify color for" + " each channel"
+                    "Must specify color for each channel"
                 )
         if "m" not in r or r["m"] not in ["g", "c"]:
             return HttpResponseBadRequest(
-                'Query parameter "m" must be' + ' present with value either "g" or "c"'
+                'Query parameter "m" must be present with value either "g" or "c"'
             )
         # TODO: What to do about z, t, and p?
         if "maps" in r:
@@ -269,7 +269,7 @@ def validate_rdef_query(func):
             rchannels = r["c"].split(",")
             if len(map_json) != len(rchannels):
                 return HttpResponseBadRequest(
-                    'Number of "maps" must' + " match number of channels"
+                    'Number of "maps" must match number of channels'
                 )
         return func(request, *args, **kwargs)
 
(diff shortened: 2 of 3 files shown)


omeroweb/webgateway/views.py
```diff
@@ -224,6 +224,58 @@ def getFirstName(self):
...
+            )
+        channels, windows, colors = _split_channel_info(r["c"])
+        # Need the same number of channels, windows, and colors
+        for i in range(0, len(channels)):
+            window = windows[i]
+            # Unspecified windows and colors are returned as None
+            # Validation requires windows to be specified
+            if window[0] is None or window[1] is None:
+                return HttpResponseBadRequest(
+                    "Must specify window for" + " each channel"
```
will-moore, line 249: This string is split in 2 (same below)? Maybe due to auto-formatting, but looks a bit strange

omeroweb/webgateway/views.py
```diff
@@ -1051,6 +1133,7 @@ def render_image_region(request, iid, z, t, conn=None, **kwargs):
 
 
 @login_required()
+@validate_rdef_query
```
will-moore, line 1136: I think we need to decide if we want to break the usage of NO rendering settings. E.g. https://idr.openmicroscopy.org/webclient/render_image/179703/ uses saved settings. Do we want that to return an error? I don't think that we use that in the webclient, but I don't know about 3rd-party apps

omeroweb/webgateway/views.py
kkoz: Yeah that was an auto-formatting thing - thanks.

omeroweb/webgateway/views.py
kkoz: I think that we should probably not allow users to make requests with no rendering settings specified. First, we will have issues with caching where the user may be getting inaccurate data. If a user loads an image with no settings specified, then updates the settings and reloads the image, their browser may just give them the cached image they loaded earlier rather than the updated one. Also, as the number of channels in images increases, it becomes increasingly unclear what the user is getting back if they don't specify anything. With 3 channels you might be able to figure it out by looking, but with 100 it's going to be very difficult.

kkoz:
Discussed the handling of endpoints without rendering query parameters. In order to preserve the functionality while preventing the caching issue, we are going to have endpoints where query parameters are not supplied get the current rendering settings, then redirect to the same endpoint but with rendering query parameters fully supplied. This way, the user will not experience caching issues, and the url will describe what rendering settings the user is seeing.

sbesson:
Assuming it's not too cumbersome, the redirection layer feels like an interesting approach that preserves some backwards-compatibility while internally construct query parameters inline with our recommendations.

As we work through this use case, this might be an opportunity to review https://omero.readthedocs.io/en/stable/developers/Web/WebGateway.html#rendering-settings and define more specifically what "current rendering settings" means. In particular, in a multi-user scenario where both the viewer is not the image owner and has her/his own rendering settings, which ones are expected to be considered as current?

will-moore:
I remembered that I am actually using the `/render_image/ID/` url (without rendering parameters) in https://github.com/will-moore/parade-crossfilter/ so I checked how this PR affects performance, since that app tends to load a lot of images...

Unfortunately, with this PR, every redirect adds quite a bit of time to the loading of images (this is testing from home on dev web-server so times are slower anyway, but the effect is the same):

without this PR:

Screenshot 2023-02-15 at 07 24 43

with this PR:

Screenshot 2023-02-15 at 07 21 02

So it still feels like this is a significantly worse experience for users of that functionality,
and we should still support this functionality if users want it (and are happy to accept any potential caching issues)

will-moore:
There's a bunch of failing tests that are likely due to this PR at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/1367/

sbesson:
At a glance, most of the integration tests are failing because the expectation was to receive a 200 HTTP error code previously instead of 302 right? 

I am a bit more concerned by the performance overhead associated with the redirection layer. How easy is it to profile the code to identify where the extra time is spend?

will-moore:
I think the redirect is slower simply because it makes 2 calls to load the image instead of 1. In both calls the rendering engine has to be initialised to load the rendering settings.
The actual loading of the jpeg appears slower in my screenshots above, but I don't *think* that it's a real difference - more likely due to the variable bandwidth of home internet.

will-moore:
These checks are causing "Split View Figure" to fail, since it uses URLs like: `render_image/ID/?c=1`

Screenshot 2023-02-17 at 13 57 59

We could fix this by supplying the full rendering settings, but I think we should probably just remove the Split View Figure from the webclient. This really pre-dates OMERO.figure and now it's not really needed. cc @jburel

jburel:
@will-moore: removing it will make sense but note that we will need to have a similar change in insight
First step will be to  deprecate it

will-moore:
@jburel It's already broken! If we deprecate it, we should release a webclient with it working (but deprecated).
That would delay this PR.
Or we fix it - but that's not worth it.
Or we just replace the Split-View figure dialog with "Use OMERO.figure" message.

kkoz:
@will-moore @jburel I discussed with @knabar and we think the best way to move forward here is to create new endpoints for render_image and render_image_region. These would have the rdef query validation decorators while the old endpoints would be as they were but would be marked as deprecated. We would recommend that going forward people use the new endpoints. This way there won't be any redirection performance issues to worry about, and applications currently using the old endpoints will be unaffected. We would then migrate clients to the new endpoints. Would like your thoughts or objections.

will-moore:
@kkoz Is this simply to deal with the Split-View figure issue?
I would prefer not to have `render_image2` and deprecate `render_image` just for this, since that will create a lot of work to migrate everything. I don't think Split-View figure has been used much - it's very old and we've promoted OMERO.figure as the way to make figures for a long time.
If we really can't remove Split-View figure in the next release, then create `render_image_unsafe/ID/` which allows `?c=1` and mark that endpoint as deprecated.

kkoz:
@will-moore No this has nothing to do with the Split-View figure issue. Sorry I should have mentioned that. It is more in response to the points made in our discussion this morning, in particular to preserve the behavior of the endpoint with no query parameters provided. I would still propose that we update the docs to suggest always providing the query parameters, but existing clients which want to use the "default settings" endpoint will still be able to do so without experiencing any changes.

will-moore:
"while the old endpoints would be as they were but would be marked as deprecated" - do you mean you want to mark the current behaviour allowing `/render_image/ID/` as deprecated? I still don't think that we should remove this behaviour (for all the reasons we discussed). This would mean we don't need new endpoints and don't need to migrate apps (same for other users). Certainly agree to updating the docs to describe the behaviour of this endpoint with appropriate warnings.

will-moore:
I think if we follow the policy of "No feature removal without first deprecating" to apply to Split-View figure (as @jburel mentioned above), then the same should apply to the removal of support for `render_image/ID/?c=1` which is what's breaking the Split-View figure.

In which case we need a PR to deprecate both of those features and release that before removing them?

kkoz:
@will-moore I'd like to make sure I understand what you think the final behavior should be after all the deprecations and feature removal is complete. Let me know if the below is correct:

- We have `render_image` and `render_image_region` endpoints (no new endpoints)
- Both endpoints support either no rendering query parameters at all or a complete set (including windows and colors for all channels)

Therefore `render_image/ID` would be a valid endpoint but `render_image/ID/?c=1` would not.

will-moore:
Yes, that's exactly what I would want. 👍

knabar:
@will-moore @jburel 

After some additional discussion, I don’t think this last proposal on its own will do exactly what we’re looking for.

One of the main goals is for the client to be able to cache basically indefinitely for best possible performance, which is not possible if the endpoint response depends on anything other than the given parameters.

Extending the proposal to get there though should be simple and not impact any users of the current endpoints, by adding two new endpoints that wrap the current endpoints to add new behavior:
* Only support full set of rendering parameters
* Add caching headers to enable unlimited caching

Since the new endpoints would just wrap the current ones, there should be a minimal amount of new code needed, without any code duplication.

I don’t see any need to remove the old endpoints, and if there are advantages to deprecate the partial rendering parameters support, we can still do that, or just leave the endpoints as they are indefinitely.

kkoz:
I added the new endpoints to give us a concrete discussion point.

omeroweb/webgateway/urls.py
```diff
@@ -107,6 +107,38 @@
...
+Params in render_image/<iid>/<z>/<t>/ are:
+    - iid:  Image ID
+    - z:    Z index
+    - t:    T index
+"""
+
+render_image_region_rdef = url(
+    r"^render_image_region_rdef/(?P<iid>[0-9]+)/(?P<z>[0-9]+)/(?P<t>[0-9]+)/$",
+    views.render_image_region_rdef,
+    name="webgateway_render_image_region",
```
will-moore, line 128: These names need updating to be unique (duplicates of the old URLs)

will-moore:
Looks good.
Hopefully that should fix the failing integration tests too...

will-moore:
We're down to just 5 failing tests now:
https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/1375/testReport/
All are render_image_region, with `'tile': '0,0,0'` or `'0,0,0,512,512'` 
or `'region': '0,0,10,10'` or `'0,0,512,512'`.

omeroweb/webgateway/views.py
```diff
@@ -1109,6 +1167,18 @@ def render_image(request, iid, z=None, t=None, conn=None, **kwargs):
     return rsp
 
 
+@login_required()
+@validate_rdef_query
+def render_image_rdef(request, iid, z=None, t=None, conn=None, **kwargs):
+    return render_image(request, iid, **kwargs)
```
will-moore, line 1173: the `z` and `t` parameters are being ignored here.
I think it might also help to pass in the `conn` so that `render_image()` doesn't have to look it up again.

will-moore (approved):
Looks good 👍.
Just need to update the PR description to document the changes here. Thanks!

will-moore:
Description looks good, thanks. Good to merge for me!

PR #448, opened 2023-02-14
Bump minimum Django version to 3.2.18

### LINKED ISSUE #447: Dependabot alert - DoS vulnerability in Django < 3.2.17
See https://github.com/ome/omero-web/security/dependabot/2

In Django 3.2 before 3.2.17, 4.0 before 4.0.9, and 4.1 before 4.1.6, the parsed values of Accept-Language headers are cached in order to avoid repetitive parsing. This leads to a potential denial-of-service vector via excessive memory usage if the raw value of Accept-Language headers is very large.

sbesson:
Includes the security fixes from
https://www.djangoproject.com/weblog/2023/feb/14/security-releases/ and https://www.djangoproject.com/weblog/2023/feb/01/security-releases/

Fixes #447

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ setup(
         "omero-py>=5.7.0",
         # minimum requirements for `omero web start`
         "concurrent-log-handler>=0.9.20",
-        "Django>=3.2,<4.0",
+        "Django>=3.2.18,<4.0",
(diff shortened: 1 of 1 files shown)


knabar (approved):
Built and started locally with no apparent issues

PR #449, opened 2023-03-01
Add change log for 5.19.0

knabar:
Started work on the change log and upgrading info for 5.19.0.

I think we should mention the "deprecation" of part of the `render_image` call and would appreciate feedback on how to describe our recommendations (see https://github.com/ome/omero-web/compare/master...knabar:maint-changelog-5.19.0?expand=1#diff-f411a5009131db5a6c7242b49ca78488f2d274ef9e5e5f4de9262a42108a5386).

Similar text should go into the documentation as well, see issue https://github.com/ome/omero-documentation/issues/2299.

See https://github.com/ome/omero-documentation/pull/2301

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -4,6 +4,18 @@ This document highlights steps that may need to be taken by developers
 when upgrading OMERO.web to ensure plugins or other customizations
 continue to function as expected.
 
+## OMERO.web 5.19.0
+
+### Partial channel lists in render_image and render_image_region web API calls
+
+When requesting images via the render_image and render_image_region web API
+calls, the desired channel rendering settings may be omitted (to use the
+default rendering settings), fully specified (to use custom settings), or 
+partially specified (to combine custom settings with default rendering 
+settings).  The latter method of partially specifying channel rendering settings
+is now considered deprecated and should not be used in any new or updated
+applications and plugins.
+
 ## OMERO.web 5.18.0
 
 ### Connector storage in Django sessions
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -8,13 +8,20 @@ continue to function as expected.
 
 ### Partial channel lists in render_image and render_image_region web API calls
 
-When requesting images via the render_image and render_image_region web API
+When requesting images via the `render_image` and `render_image_region` web API
 calls, the desired channel rendering settings may be omitted (to use the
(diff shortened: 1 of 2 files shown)


will-moore:
Maybe a way to make the deprecation more visible would be for a separate PR that adds the deprecation/upgrade note you have here, as well as a mention in the doc-string for `render_image` etc. Then the changelog could include an item "Deprecated... [PR]"` linking to the PR, since I think some users may ONLY look at the change-log and could miss the upgrading note?

sbesson:
Thanks @knabar for starting the conversation. My understanding is that the current decision is not to deprecate the endpoints themselves but only some variants of the query parameters.

At the CHANGELOG level, I agree with Will that we should explicit list the deprecations alongside the new endpoints with a link to upgrading for the developers upgrading document

At the level of the API documentation, https://github.com/ome/omero-web/blob/9ac7ade1886ed0f5e359af62eee0fa1748fa5d68/omeroweb/webgateway/views.py#L1089-L1090 is the relevant block from the docstring and points itself to https://github.com/ome/omero-web/blob/9ac7ade1886ed0f5e359af62eee0fa1748fa5d68/omeroweb/webgateway/views.py#L278-L290. 
Reading both this information and this PR, is it correct the the main change is the format of the `c` parameter i.e. `[-]ID[|wndst:wndend][#HEXCOLOR][,...]` is deprecated and future releases will only support `[-]ID|wndst:wndend#HEXCOLOR[,...]` if specified? Are there additional changes in terms of parameter requirement and/or format? 
Should we work towards an updated version of this query parameter block that we can re-use in various places (docstring, omero-documentation)?

UPGRADING.md
```diff
@@ -8,14 +8,21 @@ continue to function as expected.
...
 settings).  The latter method of partially specifying channel rendering settings
 is now considered deprecated and should not be used in any new or updated
 applications and plugins.
 
+New API calls for requesting images using fully specified channel rendering
+settings are now available as `render_image_rdef` and 
+`render_image_region_rdef`.
+
+Documentation for these API calls is available at
+https://omero.readthedocs.io/en/latest/developers/Web/WebGateway.html.
```
jburel, line 24:
```suggestion
https://omero.readthedocs.io/en/stable/developers/Web/WebGateway.html.
```

will-moore (approved):
Looks good 👍

PR #455, opened 2023-04-10
use ome channel to install Ice 3.6

jburel:
Ice 3.7 is the only version available from conda-forge channel

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -34,7 +34,7 @@ other options, please see `OMERO.py`_.
 
 ::
 
-    conda create -n myenv -c conda-forge python=3.8 zeroc-ice omero-py
+    conda create -n myenv -c ome python=3.8 zeroc-ice36-python omero-py
     conda activate myenv
     pip install omero-web
 
--- a/README.rst
+++ b/README.rst
@@ -34,8 +34,9 @@ other options, please see `OMERO.py`_.
 
 ::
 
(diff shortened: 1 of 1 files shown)


dominikl:
👍 Looks good. That's what I had to do manually to install omero-py.

dominikl (approved):

README.rst
```diff
@@ -34,8 +34,9 @@ other options, please see `OMERO.py`_.
 
 ::
 
-    conda create -n myenv -c conda-forge python=3.8 zeroc-ice omero-py
+    conda create -n myenv -c ome python=3.8 zeroc-ice36-python
```
sbesson, line 37: Looking at this from an external perspective (Glencoe is not using Conda), the mixture of 2 different channels feels slightly confusing. `conda-forge` has a 3.6.5 version of the `zeroc-ice` package -see https://anaconda.org/conda-forge/zeroc-ice/files?version=3.6.5 but this is not the latest. Its`omero-py` recipe uses a version suffix to select the correct dependency - https://github.com/conda-forge/omero-py-feedstock/blob/8185680b1f7d5e00a69cc95bf530df3c67f00a12/recipe/meta.yaml#L32.

Switching back to the ome recipe is an option but this recipe (and the `omero-py` one) has been archived in favor of the `conda-forge` packages - see https://github.com/ome/conda-zeroc-ice36-python/issues/9. So this should likely be revisited

PR #456, opened 2023-04-13
Allow Tiled Image Histograms

kkoz:
Requires https://github.com/ome/omero-server/pull/162. Removes the check in omero-web which prevents histograms from being retrieved for tiled images. Also, currently if a histogram is requested with an invalid c, z, or t value, it results in a 500 error due to an `IndexOutOfBoundsException`. This PR checks those parameters and returns 404 if they are out of bounds.

### DIFF
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -2852,7 +2852,9 @@ def histogram_json(request, iid, theC, conn=None, **kwargs):
     except omero.ApiUsageException as ex:
         logger.warn(ex)
         resObj = {"error": ex.message}
-        return HttpResponseBadRequest(json.dumps(resObj), content_type="application/json")
+        return HttpResponseBadRequest(
+            json.dumps(resObj), content_type="application/json"
+        )
     return JsonResponse({"data": histogram})
 
 
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -2836,15 +2836,6 @@ def histogram_json(request, iid, theC, conn=None, **kwargs):
     image = conn.getObject("Image", iid)
     if image is None:
         raise Http404
-    maxW, maxH = conn.getMaxPlaneSize()
-    sizeX = image.getSizeX()
-    sizeY = image.getSizeY()
-    if (sizeX * sizeY) > (maxW * maxH):
-        msg = "Histogram not supported for 'big' images (over %s * %s pixels)" % (
-            maxW,
-            maxH,
-        )
-        return JsonResponse({"error": msg})
 
     theZ = int(request.GET.get("theZ", 0))
     theT = int(request.GET.get("theT", 0))
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -2842,6 +2842,11 @@ def histogram_json(request, iid, theC, conn=None, **kwargs):
     theC = int(theC)
     binCount = int(request.GET.get("bins", 256))
 
+    if (theZ >= image.getSizeZ() or
+        theT >= image.getSizeT() or
+        theC >= image.getSizeC()):
+        raise Http404
+
     # TODO: handle projection when supported by OMERO
     data = image.getHistogram([theC], binCount, theZ=theZ, theT=theT)
     histogram = data[theC]
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -2851,6 +2851,10 @@ def histogram_json(request, iid, theC, conn=None, **kwargs):
         histogram = data[theC]
     except omero.ApiUsageException as ex:
         logger.warn(ex)
+        if ex.message == "All resolution levels larger than max plane size":
+            return HttpResponseBadRequest(
+                "All resolution levels larger than max plane size"
+            )
         return HttpResponseBadRequest("Histograms not supported for tiled images")
     return JsonResponse({"data": histogram})
 
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -2851,11 +2851,8 @@ def histogram_json(request, iid, theC, conn=None, **kwargs):
         histogram = data[theC]
     except omero.ApiUsageException as ex:
         logger.warn(ex)
-        if ex.message == "All resolution levels larger than max plane size":
-            return HttpResponseBadRequest(
-                "All resolution levels larger than max plane size"
-            )
-        return HttpResponseBadRequest("Histograms not supported for tiled images")
+        resObj = {"error": ex.message}
+        return HttpResponseBadRequest(json.dumps(resObj), content_type="application/json")
     return JsonResponse({"data": histogram})
 
 
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -2842,9 +2842,7 @@ def histogram_json(request, iid, theC, conn=None, **kwargs):
     theC = int(theC)
     binCount = int(request.GET.get("bins", 256))
 
-    if (theZ >= image.getSizeZ() or
-        theT >= image.getSizeT() or
-        theC >= image.getSizeC()):
+    if theZ >= image.getSizeZ() or theT >= image.getSizeT() or theC >= image.getSizeC():
         raise Http404
 
     # TODO: handle projection when supported by OMERO
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -2846,9 +2846,11 @@ def histogram_json(request, iid, theC, conn=None, **kwargs):
         raise Http404
 
     # TODO: handle projection when supported by OMERO
-    data = image.getHistogram([theC], binCount, theZ=theZ, theT=theT)
-    histogram = data[theC]
-
+    try:
+        data = image.getHistogram([theC], binCount, theZ=theZ, theT=theT)
+        histogram = data[theC]
+    except omero.ApiUsageException as ex:
+        return HttpResponseBadRequest("Histograms not supported for tiled images")
     return JsonResponse({"data": histogram})
 
 
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -2850,6 +2850,7 @@ def histogram_json(request, iid, theC, conn=None, **kwargs):
         data = image.getHistogram([theC], binCount, theZ=theZ, theT=theT)
         histogram = data[theC]
     except omero.ApiUsageException as ex:
+        logger.warn(ex)
         return HttpResponseBadRequest("Histograms not supported for tiled images")
     return JsonResponse({"data": histogram})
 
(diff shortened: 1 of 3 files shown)


omeroweb/webgateway/views.py
```diff
@@ -2836,21 +2836,15 @@ def histogram_json(request, iid, theC, conn=None, **kwargs):
...
     theZ = int(request.GET.get("theZ", 0))
     theT = int(request.GET.get("theT", 0))
     theC = int(theC)
     binCount = int(request.GET.get("bins", 256))
 
+    if theZ >= image.getSizeZ() or theT >= image.getSizeT() or theC >= image.getSizeC():
+        raise Http404
+
     # TODO: handle projection when supported by OMERO
     data = image.getHistogram([theC], binCount, theZ=theZ, theT=theT)
```
sbesson, line 2849: As this feature depends on a combination of a client-side and a server-side change,  I think the implementation will need to be more defensive to handle different OMERO.web/OMERO.server combinations

In the scenario where a new OMERO.web (with this PR included) would try to retrieve the histogram of a pyramidal image from an old OMERO.server, should this API call catch the ApiUsageException thrown server-side and return an appropriate JSON reponse?

sbesson: Handled in [871fe81](https://github.com/ome/omero-web/pull/456/commits/871fe8193a693c1e2b8b0333faebdeab37fd8f6e)

will-moore:
So this was in the merge-ci build today https://github.com/snoopycrimecop/omero-web/commit/dfa0b0f9c5c29fc400451949e4ff85259069e5b5 but looking at a [big image on merge-ci ](https://merge-ci.openmicroscopy.org/web/webclient/?show=image-105665) I still see:

Screenshot 2023-04-20 at 09 39 43

will-moore:
Ah - apologies, I see that this PR doesn't change the webclient UI, just the JSON api response...
So I tried https://merge-ci.openmicroscopy.org/web/webgateway/histogram_json/105665/channel/0/?theT=0&theZ=0 and I still see `Histograms not supported for tiled images` response, so maybe [the](https://github.com/ome/omero-server/pull/162) isn't in the server?

will-moore:
Ah - yes: https://github.com/snoopycrimecop/omero-server/commit/dbb07d5f59278cfb0a1e24556f7c2dbd418ca315 says that https://github.com/ome/omero-server/pull/162 still has an exclude in the description.

sbesson:
Yes, additional commits are expected on the low-level omero-server PR which is still excluded for now. This PR will be required to relax the client-side checks and allow to test the new server functionality via OMERO.web.
However, we should definitely test that the proposed change maintains the current behavior with OMERO.server up to 5.6.6 i.e. histogram requests are functional for small images and fail with an informative error for pyramidal images

sbesson:
To @will-moore's comment in https://github.com/ome/omero-web/pull/456#issuecomment-1515947298, todays' CI deployment still had the `Show histogram` functionality disabled. Looking into the code, it looks like they are additional locations where the functionality is disabled based on a check, including

- https://github.com/ome/omero-web/blob/a6f89622e3de4e3a59356a1db127a89afb17bb99/omeroweb/webclient/templates/webclient/annotations/metadata_preview.html#L471C19-L477
- https://github.com/ome/omero-web/blob/a6f89622e3de4e3a59356a1db127a89afb17bb99/omeroweb/webgateway/templates/webgateway/viewport/omero_image.html#L1079-L1086
- for OMERO.iviewer https://github.com/ome/omero-iviewer/blob/3184ec460e5d0e15575b48952d5ee409ad71c3a4/src/settings/settings.html#L74-L77

Given the complexity of the testing, happy to focus the scope of this PR to the OMERO.web webgateway endpoint. Just raising this as the new functionality will require additional changes to be fully exposed in the UI.

omeroweb/webgateway/views.py
```diff
@@ -2836,25 +2836,26 @@ def histogram_json(request, iid, theC, conn=None, **kwargs):
...
+    try:
+        data = image.getHistogram([theC], binCount, theZ=theZ, theT=theT)
+        histogram = data[theC]
+    except omero.ApiUsageException as ex:
+        logger.warn(ex)
+        if ex.message == "All resolution levels larger than max plane size":
+            return HttpResponseBadRequest(
+                "All resolution levels larger than max plane size"
+            )
+        return HttpResponseBadRequest("Histograms not supported for tiled images")
```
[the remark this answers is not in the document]
sbesson: Coming back to this, since both the contract of the API has not changed and both old and new versions of the server components will throw an `ApiUsageException`, I would be inclined to remove the client-side checks and simply pass `ex.message` to the response?

For newer versions of OMERO.web running against older versions of OMERO.server (5.6.7 or earlier), this means the error message will be `This method can not handle tiled images yet.` instead of `{"error": "Histogram not supported for 'big' images (over 3192 * 3192 pixels)"}`. This is slightly less informative but arguably not significantly so.

sbesson (changes requested):
Without this PR:

- the `webgateway/histogram_json` endpoint would return a 200 reponse with a message `{"error": "Histogram not supported for 'big' images (over 3192 * 3192 pixels)"}`  when using an image larger than the maximum plane size independently of the resolution
- the `webgateway/histogram_json` endpoint would return a valid histogram for a image smaller than the maximum plane size
- the `webgateway/histogram_json` endpoint would return a 500 error code when specifying an invalid channel index or an invalid Z/T index as a query parameter

With this PR:
- specifying an invalid channel, Z or T index results in a 404 error code 
- using a version of OMERO.server with the server-side histogram support for tiled pyramidal images (5.6.8+)
   - the `webgateway/histogram_json` endpoint returns a 400 response with a response body (`All resolution levels larger than max plane size`) for large image which has either no resolution or only resolutions larger than the max plane size
   - the `webgateway/histogram_json` endpoint returns a 200 response with a valid histogram for either images smaller than the max plane size or large images with at least one resolution level smaller than the max plane size 
- using OMERO.server 5.6.7  without the server-side histogram support for tiled pyramidal images
   - the `webgateway/histogram_json` endpoint returns a 400 response with a response body `All resolution levels larger than max plane size` for all images larger than 3Kx3K
   - the endpoint returns a 200 response with the correct histogram data for small images
 
Overall,
- the server-side histogram support is correctly exposed in the OMERO.web endpoint 
- the error codes are improved and consistent
- the response for large images missing resolution levels is informative
- one inline suggestion to simplify the handling of `ApiUsageException` returned by the server and simply use the server message to the HTTP reponse.

The only other point possibly up for discussion is wether 400 is the best error code to set in the HTTP response when the server replies an `ApiUsageException`. We have had this discussion with @kkoz  and have not been able to find a HTTP error code that felt clearly more adapted. The current error code is consistent with the decisions made for the image-region micro-service https://github.com/glencoesoftware/omero-ms-image-region/blob/c7e3c1a43f89af86c402310cbf4b00ddfa6c0508/src/main/java/com/glencoesoftware/omero/ms/image/region/ImageRegionVerticle.java#L471-L475 but happy to hear suggestions of alternate 4xx or 5xx code.

knabar:
@sbesson 

> For newer versions of OMERO.web running against older versions of OMERO.server (5.6.7 or earlier), this means the error message will be `This method can not handle tiled images yet.` instead of `{"error": "Histogram not supported for 'big' images (over 3192 * 3192 pixels)"}`. 

To clarify, would this change the response from a JSON object to a string?  If so, that could potentially introduce runtime errors in clients.

sbesson:
> To clarify, would this change the response from a JSON object to a string? If so, that could potentially introduce runtime errors in clients.

With the current proposal, there are two changes in the response  for an image without pyramidal images (independently of whether the server can or cannot create such histograms):
- the status code is changed from 200 to 400
- the response is changed from a JSON with an `error` attribute to a string
Both of these changes have client implications as you suggest.

To some extent this is another possible response to the concerns raised in the last paragraph https://github.com/ome/omero-web/pull/456#pullrequestreview-1697057007 i.e. consider that OMERO.web should return a successful request and make it the consumer responsibility to check the JSON response for possible errors.

knabar:
Looking at the latest change I tried to find out if it's guaranteed that `ApiUsageException` has a `message` property that is a type (hopefully a string) handled by `json.dumps`, but it looks like it's one of the auto-generated Ice classes, so I just won't go there.  (To handle generic exceptions, people use something like https://stackoverflow.com/a/45532289/34171)

So, looks good to me :+1:

sbesson (approved):
Retested using a set of [synthetic samples](https://bio-formats.readthedocs.io/en/stable/developers/generating-test-images.html) generated  with different sizes and number of resolutions. The following tables summarize the the HTTP response to the `histogram_json` endpoint for all scenarios including the status code, content-type and response body

Without this PR and with a recent version of OMERO.server (5.6.8+)

| Image dimensions (XY) | Resolutions | Statut code | Content-Type     | Reponse body                                                                        |
|-----------------------|-------------|-------------|------------------|---------------------------------------------------------------------------------|
| 512x512               | 1           | 200         | application/json | {"data": ... }                                                                  |
| 10000x10000           | 2           | 200         | application/json | {"error": "Histogram not supported for 'big' images (over 3192 * 3192 pixels)"} |
| 10000x10000           | 10           | 200         | application/json | {"error": "Histogram not supported for 'big' images (over 3192 * 3192 pixels)"} |

Without this PR and with an older version of OMERO.server (5.6.7)

| Image dimensions (XY) | Resolutions | Statut code | Content-Type     | Reponse                                                                         |
|-----------------------|-------------|-------------|------------------|---------------------------------------------------------------------------------|
| 512x512               | 1           | 200         | application/json | {"data": ... }                                                                  |
| 10000x10000           | 2           | 200         | application/json | {"error": "Histogram not supported for 'big' images (over 3192 * 3192 pixels)"} |
| 10000x10000           | 10           | 200         | application/json | {"error": "Histogram not supported for 'big' images (over 3192 * 3192 pixels)"} |


With this PR and with a recent version of OMERO.server (5.6.8+)

| XY dimensions | Resolutions | Status code | Content-Type     | Reponse                                                       |
|---------------|-------------|-------------|------------------|---------------------------------------------------------------|
| 512x512       | 1           | 200         | application/json | {"data": ... }                                                |
| 10000x10000   | 2           | 400         | application/json | {"error": "All resolution levels larger than max plane size"} |
| 10000x10000   | 10          | 200         | application/json | {"data": ... }                                                |

With this PR and with an older version of OMERO.server (5.6.7)

| XY dimensions | Resolutions | Status code | Content-Type     | Reponse                                                   |
|---------------|-------------|-------------|------------------|-----------------------------------------------------------|
| 512x512       | 1           | 200         | application/json | {"data": ... }                                            |
| 10000x10000   | 2           | 400         | application/json | {"error": "This method can not handle tiled images yet."} |
| 10000x10000   | 10          | 400         | application/json | {"error": "This method can not handle tiled images yet."} |


This outcome is consistent with the expectations we defined yesterday with @knabar and @kkoz i.e.
- when the server is able to compute the histogram, OMERO.web should respond with the 200 status code and a JSON body containing the histogram under the `data` field
- when the server is unable to compute the histogram, either because it's too old or because the data is lacking pyramidal level, OMERO.web should respond with the 400 status code and a JSON body response containing an `error` field indicating the underlying server message
For downstream clients, the main implications is that the status code can now be tested to check whether the request was successful and the error message has slightly changed when using an old server.

PR #457, opened 2023-04-23
Declare portalocker as external dependency

sbesson:
Companion PR to https://github.com/ome/omero-py/pull/370

This should fix the nightly CI builds (see e.g. [log](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-web/1639/console)) by replacing the dependency on `omero_ext.portalocker` (which is currently proposed for removal in the PR above) by the upstream `portalocker` package.

PR #458, opened 2023-04-24
Compatibility with Django 4.0

chris-allan:
## Introduction

This PR is a follow on from the Django 3.2.x upgrade #356 and #435. Django 3.2.x leaves LTS in 12 months in April 2024 [^1] so it would be prudent to make ourselves ready for that by getting the codebase prepared to first support at least Django 4.0.x. Django 4.0.x went end of life in August 2022.

The strategy followed in this PR is to first allow Django 4.0.x and then make all the changes in a way that is also still compatible with Django 3.2. This should allow us to work with both versions and then revert the singular commit allowing Django 4.0 if desired and when we are ready.

Release notes for the major versions between our current 3.2.x and the 4.0.x target:

* https://docs.djangoproject.com/en/4.2/releases/4.0/

The `django-pipeline` dependency has been updated to the latest version in order to support Django 4.0.x. `pytz` support has been deprecated and the dependency removed from Django itself so it has been added. All other dependencies have been left alone.

__NOTE:__ The supported Python versions for Django 4.0.x are currently 3.8, 3.9, and 3.10. 3.11 is only supported from Django 4.1.x forward.

## Proposed path to release

1. Decide on the new target version; I've selected 6.0.0 for the purposes of discussion
2. Release new versions of several OMERO.web plugins/applications with version pinning of `omero-web` to `<6.0.0`
3. Publicly announce that `omero-web` 6.0.0 __will__ include support for Django 4.0.x (which results in all deprecations in 3.0.x and 3.1.x being dropped) and publish the ideal timeline
4. Deploy the test candidate and iterate until we are comfortable enough to publish a release candidate for community consumption
5. Continue to iterate, publishing new release candidates if required, until we are comfortable enough to make a release; this may include being comfortable with a subset of Django 4.0.x compliant OMERO.web plugins/applications
6. Make new releases of compliant OMERO.web plugins with version pinning of `omero-web` to `>=6.0.0`

### Django 4.0.x (deprecations from Django 3.0.x removed)

See:
* https://docs.djangoproject.com/en/4.2/releases/3.0/#deprecated-features-3-0

Detail:

* django.utils.translation.ugettext(), ugettext_lazy(), ugettext_noop(), ungettext(), and ungettext_lazy() are deprecated in favor of the functions that they’re aliases for: [django.utils.translation.gettext()](https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.translation.gettext), [gettext_lazy()](https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.translation.gettext_lazy), [gettext_noop()](https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.translation.gettext_noop), [ngettext()](https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.translation.ngettext), and [ngettext_lazy()](https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.translation.ngettext_lazy).
* To avoid possible confusion as to effective scope, the private internal utility is_safe_url() is renamed to url_has_allowed_host_and_scheme(). That a URL has an allowed host and scheme doesn’t in general imply that it’s “safe”. It may still be quoted incorrectly, for example. Ensure to also use [iri_to_uri()](https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.encoding.iri_to_uri) on the path component of untrusted URLs.

### Django 4.0.x (deprecations from Django 3.1.x removed)

See:
* https://docs.djangoproject.com/en/4.2/releases/3.1/#deprecated-features-3-1

Detail:
* The HttpRequest.is_ajax() method is deprecated as it relied on a jQuery-specific way of signifying AJAX calls, while current usage tends to use the JavaScript [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).
* django.conf.urls.url() alias of [django.urls.re_path()](https://docs.djangoproject.com/en/4.2/ref/urls/#django.urls.re_path) is deprecated.
* The {% ifequal %} and {% ifnotequal %} template tags are deprecated in favor of [{% if %}](https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#std-templatetag-if).

[^1]: https://www.djangoproject.com/download/#supported-versions

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,7 @@ setup(
         # minimum requirements for `omero web start`
         "concurrent-log-handler>=0.9.20",
         "Django>=3.2.19,<4.0",
-        "django-pipeline==2.1.0",
+        "django-pipeline==2.0.7",
         "django-cors-headers==3.7.0",
         "whitenoise>=5.3.0",
         "gunicorn>=19.3",


chris-allan:
Yeah, the last time I think we decided 5.12.0 or whatever it was instead of 6.0.0. The changes here could easily be 5.20.0.

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#1424](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/1424/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/1424/consoleText) for more details.
Possible conflicts:
  - PR #17 will-moore 'Right panel performance fixes'
    - omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
    - omeroweb/webclient/urls.py
  - PR #457 sbesson 'Declare portalocker as external dependency'
    - setup.py

--conflicts

chris-allan:
@jburel: Force pushing these commits messes and rewriting all the history here messes up some other workflows I have going. Is the only thing that needed to happen a fix of the merge conflict from `master` as a result of #457 getting merged?

jburel:
Sorry for messing the history and the workflows. This is the only conflict

chris-allan:
Understood. I'll merge `master` in and force push again unless you're depending on the new `HEAD` here?

jburel:
Nothing depends on it at the moment

chris-allan:
:+1:

All done. Had to re-run one of the jobs due to a PyPI gateway timeout when downloading one of the wheels.

jburel:
Sorry I did not realise you were using this branch in a CI infra

will-moore:
This was in the [merge build today](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/1430/console), which coincided with 6 test failures: https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/1438/testReport/ related to image rendering.

E.g. https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/1438/testReport/OmeroPy.test.integration.gatewaytest.test_image/TestImage/testRenderJpegRegion/ has 

```
serverExceptionClass = ome.conditions.InternalException
    message =  Wrapped Exception: (java.lang.RuntimeException):
org.springframework.dao.DataAccessResourceFailureException: Could not obtain current Hibernate Session; nested exception is org.hibernate.HibernateException: No Hibernate Session bound to thread, and configuration does not allow creation of non-transactional one here
}
```
but I don't know if they're related to this PR?

will-moore:
Testing locally and on merge-ci - all working fine.

setup.py
```diff
@@ -54,13 +54,14 @@ def read(fname):
...
-        "Django>=3.2.18,<4.0",
-        "django-pipeline==2.0.7",
+        "Django>=3.2.18,<4.1",
+        "django-pipeline==2.1.0",
         "django-cors-headers==3.7.0",
         "whitenoise>=5.3.0",
         "gunicorn>=19.3",
         "omero-marshal>=0.7.0",
         "Pillow",
+        "pytz",
```
will-moore, line 64: Is there a new requirement for `pytz`?

chris-allan: Yes, the reasons are detailed in the PR description.

will-moore:
NB: removed my comment about failing tests above as Java tests are also failing so not due to this PR.

will-moore:
Opened PRs for various web apps (see links above ⬆️)

I know that for the last [Django upgrade](https://github.com/ome/omero-web/pull/356) we released all the apps with `omero-web` dependency pinned, but then we removed that and re-released them all *before* the `omero-web` Django upgrade was released. So this seems like a fair bit of unnecessary extra work.

chris-allan:
Since all these changes are completely valid when in use with Django 3.2 we don't necessarily have to follow that pattern. However, I'd expect we will have to when we raise the Django requirement to 4.2. It'll be compatibility chaos otherwise.

will-moore:
Excluding to allow merging of #466 which conflicts...

--exclude

will-moore:
Checking pages with updated templates on merge-ci (with Django 3.2.19):
 - https://merge-ci.openmicroscopy.org/web/webadmin/experimenter/edit/2/
 - Activities with various jobs - script, delete, chgrp
 - Partial delete/chgrp of Fileset
 - Right panel toolbar
 - Tags in right panel
 - Hierarchy panel
 - Preview panel owner thumb
 - Tags page
 - Group / User change menu
 - Change group target
 - pagination controls
 - history page
 - scripts dialog
 - search form

sbesson (approved):
Code changes all look good and in agreement with the upstream changes.

Tested on a deployment alongside a series of Web apps with Django upgraded to 4.0.10. Several endpoints and forms modified by this PR were successfully tested. 

As next steps, f I understand correctly, there are two possible routes here:

- either revert the commit Django uncapping Django and starting releasing forward compatibility improvements in OMERO.web releases dependent on Django 3.2
- or starting to migrate the code base towards Django 4.2

From my side, happy to support either option and/or discuss trade-offs.

setup.py
```diff
@@ -54,13 +54,14 @@ def read(fname):
         "omero-py>=5.7.0",
         # minimum requirements for `omero web start`
         "concurrent-log-handler>=0.9.20",
-        "Django>=3.2.19,<4.0",
-        "django-pipeline==2.0.7",
+        "Django>=3.2.19,<4.1",
+        "django-pipeline==2.1.0",
```
will-moore, line 58: As discussed in web and team meetings: plan is to revert these version bumps so that Django 3.2.x is used (which is what we're testing on merge-ci). This will also be compatible with existing web apps, allowing this PR to be released without needing to wait for new releases of apps.

will-moore (changes requested):

will-moore (approved):
LGTM - thanks.
`"django-pipeline==2.1.0"` is compatible with Django 3.2 so it's fine to keep that change.

chris-allan:
> LGTM - thanks. `"django-pipeline==2.1.0"` is compatible with Django 3.2 so it's fine to keep that change.

Just make sure it's still okay with Python 3.6 in the current Dundee CI environment. If it's a problem, let me know and we can revert e0a7d685280e1b463cb0b9faa1b8efdb08dc99f4 as well.

will-moore:
Hmm - yes I see that Python 3.6 was removed from the `classifiers` in latest release: https://github.com/jazzband/django-pipeline/blob/2.1.0/setup.py

However, merge-ci is running on Python 3.6 and running `omero web start` which runs the `collectstatic` which uses pipeline (I think).
I don't think that the deployment uses the output of `django-pipeline`, so I don't think we have any validation of those generated assets.
So it's probably OK, but reverting that change is probably less work than checking that `django-pipeline 2.1.0` works on python 3.6.

PR #465, opened 2023-05-03
Add MultipleFileField to handle and clean multiple files

will-moore:
Fixes #464

This adds the update suggested at the last link below to conform to the security release below:

https://github.com/ome/omero-web/pull/410
https://www.djangoproject.com/weblog/2023/may/03/security-releases/
https://github.com/django/django/commit/eed53d0011622e70b936e203005f0e6f4ac48965

To test:
 - Build should be green (using latest Django 3.2.19)
 - Multiple file upload should still be working
 - When uploading multiple files to file annotations, ALL files should be validated (e.g. to prevent empty files to be uploaded) instead of only the last file being validated.

Issue #464, opened 2023-05-03
Incompatibility with latest Django 3.2.19 security release 


sbesson:
See https://www.djangoproject.com/weblog/2023/may/03/security-releases/

Django 3.2.19 disabled the ability to upload multiple files using one form field (see https://github.com/django/django/commit/eed53d0011622e70b936e203005f0e6f4ac48965). Unfortunately, we started consuming this mechanism in https://github.com/ome/omero-web/pull/410 to allow multiple file annotations to be uploaded  via the UI.

Currently if Django is installed or upgraded to 3.2.19, OMERO.web 5.17+ will fail to start with the following exception

```
ValueError: ClearableFileInput doesn't support uploading multiple files.
```

As an immediate workaround, capping Django to 3.2.18 should restore the ability to start OMERO.web. This issue should likely be addressed properly in a patch release of OMERO.web restoring compatibility with the Django 3.2.x line.

My inclination would be to disable the feature introduced in #410 for now and look into reintroducing it at a later date using a supported API. 

/cc @knabar @chris-allan @will-moore @pwalczysko @jburel

will-moore:
Looks like the `MultipleFileInput` subclass in https://github.com/django/django/commit/eed53d0011622e70b936e203005f0e6f4ac48965 is worth trying (and preferable to removing the feature)?

wheresvic:
Unfortunately this problem is manifesting itself on a fresh omero web install and leads to the following error:

```
Clearing expired sessions. This may take some time... Traceback (most recent call last):
  File "manage.py", line 75, in <module>
    execute_from_command_line(sys.argv)
  File "/opt/omero/web/venv3/lib/python3.8/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/opt/omero/web/venv3/lib/python3.8/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/opt/omero/web/venv3/lib/python3.8/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/opt/omero/web/venv3/lib/python3.8/site-packages/django/core/management/base.py", line 393, in execute
    self.check()
  File "/opt/omero/web/venv3/lib/python3.8/site-packages/django/core/management/base.py", line 419, in check
    all_issues = checks.run_checks(
[...]

PR #466, opened 2023-05-04
Avoid the creation of empty file annotations

sbesson:
Related to #464 

## Summary 

The default contract of the `FileField` used in the files annotation form is to "validate that non-empty file data has been bound to the form." - see https://docs.djangoproject.com/en/3.2/ref/forms/fields/#django.forms.FileField
Interestingly this check is not happening when adding file annotations. However, the emptiness of the file is definitely checked when uploading a photo in the user settings.

The difference comes from the fact the uploaded files are not bound to the files annotation form. Thus the is_valid() check is largely a no-op and this UI can create annotations with empty files.

This PR fixes this issue by properly binding the uploaded files to the form - see https://docs.djangoproject.com/en/3.2/ref/forms/api/#binding-uploaded-files

## Testing

There are a few scenarios to test here:
- using OMERO.web 5.19.0 and Django 3.2.18, it should be possible to create empty file annotation using the UI. This can be done using the file attachment form and selecting one of multiple files. The annotation should display `0B`
- using this PR and Django 3.2.18, attaching empty files via the dialog UI should be:
   * disallowed when selecting single files
   * disallowed when selecting multiple files with the last selected file being empty
   * allowed when selecting multiple files with the last selected file being non empty - see https://www.djangoproject.com/weblog/2023/may/03/security-releases/.
- using this PR, #465 and Django 3.2.19, the creation of empty file annotations should be disallowed in all cases

### DIFF
--- a/omeroweb/webclient/views.py
+++ b/omeroweb/webclient/views.py
@@ -2399,7 +2399,9 @@ def annotate_file(request, conn=None, **kwargs):
 
     if request.method == "POST":
         # handle form submission
-        form_file = FilesAnnotationForm(initial=initial, data=request.POST.copy())
+        form_file = FilesAnnotationForm(
+            initial=initial, data=request.POST.copy(), files=request.FILES
+        )
         if form_file.is_valid():
             # Link existing files...
             files = form_file.cleaned_data["files"]
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ setup(
         "omero-py>=5.7.0",
         # minimum requirements for `omero web start`
         "concurrent-log-handler>=0.9.20",
-        "Django>=3.2.18,<4.0",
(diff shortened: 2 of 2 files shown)


snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#1436](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/1436/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/1436/consoleText) for more details.
Possible conflicts:
  - PR #458 chris-allan 'Compatibility with Django 4.0'
    - setup.py

--conflicts

will-moore (approved):
Testing on merge-ci, this successfully prevents the upload and annotation of Objects with empty Files.

PR #469, opened 2023-05-09
Temp setGroupForSession to edit current group

### LINKED ISSUE #163: Bug: Edit current group name
See https://www.openmicroscopy.org/qa2/qa/feedback/29225/

```
serverStackTrace = ome.conditions.ValidationException: cannot rename the current group context '***'
at ome.logic.AdminImpl.updateGroup(AdminImpl.java:619)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```

If an Admin renames the group they are currently logged in to, you get a ValidationException.

There is a `can_modify_group` flag, which could be set to False if you're editing your current group. 
https://github.com/ome/omero-web/blob/e5b30a74bab82dffe26a3df4cf6f57a104e9e723/omeroweb/webadmin/views.py#L715

But that might not be the nicest usability - No info to tell you why you can't edit group. Also prevents changing permissions on the group, which I don't *think* should be disabled. Don't know about editing description.
Might need another flag 'current_group' to disable those fields instead of hiding them, and also showing tooltip.

will-moore:
Fixes #163.

This uses `setGroupForSession()` to temporarily switch to a different group if a user tries to edit their current group.

To test:
 - To find your default group, use the command line `omero login` as an Admin which will show the name of your current group
 - Log In to web admin, try to edit the name of that group
 - Should work without error

### DIFF
--- a/omeroweb/webclient/webclient_gateway.py
+++ b/omeroweb/webclient/webclient_gateway.py
@@ -1477,6 +1477,16 @@ class OmeroWebGateway(omero.gateway.BlitzGateway):
 
         """
 
+        temp_switch = False
+        if self.getEventContext().groupId == group.id:
+            # can't update current group
+            temp_switch = True
+            # find a group to temp switch to...
+            gids = [gid for gid in self.getEventContext().memberOfGroups if (gid != 0 and gid != group.id)]
+            if len(gids) == 0:
+                return ["You can't edit your current group!"]
+            self.setGroupForSession(gids[0])
+
         up_gr = group._obj
         up_gr.name = rstring(str(name))
         up_gr.description = (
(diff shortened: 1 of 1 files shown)


will-moore:
@jburel - This seems to be working to allow the editing of your current group. 

If you're only a member of 1 group then editing it will fail, since you don't have any other groups you can switch to.
This will already fail with the ValidationException reported above, but I guess we could also improve that error message.

Otherwise, editing should "just work".

sbesson:
Based on my understanding of the semantics of the Web admin UIm the ability to edit a group is restricted to admins in OMERO.web. If so, shouldn't this user always be member of at least the `system` group as per its admin privileges and could we make use of this assumption?

PR #471, opened 2023-05-16
Add samesite cookie settings

knabar:
Allow setting `samesite` cookie properties, required to properly support CORS configurations:

```
omero config set omero.web.csrf_cookie_samesite Lax
omero config set omero.web.session_cookie_samesite None
```

Reference: 
* https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-SESSION_COOKIE_SAMESITE
* https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-CSRF_COOKIE_SAMESITE

### DIFF
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -551,6 +551,28 @@ CUSTOM_SETTINGS_MAPPINGS = {
             "OMERO.web."
         ),
     ],
+    "omero.web.csrf_cookie_samesite": [
+        "CSRF_COOKIE_SAMESITE",
+        "Lax",
+        str,
+        (
+            "The value of the SameSite flag on the CSRF cookie. "
+            "This flag prevents the cookie from being sent in cross-site "
+            "requests thus preventing CSRF attacks and making some methods of "
+            "CSRF session cookie impossible."
+        ),
+    ],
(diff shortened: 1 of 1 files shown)


will-moore:
It would be great to add some guidance for using these settings to the docs at https://omero.readthedocs.io/en/v5.6.7/sysadmins/unix/install-web/walkthrough/omeroweb-install-centos7-ice3.6.html#setting-up-cors

Even after reading the Django docs, I'm not clear on a use-case for setting these to a non default value?

knabar:
@will-moore Our use case is that we have JavaScript applications that need to request data from OMERO.web via API calls.  Since these applications run under different domain names than OMERO.web, the browser will not send the CSRF and session cookies for the active OMERO.web session with those requests.

Here is the best explanation of the different `SameSite` settings I could find so far:
https://web.dev/samesite-cookies-explained/

knabar:
It’s possible - there are so many restrictions though nowadays that it’s hard to pinpoint sometimes

will-moore (approved):
Checking against https://docs.djangoproject.com/en/3.2/ref/settings/ - looks good
👍

knabar:
Added to documentation at https://github.com/ome/omero-documentation/pull/2317

PR #475, opened 2023-05-25
Key value pairs height

will-moore:
See https://forum.image.sc/t/expansion-height-for-key-value-pairs/81548

The removes the restriction on height of the Key-Value Pairs panel, allowing it to expand to show all its content.

Before: (e.g. https://idr.openmicroscopy.org/webclient/?show=image-1884807)

Screenshot 2023-05-25 at 12 50 11

After: (e.g. https://merge-ci.openmicroscopy.org/web/webclient/?show=image-18188)

Screenshot 2023-05-25 at 12 51 05

Also cleaned some commented/invalid CSS and removed an unnecessary element that causes a gap after the Comments pane (see Before screenshot above).

### DIFF
--- a/omeroweb/webclient/static/webclient/css/layout.css
+++ b/omeroweb/webclient/static/webclient/css/layout.css
@@ -1009,7 +1009,6 @@
 
     .mapAnnContainer {
         position: relative;
-        max-height: 250px;
         overflow-y: auto;
     }
     .keyValueTable {
--- a/omeroweb/webclient/static/webclient/css/layout.css
+++ b/omeroweb/webclient/static/webclient/css/layout.css
@@ -553,8 +553,6 @@
         top:6px ;bottom:0px; left:0px; right:0px;
         position: absolute;
         font-size:1.1em;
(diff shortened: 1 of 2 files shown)


jburel:
A Scrollpane could be considered since the list could be long and it will push the entries out of the viewport

will-moore:
A Scrollpane has been considered (that's what we have now) but it has limitations.
I think it's OK to push the other tabs out of the viewport when you want to view the contents of Key-Value Pairs.
In the 2nd screenshot above, you get much more of what you want to see (KVPs) instead of `Tables 1`, `Tags 0`, `Ratings 0`, `Attachments 0`, `Others 0`.

dominikl:
Looks good to me. I'd agree, it's better than the internal KVP scrollpane. If the KVP panel really is extremely long and you want to see the attachments, comments, etc you can simply collapse the KVP panel.

will-moore:
This is now deployed on idr-testing. cc @francesw @jrswedlow

pwalczysko:
lgtm on idr-testing

PR #487, opened 2023-07-05
Use Image.LANCZOS alias rather than Image.ANTIALIAS

sbesson:
These aliases are strictly equivalent but the latter has been deprecated in Pillow 9.1 and dropped in Pillow 10 - see https://pillow.readthedocs.io/en/stable/releasenotes/10.0.0.html

See also https://github.com/ome/omero-py/pull/376

### DIFF
--- a/omeroweb/webclient/webclient_gateway.py
+++ b/omeroweb/webclient/webclient_gateway.py
@@ -89,7 +89,7 @@ def defaultThumbnail(size=(120, 120)):
     if len(size) == 1:
         size = (size[0], size[0])
     img = Image.open(settings.DEFAULT_IMG)
-    img.thumbnail(size, Image.ANTIALIAS)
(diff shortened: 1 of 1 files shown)


knabar:
For future reference, `Image.LANCZOS` was also deprecated in Pillow 9.1.0 but reinstated in 9.4.0 (see https://pillow.readthedocs.io/en/stable/releasenotes/9.4.0.html#restored-image-constants), so there's a lot of outdated Google results about it being deprecated.

knabar (approved):

sbesson:
Thanks @knabar, there was indeed a deprecation dance happening upstream. Also adding a bit of additional context, `Image.LANCZOS` was chosen rather than the `Image.Resampling.LANCZOS` primarily because the second form would require to introduce a minimal dependency on `Pillow>=9.1.0` which introduced the enum classes.

PR #496, opened 2023-07-26
Update changelog for 5.22.1

knabar:
.


sbesson (approved):

jburel (approved):

jburel:
The only point is that ``Django 4.2 will be required in the next minor release.`` was mentioned in the previous note

PR #498, opened 2023-08-01
Cache script names

knabar:
The right panel re-fetches all available scripts on every render, which can take a significant amount of time.

This PR adds a very simplistic cache to fetching the script names, still with the following shortcomings:

- still fetching script names at least every 60 seconds.  Since only three script names are checked, and their existence likely never changes over the lifetime of the OMERO installation, a single check on startup with indefinite caching might be possible.
- caching happens in class variables and is specific to this function, so is not reusable. A generic memoizer or cache decorator would be preferable.

@mabruce

### DIFF
--- a/omeroweb/webclient/controller/container.py
+++ b/omeroweb/webclient/controller/container.py
@@ -320,7 +320,7 @@ class BaseContainer(BaseController):
         Get the file names of all scripts
         """
         now = time.time()
-        if now - BaseContainer._list_scripts_cache_timestamp < 60:
+        if now - BaseContainer._list_scripts_cache_timestamp < 600:
             # cache 10 minutes
             return BaseContainer._list_scripts_cache
 
--- a/omeroweb/webclient/controller/container.py
+++ b/omeroweb/webclient/controller/container.py
@@ -27,6 +27,7 @@ import omero
 from omero.rtypes import rstring, rlong, unwrap
 from django.utils.encoding import smart_str
 import logging
+import time
 
 from omeroweb.webclient.controller import BaseController
 from omeroweb.webgateway.views import _bulk_file_annotations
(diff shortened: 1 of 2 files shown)


omeroweb/webclient/controller/container.py
```diff
@@ -315,6 +319,11 @@ def list_scripts(self):
         """
         Get the file names of all scripts
         """
+        now = time.time()
+        if now - BaseContainer._list_scripts_cache_timestamp < 60:
+            # cache 10 minutes
```
will-moore, line 324: Cache is currently 1 minute (not 10)?

will-moore:
Looks good and makes sense.

We load some settings from the server for the session at https://github.com/ome/omero-web/blob/5457b2dcc8359869314ee0779e603f973201a959/omeroweb/decorators.py#L287 so this could be added there, or use the same approach elsewhere?

Performance of loading the right panel is also impacted by the Download menu, which checks the size of every selected Image (expensive if lots of images selected). My PR at https://github.com/ome/omero-web/pull/17 mitigates that by only loading the Download menu when the user clicks on it.
We could do the same for scripts menu (since it checks the sizes of every Image until it finds one with `sizeC > 1`), although maybe not necessary at this point.

knabar:
@will-moore 

I changed the approach to what is done in https://github.com/ome/omero-web/blob/5457b2dcc8359869314ee0779e603f973201a959/omeroweb/decorators.py#L287 and am storing the retrieved scripts in the request session now.

If the `list_scripts` or `listFigureScripts` are called from outside `omero-web`, the code should still function, as the `request` argument is not required.

will-moore (approved):
Tested locally and the code is working as expected.
Not included in the merge-ci build as this was "draft" PR until just today - can test there on Monday

will-moore:
This is now being included in merge-ci build and tested there today - all looking good 👍

PR #505, opened 2023-10-25
Test proposed scripts button changes

will-moore:
I was just trying out some suggestions from https://forum.image.sc/t/mini-ui-edits-for-maxi-ux-benefits/87821 and thought I'd open a PR just to save them (and give others a chance to try)...

Before:
Screenshot 2023-10-25 at 12 01 06

After:
Screenshot 2023-10-25 at 12 01 23

### DIFF
--- a/omeroweb/webclient/templates/webclient/base/includes/activities_info.html
+++ b/omeroweb/webclient/templates/webclient/base/includes/activities_info.html
@@ -186,6 +186,7 @@
         <li id="queue">
 			
             <span class="toolbar_button" id="launch_activities" title="Activities">
+                Activities
                 <span id="jobstatus" class="notifier">0</span>
                 <span id="runningStatus" style="display:none;"></span>
             </span>
--- a/omeroweb/webclient/templates/webclient/base/includes/script_launch.html
+++ b/omeroweb/webclient/templates/webclient/base/includes/script_launch.html
@@ -21,7 +21,9 @@
 
     <li>
         <!-- Script launcher -->
-        <span id="scriptButton" class="scriptButton toolbar_button" href="{% url 'list_scripts' %}" title="Run Script"></span>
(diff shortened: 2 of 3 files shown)


will-moore:
cc @knabar - What do you think of this change?
@jburel and @pwalczysko are in favour of making these buttons configurable and not to change the default behaviour. I agree that this would minimise the impact: no changes needed for omero-guides etc and this includes Glencoe.
But, I think this change looks like an improvement and configuration should only be used if different users *need* different UIs, since it has a cost for us to implement/test and users to configure.

knabar:
@will-moore My initial thought is that the easiest would be to utilize CSS to have users who want to do so change the buttons.

For example, by adding the following CSS, you can change the button from having an icon to having text, without needing to add configuration settings or changing the behavior for everybody:

```css
#scriptButton::before {
    content: 'Scripts';
    line-height: 23px;
    color: lightgrey;
    font-size: larger;
    padding: 5px;
}
#scriptButton {
    background: none;
}
```

Getting the CSS onto the page would (I think) currently require overriding a template, but we have that process already.

will-moore:
@knabar "easiest" for us, since we do nothing! But not so easy for users. Do we have any docs for overriding a template (other than the Django docs themselves)?

My feeling is that I agree with https://forum.image.sc/t/mini-ui-edits-for-maxi-ux-benefits/87821 that this is actually an improvement for all users, so we should just make the change for everyone.

pwalczysko (changes requested):
LGTM on merge-ci. But the discussion here is obviously important with regards to the customization strategies. Can I ask whether the discussion has crystalized with respect to this very PR ?

will-moore:
I think the consensus is to not change this for everyone as proposed in this PR but instead to improve the customisation workflow and docs.
I'll close this now...

PR #512, opened 2023-11-14
Add Python 3.11 to the testing matrix

sbesson:
Consumes the upstream changes released in https://github.com/ome/omero-py/blob/master/CHANGELOG.md#5170-november-2023 to add Python 3.11 support, this should increase the testing coverage of OMERO.web on Python 3.11


jburel (approved):
This matches the changes i made when evaluating the impact on adding python 3.11 to the testing matrix

PR #514, opened 2023-11-16
Add support for ?show=fileset-123

will-moore:
Often it's useful to find Images etc by Fileset, e.g. when an import completes with...

```
...
Other imported objects:
Fileset:6314846

==> Summary
444 files uploaded, 1 fileset created, 1 image imported, 0 errors in 0:13:35.110
```

This PR adds support for `webclient/?show=fileset-6314846`. Since we can't select a Fileset in the webclient, we simply pick a Image from that Fileset, similar to support for `?show=roi-123`.

### DIFF
--- a/omeroweb/webclient/show.py
+++ b/omeroweb/webclient/show.py
@@ -276,7 +276,7 @@ class Show(object):
             params = omero.sys.ParametersI()
             params.addId(attributes["id"])
             params.page(0, 1)
-            query = "select img from Image img where img.fileset.id=:id"
+            query = "select img from Image img where img.fileset.id=:id order by img.id"
             first_image = self.conn.getQueryService().findAllByQuery(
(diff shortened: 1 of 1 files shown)


knabar:
@will-moore Works fine for me, but it does seem to pick the last (or a random?) image in the fileset; would it make sense to always select the first image (presumably lowest ID)?

will-moore:
Thanks @knabar - fixed now to pick the "first" image (lowest ID).

knabar (approved):

pwalczysko (commented):
built the omero-web locally as our merge-ci is not available.
Indeed, when a second image in the fileset is selected, than a url like /myserlver/weblcient/?show=fileset-79 shows the first image in that fileset. Lgtm. Thanks @knabar for going ahead, just adding this for completeness sake.

Issue #517, opened 2023-12-05
Performance with large numbers of FileAnnotations


chris-allan:
On a busy long running system, particularly one where image data is being run through one or more analytical pipelines, the number of `FileAnnotations` in a given group can easily number in the 1000s. We have exposure to some systems where it numbers in the 10s of thousands. The current "File Attachment" dialog assumes that it can provide a list of every `FileAnnotation` not currently linked to the selected object(s) for linking; performing quite poorly at 10^3, terribly at 10^4 and causes server errors at 10^5.

Not sure if this is also be a problem for OMERO.insight, @jburel + @dominikl.

This SQL query gives a `FileAnnotation` vs. group summary to get a sense of the scale on a particular install:

```
SELECT count(*), experimentergroup.id, experimentergroup.name, experimentergroup.description FROM annotation
    JOIN experimentergroup ON annotation.group_id = experimentergroup.id
        WHERE discriminator = '/type/OriginalFile/' 
            GROUP BY experimentergroup.id;
```

/cc @emilroz, @muhanadz, @erindiel

chris-allan:
Thanks, @will-moore.

I've done a little bit of homework to hopefully help us make some good decisions here. Multiple parent support was added in ome/openmicroscopy#570 which was a rebase of ome/openmicroscopy#557. The initial implementation for "orphan" annotation lookup was added in ome/openmicroscopy@e4534d37711d3d8c24fa09666781d859e750ce69 as part of the refactoring detailed in https://trac.openmicroscopy.org/ome/ticket/3650. The naming is unfortunate as what are looked up are not really orphans but rather annotations that are not linked to the parent object(s). Furthermore, there are various hidden features, such as how objects with an unset namespace are handled, that remain undocumented. It's clear from limited investigation that much of our code relies on these undocumented features.

The call stack is loosely:

`omeroweb.webclient.views.annotate_file()`

to

`omeroweb.webclient.controller.container.getFilesByObject()`

to

`omero.gateway.BlitzGateway.listOrphanedAnnotations()`

As briefly raised by @joshmoore on #499, if we want to handle large numbers of FileAnnotations in a better way, predictable counting and paging are baseline requirements. Achieving this with the current API would be a significant refactor across both `omero-py` and `omero-web` and require updating the `omero-py>=5.7.0` dependency (over 3 years old) in `omero-web`.

In order to tackle performance and safety, there are three main problem areas:

[...]

PR #519, opened 2023-12-08
Limit number of FileAnnotations retrieved for OMERO.web UI

knabar:
Addresses issue #517

The current "Choose attachment" dialog retrieves a list of every `FileAnnotation` not currently linked to the selected object(s) for linking, causing poor performance or failures on systems with a large number of `FileAnnotation`s.

The goal of this PR is to address the potential performance impact only; any other improvements e.g. to the UI should be addressed in separate PRs.

Main discussion points:

### API changes

1. `getFilesByObject` now returns a tuple of the number of total files and a list of `FileAnnotationShim`s; previously the method returned a single list of `FileAnnotation`s. 
2. In case of no objects passed to the controller, `getFilesByObject` previously returned all `FileAnnotation`s, and will now raise a `ValueError`.  This also means `listFileAnnotations` is no longer used and could be deprecated.

### Testing

There are currently no stated expectations to the exact behavior of the "Choose attachment" dialog or the `getFilesByObject` method.  The UI also does not expose all of the available options of the method.  With the large number of permutations of options and object types (and multiple object types, which are not handled), exhaustive testing will be difficult.  Current testing is likely minimal.

- We should document the expected behavior of the "Choose attachment" dialog in the possible situations.
- A testing plan is needed.

Potential pitfalls:
- It is currently possible for multiple users to link the same annotation to the same parent object in a read-annotate group

### Discussion

- What should be the limit on number of `FileAnnotation`s returned?  Currently set to `100`, but suggested to increase to several hundred.
- How should `FileAnnotation`s be sorted? Currently by descending `updateEvent.time`, but suggested to use `name` as previously.
- Indication in the dialog that the results are truncated; space is currently limited, but some indication that not all available annotations are shown is likely needed.

### DIFF
--- a/omeroweb/webclient/templates/webclient/annotations/files_form.html
+++ b/omeroweb/webclient/templates/webclient/annotations/files_form.html
@@ -60,7 +60,10 @@
 
 <p>
     <h1>And/or attach an existing File:</h1>
-    <span>Select the attachments to add (total: {{ total_files }})</span>
+    <span>
+      Select the attachments to add
+      (total: {{ total_files }}{% if total_files > 500 %}, showing 500{% endif %})
+    </span>
     
     <div>{{ form_file.files }}</div>
     <!-- hidden fields used to specify the objects we're tagging -->
--- a/omeroweb/webclient/templates/webclient/annotations/files_form.html
+++ b/omeroweb/webclient/templates/webclient/annotations/files_form.html
@@ -60,7 +60,7 @@
 
 <p>
     <h1>And/or attach an existing File:</h1>
-    <span>Select the attachments to add</span>
+    <span>Select the attachments to add (total: {{ total_files }})</span>
     
     <div>{{ form_file.files }}</div>
     <!-- hidden fields used to specify the objects we're tagging -->
--- a/omeroweb/webclient/templates/webclient/annotations/files_form.html
+++ b/omeroweb/webclient/templates/webclient/annotations/files_form.html
@@ -62,7 +62,7 @@
     <h1>And/or attach an existing File:</h1>
     <span>
       Select the attachments to add
-      (total: {{ total_files }}{% if total_files > 500 %}, showing 500{% endif %})
+      (total: {{ total_files }}{% if total_files > max_files %}, showing {{ max_files }}{% endif %})
     </span>
     
     <div>{{ form_file.files }}</div>
--- a/omeroweb/webclient/controller/container.py
+++ b/omeroweb/webclient/controller/container.py
@@ -24,10 +24,11 @@
 #
 
 import omero
-from omero.rtypes import rstring, rlong, unwrap
+from omero.rtypes import rstring, rlist, rlong, unwrap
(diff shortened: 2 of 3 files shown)


will-moore:
With a bunch of FileAnnotations created with https://github.com/will-moore/python-scripts/blob/main/create_file_annotations.py this looks like:

Screenshot 2023-12-11 at 09 47 48

If I add a bunch of FileAnnotations, then these are correctly excluded from the dialog next time I open it and the count goes down accordingly.

Do we want to boost the page size and sort alphabetically as discussed?

knabar:
> Do we want to boost the page size and sort alphabetically as discussed?

@will-moore I'm working on that today.

will-moore:
I think this PR is failing tests at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/lastCompletedBuild/testReport/OmeroWeb.test.integration.test_annotate/TestFileAnnotations/test_batch_add_fileannotations_2_/

The test is failing to exclude an annotation previously linked to the selected Projects.
I tested in merge-ci webclient and found the same thing with FileAnnotations previously created by upload. This may be because they don't have a namespace??

pwalczysko (changes requested):
the feature works, but I am being offered in the dialog attachments which I already attached. Basically concurring with @will-moore ' testing results

knabar:
@will-moore @pwalczysko I added the `NULL` check for the namespace to bring the new query more in line with the old one

omeroweb/webclient/templates/webclient/annotations/files_form.html
```diff
@@ -60,7 +60,10 @@ <h1>{% trans "Upload a new file" %}:</h1>
 
 <p>
     <h1>And/or attach an existing File:</h1>
-    <span>Select the attachments to add</span>
+    <span>
+      Select the attachments to add
+      (total: {{ total_files }}{% if total_files > 500 %}, showing 500{% endif %})
```
will-moore, line 65: Minor point: hard-coding `500` here and in `container.py` feels a bit brittle.
I guess you could make a `MAX_FILES` variable and add it into the `context`, or add `'file_count': len(files)` to the page context and then do `{% if total_files > file_count %}, showing {{ file_count }} {% endif %}`

will-moore:
Working fine: 

Screenshot 2023-12-13 at 15 37 55

pwalczysko:
Works as expected on merge-ci. lgtm

pwalczysko (approved):

omeroweb/webclient/templates/webclient/annotations/files_form.html
knabar: Thanks @will-moore, I wanted to do exactly that and then promptly forgot about it. Fixed now.

will-moore (approved):
Changes look good - tested locally (merge-ci unavailable just now, but had tested there earlier) - all working.

pwalczysko (approved):

PR #529, opened 2024-01-26
Preview viewer uses omero-web-zarr s3 data if available

will-moore:
**Experimental!**

This corresponds to https://github.com/ome/omero-web-zarr/pull/15.
In that `omero-web-zarr` PR, we check whether OME-Zarr data is publicly accessible via s3 URL and if so then we can use that data for `/zarr/render_image/` doing the rendering in python omero-web code rather than using the server.

In this PR, we test for the availability of the `omero-web-zarr` plugin and the public s3 url, and if we have that then the Preview viewer uses `/zarr/` as the `base url`.


will-moore:
Following discussion with IDR today, closing this for now since we want to make minimal changes for now.

PR #530, opened 2024-02-06
"Others" annotation display tweaks

DavidStirling:
Lately we've been experimenting with [ListAnnotation](https://omero.readthedocs.io/en/stable/developers/Model/EveryObject.html#omero-model-class-listannotation) objects as a means to group other annotations together. Within OMERO.web's right panel these are apparently displayed under the "Others" section when populating the annotation list.

My understanding is that this section is intended to provide a basic display of miscellaneous or custom annotation types. The display attempts to show an entry for each with the format "Type: Value".

However, for ListAnnotations specifically there is no actual value field associated with them. The end result is this:

Screenshot 2024-02-06 at 09 35 33

To further complicate things, the tooltips providing more details are bound to the value field, meaning that the user cannot see any further details about these annotations. Attach 10 ListAnnotation objects and you'll just see the word "List" 10 times.

To attempt to improve this I've modified the code that handles this to fall back to using the 'name' field if no other value fields are present. This field is common to all annotation types (though optional) and might be a sensible thing to display in the absence of anything else. The result would be:

Screenshot 2024-02-06 at 09 35 15

Having some data in the name field also then restores the ability to view tooltips.

Unfortunately the **webclient json API** for annotations currently seems to return everything except for the name field, I'm not sure why. For the sake of this PR I've added that field to the returned results. Hope that doesn't break anything.

**Testing:** Create ListAnnotations with name/description/namespace info and attach them to an object. Without PR: OMERO.web right panel only shows "List" for each. With PR: Right panel shows "List: [name]". Other BasicAnnotation subtypes should still display their respective values, though their tooltips will now include the "name" field.

### DIFF
--- a/omeroweb/webclient/static/webclient/javascript/ome.right_panel_customanns_pane.js
+++ b/omeroweb/webclient/static/webclient/javascript/ome.right_panel_customanns_pane.js
@@ -85,7 +85,7 @@ var CustomAnnsPane = function CustomAnnsPane($element, opts) {
                     ann.addedBy = [ann.link.owner.id];
                     // convert 'class' to 'type' E.g. XmlAnnotationI to Xml
                     ann.type = ann.class.replace('AnnotationI', '');
-                    var attrs = ['textValue', 'timeValue', 'termValue', 'longValue', 'doubleValue', 'boolValue'];
+                    var attrs = ['name', 'textValue', 'timeValue', 'termValue', 'longValue', 'doubleValue', 'boolValue'];
                     attrs.forEach(function(a){
                         if (ann[a] !== undefined){
                             ann.value = ann[a];
--- a/omeroweb/webclient/static/webclient/javascript/ome.right_panel_customanns_pane.js
+++ b/omeroweb/webclient/static/webclient/javascript/ome.right_panel_customanns_pane.js
@@ -85,7 +85,7 @@ var CustomAnnsPane = function CustomAnnsPane($element, opts) {
                     ann.addedBy = [ann.link.owner.id];
                     // convert 'class' to 'type' E.g. XmlAnnotationI to Xml
                     ann.type = ann.class.replace('AnnotationI', '');
-                    var attrs = ['name', 'textValue', 'timeValue', 'termValue', 'longValue', 'doubleValue', 'boolValue'];
+                    var attrs = ['textValue', 'timeValue', 'termValue', 'longValue', 'doubleValue', 'boolValue'];
                     attrs.forEach(function(a){
                         if (ann[a] !== undefined){
                             ann.value = ann[a];
--- a/omeroweb/webclient/templates/webclient/annotations/custom_ann_tooltip.html
+++ b/omeroweb/webclient/templates/webclient/annotations/custom_ann_tooltip.html
@@ -21,6 +21,7 @@
 
 <span class="tooltip_html" style='display:none'>
     <b>ID:</b> {{ ann.id }}<br />
+    {% if ann.name %}<b>Name:</b> {{ ann.name }}<br />{% endif %}
     {% if ann.ns %}<b>Namespace:</b> {{ ann.ns }}<br />{% endif %}
     {% if ann.description %}<b>Description:</b> {{ ann.description }}<br />{% endif %}
     <b>Owner:</b> {{ ann.getOwner.getFullName }}<br />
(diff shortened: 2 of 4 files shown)


sbesson (commented):
From a code perspective, the changes proposed here make sense to me. The `name` attribute has been introduced on all annotations in OMERO 5.1.0 (and specifically https://github.com/ome/openmicroscopy/pull/3362), so I suspect the code had simply never been updated accordingly.

Adding a few `ListAnnotation` objects using 

```
omero obj new DatasetAnnotationLink parent=<Object>:<id> child=$(omero obj new ListAnnotation name=<name> ns=<ns>)
```

the UI now displays the new elements with the name populated

Screenshot 2024-02-07 at 08 51 36

DavidStirling:
Thanks for the background @sbesson, I guess that does raise whether it's worth also adding the "name" field to tooltips for other annotation types? For this PR I only updated the "custom annotation" template.

omeroweb/webclient/static/webclient/javascript/ome.right_panel_customanns_pane.js
```diff
@@ -85,7 +85,7 @@ var CustomAnnsPane = function CustomAnnsPane($element, opts) {
                     ann.addedBy = [ann.link.owner.id];
                     // convert 'class' to 'type' E.g. XmlAnnotationI to Xml
                     ann.type = ann.class.replace('AnnotationI', '');
-                    var attrs = ['textValue', 'timeValue', 'termValue', 'longValue', 'doubleValue', 'boolValue'];
+                    var attrs = ['name', 'textValue', 'timeValue', 'termValue', 'longValue', 'doubleValue', 'boolValue'];
```
will-moore, line 88: I think it's confusing to use `name` here since this logic is really for finding the `value` of various annotations and it's not clear why for some annotations that you'd want the name instead.

I would think it's better to update the display logic for List Annotations to show the `name` if there is no value.

DavidStirling: @will-moore That's absolutely fair, my rationale here was to try to touch as little as possible but updating the template might be more sensible with wider input.

The current rendering logic is [here](https://github.com/ome/omero-web/blob/9841167a70df2f0174aed2820bcd302873b53df3/omeroweb/webclient/templates/webclient/annotations/customanns_underscore.html#L8-L14). There is already a specific case for Xml annotations.

I could see a few ways of approaching this. You could have another case specifically for `List` annotation types, which would address the problem here but not for any other custom annotation types the user has defined. That said, I'm not sure if anyone actually does that sort of thing.

As a broader issue it looks like, as with `ann.name`, the `ann.value` fields on other BasicAnnotation subtypes are also all optional, so these types will also fail to show a tooltip if there is no value data associated with them.

Perhaps it'd be worth implementing logic along the lines of:

- If it's xml, show the xml details
- else if it has a value, show the value
- else if it's a List and has a name, show the name,
- otherwise show something like "(No data)"

I think that might ensure that at least something is displayed in a manner that's predictable and consistent for users?

will-moore: Sounds good 👍 
What about trying to show more info for List Annotations? I would have thought that users would find it handy to be able to see the length of the list or even some summary of the contents of the list? E.g. `[Tag, Tag, Comment]` or the values , ns of these? Although I realise that would involve more work on loading these into the JSON in tree.py.

will-moore:
There are a bunch of tests failing at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/349/ due to the addition of `Name: None` to the JSON.

Need to update the expected JSON values at e.g. https://github.com/ome/openmicroscopy/blob/develop/components/tools/OmeroWeb/test/integration/test_tree_annotations.py#L307

DavidStirling:
Okay, pushed a revised template which should now handle everything as proposed above instead of fudging the 'value' parameter.

This should now handle cases where BasicAnnotation subtypes have missing values, like so:

Screenshot 2024-02-07 at 10 52 53

There's probably more useful info for Lists that could be shown as Will mentioned, but for now I hope this will suffice. If ListAnnotations start getting used more in practice then it might be something for a new PR.

Didn't realise that tests are in a different repo, I guess I should open a companion PR there?

will-moore:
Yes, you'll need a companion PR for tests. Legacy of omero-web once being in the openmicroscopy repo.

will-moore (approved):
Looks good, thanks.
I think the tests didn't run last night but should be fixed...

PR #534, opened 2024-02-22
get strings from omero.cmd.ERR and omero.CmdError correctly

### LINKED ISSUE #481: AttributeError: 'ERR' object has no attribute 'message'
https://www.openmicroscopy.org/qa2/qa/feedback/33051/

```
Traceback (most recent call last):

File "/omero/venv3/lib64/python3.6/site-packages/django/core/handlers/exception.py", line 47, in inner
response = get_response(request)

File "/omero/venv3/lib64/python3.6/site-packages/django/core/handlers/base.py", line 181, in _get_response
response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/omero/venv3/lib64/python3.6/site-packages/omeroweb/decorators.py", line 544, in wrapped

File "/omero/venv3/lib64/python3.6/site-packages/omeroweb/decorators.py", line 603, in wrapper
context = f(request, *args, **kwargs)

File "/omero/venv3/lib64/python3.6/site-packages/omeroweb/webclient/views.py", line 3555, in activities
rv = graphResponseMarshal(conn, rsp)

File "/omero/venv3/lib64/python3.6/site-packages/omeroweb/webgateway/marshal.py", line 448, in graphResponseMarshal
rv["error"] = rsp.message

AttributeError: 'ERR' object has no attribute 'message'
```

will-moore:
Fixes #481 

This fixes 2 similar bugs where we're handling `omero.CmdError` or `omero.cmd.ERR` objects and converting into strings.

Both were previously failing with `AttributeError: 'ERR' object has no attribute 'message'`.

Changes based on using the appropriate attributes as tested at https://github.com/ome/omero-web/issues/481#issuecomment-1959903979

### DIFF
--- a/omeroweb/webclient/webclient_gateway.py
+++ b/omeroweb/webclient/webclient_gateway.py
@@ -1591,7 +1591,7 @@ class OmeroWebGateway(omero.gateway.BlitzGateway):
         try:
             cb = self.c.submit(command, loops=120)
         except omero.CmdError as ex:
-            message = ex.err.message
+            message = str(ex.err)
         finally:
             if cb:
                 cb.close(True)
--- a/omeroweb/webgateway/marshal.py
+++ b/omeroweb/webgateway/marshal.py
@@ -445,7 +445,7 @@ def graphResponseMarshal(conn, rsp):
     rv = {}
     if isinstance(rsp, omero.cmd.ERR):
         rsp_params = ", ".join(["%s: %s" % (k, v) for k, v in rsp.parameters.items()])
-        rv["error"] = rsp.message
+        rv["error"] = str(rsp.category)
         rv["report"] = "%s %s" % (rsp.name, rsp_params)
     else:
         included = rsp.responses[0].includedObjects


sbesson (commented):
The usage of `str(omero.CmdError)` matches the intent of this API introduced in https://github.com/ome/omero-py/pull/151

For the `omero.cmd.ERR` handling, as an element of comparison the CLI implementation in https://github.com/ome/omero-py/blob/0dc60bd67a5f9281e2063a0a3aeeaa119c14447d/src/omero/cli.py#L1907-L1927 uses `rsp.name` as the error message and `rsp.parameters` as the report. Would it make sense to align the behavior in OMERO.web

will-moore:
I tried `str(err)` for both cases at https://github.com/ome/omero-web/issues/481#issuecomment-1959903979 and it gave sub-optimal results.
Not sure exactly what the difference is between `omero.CmdError` and `omero.cmd.ERR`!

will-moore:
@sbesson updated as suggested to use `str(omero.CmdError)` and `str(omero.cmd.ERR.name)`

will-moore:
cc @knabar This is not an urgent PR, but it is a trivial fix so could maybe get reviewed and included in the next release?

sbesson:
@will-moore I have not had a chance to get back to those. For both updates, do you have scenarios that would allow to reproduce the original errors and asses the impact of the changes functionally?

will-moore:
@sbesson No, unfortunately I don't have scenarios to reproduce those errors. We only have stack traces from QA, but don't know what caused them (difficult to understand from the stack traces themselves).
I could try guessing a few, but not sure where to start

knabar:
@will-moore I'm curious if these missing attributes existed in the past, and if so, when they were removed or renamed.  Discussed with @sbesson a bit and he pointed out this could be related to Python 2/3 upgrades as well.

Overall I feel we need a bit more investigation, so I'd like to hold off on this one until 5.26.0, which could come fairly soon anyway.

will-moore:
@knabar I don't think they've ever existed. It was just assumed that an exception would have a `message` attribute.

PR #535, opened 2024-02-23
Initial working browser history

will-moore:
Work in progress...

When you browse in webclient, the URL updates to reflect the currently selected object.
You can use browser history (back and forward buttons) to go to previously-selected objects etc.

TODO: need to handle selection of Wells....


will-moore:
@joshmoore - reported "back button" as a missing feature of webclient, which is what this PR aims to address

joshmoore:
Just for reference, in my case I was clicking on a large number of HCS wells and then on given fields, waiting for the right-hand panel to load to see what the annotations on that image were, and then moving on to the next field or well.

PR #536, opened 2024-02-23
imgData JSON with no stateful services

will-moore:
It has been noted in some cases that the Preview panel and the right panel of iviewer can take a long time to load, since we currently initialise various stateful services (which requires initialising them with the original image files). This means that the viewers are slower to load than if we just get the data from the DB, particularly if access to original data is slow.

Stateful services usage is replaced as follows:

 - Instead of using rendering engine for settings, we load rendering def
 - Instead of the rawPixelsStore for `image.getPixelRange()`, we lookup values based on pixel type
 - Instead of using rendering engine for `requiresPixelsPyramid()` we use `image.requiresPixelsPyramid()` which simply uses the size cut-off.
 - IF the image is BIG, *then* we load the rendering engine to get the actual pyramid resolutions

To test:
 - Image viewer should work as before but load faster
 - The only other change in behaviour is for small images that have a pyramid, such as small NGFF images. These will appear as small images in the Preview panel and will load as whole planes instead of tiles in iviewer.
 - Instead of using `re.lookupRenderingDef(pid, ctx)` for the logic of "use my rdef or use the owners", this is now encoded in `webgateway.util.get_rendering_def()`
 - If no rendering def exists, we use rendering engine to create one as before.

Also see performance testing with NGFF data on idr-testing at https://github.com/ome/omero-web/pull/537#issuecomment-2004001921
Summarised as follows:
 - Don't expect this PR to noticeably improve performance for *most* data in OMERO
 - For data where we do notice a significant improvement (e.g. idr0090 NGFF data) even if we always use RenderingEngine to test/load the pyramid levels (ie. NO api change), this PR still has a performance improvement.

PR #531, opened 2024-02-06
Removal of python-future compatibility code 

knabar:
This PR implements issue #528 to review the OMERO.web source code to update all paths with either Python 2 or Python 2/3 logic to pure Python 3+, similar to @sbesson's work for OMERO.py at https://github.com/ome/omero-py/pull/390.

PR #537, opened 2024-02-25
Img data json django32

will-moore:
This is identical to #536, except this is based off v5.22.1 and so is compatible with Django 3.2.

# Performance testing

NB: this PR includes TEMP A/B testing

- A) Images with EVEN numbered IDs: - we check the size of the image and only load rendering engine for resolution levels if it's BIG
- B) images with ODD numbered IDs - we ALWAYS load the rendering engine to check `re.requiresPixelsPyramid()`, if so, we also load resolution levels.

I would expect A (EVEN numbers) to be faster for small images BUT it seems that B is faster?!?! Possibly because the size threshold is using config service to get the BIG image limit.

# Memo file requirement

I had hoped that by checking image size, small images (EVEN IDs) would NOT need to read memo file in order to load `imgData` (required for iviewer right panel).

BUT, seems that the `imgData` won't load for e.g. idr0090 images where memo file is missing. e.g. https://idr-testing.openmicroscopy.org/webclient/imgData/12542998/ (504 timeout).

### DIFF
--- a/omeroweb/webgateway/marshal.py
+++ b/omeroweb/webgateway/marshal.py
@@ -64,8 +64,10 @@ def getChannelsNoRe(image):
                 left outer join fetch c.statsInfo
                 where p.id=:id"""
     pixels = conn.getQueryService().findByQuery(query, params, conn.SERVICE_OPTS)
-    return [ChannelWrapper(conn, c, idx=n, img=image)
-            for n, c in enumerate(pixels.iterateChannels())]
+    return [
+        ChannelWrapper(conn, c, idx=n, img=image)
+        for n, c in enumerate(pixels.iterateChannels())
+    ]
 
 
 def eventContextMarshal(event_context):
@@ -113,7 +115,8 @@ def getPixelRange(image):
         PixelsTypeint32: -2147483648,
         PixelsTypeuint32: 0,
         PixelsTypefloat: -2147483648,
-        PixelsTypedouble: -2147483648}
+        PixelsTypedouble: -2147483648,
+    }
     maxVals = {
         PixelsTypeint8: 127,
         PixelsTypeuint8: 255,
@@ -122,7 +125,8 @@ def getPixelRange(image):
         PixelsTypeint32: 2147483647,
         PixelsTypeuint32: 4294967295,
         PixelsTypefloat: 2147483647,
-        PixelsTypedouble: 2147483647}
+        PixelsTypedouble: 2147483647,
+    }
     pixtype = image.getPrimaryPixels().getPixelsType().getValue()
     return [minVals[pixtype], maxVals[pixtype]]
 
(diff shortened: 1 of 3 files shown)


snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#468](https://snoopycrimecop-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/468/). See the [console output](https://snoopycrimecop-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/468/consoleText) for more details.
Possible conflicts:
  - PR #531 knabar 'Removal of python-future compatibility code '
    - omeroweb/webgateway/marshal.py
    - omeroweb/webgateway/util.py
  - PR #534 will-moore 'get strings from omero.cmd.ERR and omero.CmdError correctly'
    - omeroweb/webgateway/marshal.py
  - PR #536 will-moore 'imgData JSON with no stateful services'
    - omeroweb/webgateway/marshal.py
    - omeroweb/webgateway/util.py

--conflicts

omeroweb/webgateway/marshal.py
```diff
@@ -264,9 +345,36 @@ def pixel_size_in_microns(method):
...
+    # If image is big - need to load RE to get resolution levels
+    # NB: some small images like OME-Zarr can have pyramids
+    # but these will be ignored
+
+    # TEMP - for A/B testing only use size test if ID is EVEN number!
+    if image.id % 2 == 0 and not image.requiresPixelsPyramid():
+        rv["tiles"] = False
+    else:
+        if load_re(image, rv):
+            levels = image._re.getResolutionLevels()
```
will-moore, line 363: NB: TEMP: only use the `image.requiresPixelsPyramid()` if image ID is an EVEN number.

will-moore:
Testing with this PR deployed on idr-testing definitely felt faster to load right panel of iviewer.
Closing now.

will-moore:
Redeploying the A/B/C testing on idr-testing for more perf testing...

will-moore:
Summary of performance testing...

In the chart below, each green bar is the http response time. Blue (where shown - 1st 2 sections only) is the time to load rendering engine as reported in the JSON response with the ABC testing enabled for this PR.

Testing in 4 sessions below:
 - 1st session: idr0010 with this PR
 - 2nd session: idr0090 with this PR
 - 3rd session: idr0090 without this PR
 - 4th session: idr0010 without this PR
 
Screenshot 2024-03-18 at 13 44 22

Conclusions:
 - For idr0010, this PR makes very little (no discernible) difference. Times to load rendering engine to get pyramid levels are about 0.04 secs. Probably about the same for other stateful services: insignificant compared to overall http response times of 0.5 secs
 - For idr0090, loading rendering engine to get pyramid levels takes about 0.5 seconds.
 - Even when we *do* load RE (for pyramid levels) in this PR, http response is about 1.4 secs, which is faster than without this PR (about 1.9 secs) suggesting that for idr0090, `rawPixelsStore` for pixels min/max and `renderingEngine` loading of rendering settings together add an extra 0.5 secs in total (for idr0090).

To summarise:
 - Don't expect this PR to improve performance for most data in OMERO
 - Even if we *always* use RenderingEngine to test/load the pyramid levels, this PR still has a performance improvement.

PR #539, opened 2024-03-06
Fix incorrect maximum length of webadmin and webclient form fields

chris-allan:
Fixes incorrect maximum length of webadmin and webclient form fields. Having incorrect lengths causes scenarios where certain strings are truncated on save or action when used with these forms.

### DIFF
--- a/omeroweb/webadmin/forms.py
+++ b/omeroweb/webadmin/forms.py
@@ -145,7 +145,6 @@ class ExperimenterForm(forms.Form):
 
         if "with_password" in kwargs["initial"] and kwargs["initial"]["with_password"]:
             self.fields["password"] = forms.CharField(
-                max_length=255,
                 widget=forms.PasswordInput(attrs={"size": 30, "autocomplete": "off"}),
             )
             self.fields["confirmation"] = forms.CharField(
--- a/omeroweb/webadmin/forms.py
+++ b/omeroweb/webadmin/forms.py
@@ -66,6 +66,7 @@ class ForgottonPasswordForm(forms.Form):
         super(ForgottonPasswordForm, self).__init__(*args, **kwargs)
         self.fields["server"] = ServerModelChoiceField(Server, empty_label=None)
         f = forms.CharField(
+            max_length=255,
             widget=forms.TextInput(attrs={"size": 28, "autocomplete": "off"}),
         )
         self.fields["username"] = f
--- a/omeroweb/webadmin/forms.py
+++ b/omeroweb/webadmin/forms.py
@@ -48,11 +48,10 @@ class LoginForm(forms.Form):
         self.field_order = ["server", "username", "password"]
 
     username = forms.CharField(
-        max_length=50,
+        max_length=255,
         widget=forms.TextInput(attrs={"size": 22, "autofocus": "autofocus"}),
     )
     password = forms.CharField(
-        max_length=50,
         widget=forms.PasswordInput(attrs={"size": 22, "autocomplete": "off"}),
     )
 
@@ -67,7 +66,6 @@ class ForgottonPasswordForm(forms.Form):
         super(ForgottonPasswordForm, self).__init__(*args, **kwargs)
         self.fields["server"] = ServerModelChoiceField(Server, empty_label=None)
         f = forms.CharField(
-            max_length=50,
             widget=forms.TextInput(attrs={"size": 28, "autocomplete": "off"}),
         )
         self.fields["username"] = f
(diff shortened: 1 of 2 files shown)


omeroweb/webadmin/forms.py
```diff
@@ -67,7 +66,6 @@ def __init__(self, *args, **kwargs):
         super(ForgottonPasswordForm, self).__init__(*args, **kwargs)
         self.fields["server"] = ServerModelChoiceField(Server, empty_label=None)
         f = forms.CharField(
-            max_length=50,
```
sbesson, line 70: As this is a `username` should the max length limit be updated to 255 to match the database limit?

sbesson (commented):
One question apart, all the updated length limitations in OMERO.web now match the size limits in the database with the current schema, namely:

- `Experimenter.omeName`, `Experimenter.firstname`, `Experimenter.middlename`, `Experimenter.lastname`, `Experimenter.email`, `Experimenter.institution` are columns with a VARCHAR(255) data type
- all container names e.g. `Project.name`, `Dataset.name` are columns with a TEXT data type
- `LogicalChannel.name` is a column with a VARCHAR(255) data type
- `LogicalChannel.fluor` is a column with a TEXT data type
- `Dichroic.lotnumber`, `Dichroic.serialNumber`, `Dichroic.manufacturer`, `Dichroic.model` are columns with a VARCHAR(255) data type
- `Microscope.lotnumber`, `Microscope.serialNumber`, `Microscope.manufacturer`, `Microscope.model` are columns with a VARCHAR(255) data type
- `Objective.lotnumber`, `Objective.serialNumber`, `Objective.manufacturer`, `Objective.model` are columns with a VARCHAR(255) data type
- `Filter.lotnumber`, `Filter.serialNumber`, `Filter.manufacturer`, `Filter.model`, Filter.filterwheel` are columns with a VARCHAR(255) data type
- `Detector.lotnumber`, `Detector.serialNumber`, `Detector.manufacturer`, `Detector.model` are columns with a VARCHAR(255) data type
- `Lightsource.lotnumber`, `Lightsource.serialNumber`, `Lightsource.manufacturer`, `Lightsource.model` are columns with a VARCHAR(255) data type


Finally `Password.hash` is a column with a VARCHAR(255) data type but that only applies to experiments which password is managed in the database. For all external authentication mechanisms like LDAP, this restriction does not necessary apply so dropping completely the `max_length` imposed in the OMERO.web form is the correct approach.

omeroweb/webadmin/forms.py
chris-allan: Yes, fixed in 3acab1ee69c61fa8b2844d18134d38181017c975.

chris-allan:
> One question apart

Did you have a question beyond the query about the username field in `ForgottonPasswordForm`?

> Finally Password.hash is a column with a VARCHAR(255) data type but that only applies to experiments which password is managed in the database.

Sort of. Its length is irrelevant regardless of the password source as the hash is of fixed length regardless of the size of the input data being hashed.

sbesson (approved):
> Did you have a question beyond the query about the username field in ForgottonPasswordForm?

No, the question on the ForgottenPassword form username length was my only outstanding question.

> Sort of. Its length is irrelevant regardless of the password source as the hash is of fixed length regardless of the size of the input data being hashed.

True i.e. Yet-another-reason to lift all length restrictions imposed on the user password.

From my side, all looks good from a code review.

will-moore (approved):
I wasn't sure if this needs any functional testing (or what's the easiest way to show the original errors)? If you exceed max_length when filling in a form, the form is invalid, so that should fail at validation (if we do it) rather than populating the DB with truncated values?

But the code changes look good. 👍

chris-allan:
> I wasn't sure if this needs any functional testing (or what's the easiest way to show the original errors)?

The easiest way to see it is to create something like this:

```
omero obj new Project name=$(python -c 'print("a" * 250 + "b", end="")')
```

You will now not be able to modify the name of this object in due to form validation. However, since the field actually has 251 characters on the client, if you copy and paste the name back into the name field or into another name field it will be silently truncated. There are lots of other weird examples if you modify text in the middle of the long string, etc.

The worst example is the password field where you cannot actually see what's typed in, people are used to visually truncated password fields, and silent truncation happens underneath on paste.

will-moore:
Thanks. Tested locally with this branch and could update e.g. Project name to a string longer than 250 characters.
Editing the same project when I reverted to master branch wasn't possible with the full name. Looks good.

will-moore:
This seems to be included in merged PRs today at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/478/console but I'm not seeing it at https://merge-ci.openmicroscopy.org/web/webclient/ 
E.g. Edit Project `<input>` still has `maxlength="250"`. Something up with the build...?

will-moore:
This is being deployed on merge-ci now and is working fine.

PR #474, opened 2023-05-24
Detect and display non-zero returncode from scripts

will-moore:
This uses the `returncode` returned from running scripts to show the Error icon in Activities dialog:

Feature requested at https://forum.image.sc/t/omero-scripts-fail-icon/81495

To test:
 - Run a script that throws an Exception *after* setting message, e.g. `client.setOutput("Message", rstring("OK so far..."))`
   - The Message will be shown, along with Error status (first result below)
 - Run a script that throws an Exception without setting "Message" (typical behaviour)
   - This will display a message including the `returncode` from the failure.
 - Otherwise working scripts should behave as before (2 other examples below)

Screenshot 2023-10-18 at 10 24 08

The `context` used to render the html page in the example above looks like this:

```
{
  "sizeOfJobs": 4,
  "jobs": [
    {
      "job_type": "script",
      "job_name": "Images From ROIs",
      "start_time": "datetime.datetime(2023, 10, 18, 10, 15, 20, 121058)",
      "status": "finished",
      "returncode": 1,
      "Message": "OK so far...",
      "stderr": 3725,
      "results": {},
      "id": "52dd4857-9ca1-4dd1-ad0b-5a78d3fdd235",
      "key": "ProcessCallback/52dd4857-9ca1-4dd1-ad0b-5a78d3fdd235 -t -e 1.1:tcp -h 172.19.0.4 -p 37387 -t 60000",
      "new": true
    },
    {
      "job_type": "script",
      "job_name": "Dataset To Plate",
      "start_time": "datetime.datetime(2023, 10, 18, 10, 16, 0, 665853)",
      "status": "finished",
      "returncode": 1,
      "Message": "Script exited with failure. (returncode=1)",
      "stderr": 3724,
      "results": {},
      "id": "087c20ed-0617-4931-9ef4-91046bd1181e",
      "key": "ProcessCallback/087c20ed-0617-4931-9ef4-91046bd1181e -t -e 1.1:tcp -h 172.19.0.4 -p 37387 -t 60000"
    },
    {
      "job_type": "script",
      "job_name": "Dataset To Plate",
      "start_time": "datetime.datetime(2023, 10, 18, 10, 15, 20, 121058)",
      "status": "finished",
      "returncode": 0,
      "Message": " New plate created: From_ROIs but could not be attached.",
      "results": {
        "New_Object": {
          "id": 101,
          "type": "Plate",
          "browse_url": "/webclient/userdata/?show=plate-101",
          "name": "From_ROIs"
        }
      },
[...]

PR #541, opened 2024-03-07
Don't overwrite kwargs in activities json

will-moore:
Reported at https://github.com/ome/omero-figure/issues/545

This fixes a bug introduced in #474, the `kwargs` parameter overwrites the `kwargs` from the function `def activities(request, conn=None, **kwargs):` causing various issues.

E.g. the detection of `json` template fails at https://github.com/ome/omero-web/blob/4d8d25ef7528e49cadfb5b777a47f6953c198fba/omeroweb/webclient/views.py#L3930 so the format of the JSON returned to OMERO.figure is broken, as reported above.

To test:
 - Open the developer Console in your browser
 - Export a figure from OMERO.figure
 - Check that the export completes and the Download button is shown with no errors in the Console

### DIFF
--- a/omeroweb/webclient/views.py
+++ b/omeroweb/webclient/views.py
@@ -3851,16 +3851,16 @@ def activities(request, conn=None, **kwargs):
                     try:
                         # we can only retrieve this ONCE - must save results
                         results = proc.getResults(0, conn.SERVICE_OPTS)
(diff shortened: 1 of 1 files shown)


knabar (approved):
Verified that this fixes the issue with PathViewer

PR #543, opened 2024-03-13
Make long-running tasks like "Save to all" more robust

knabar:
Work in progress

The "Save to all" function in the Preview panel can take a long time to complete for datasets or plates that have lots of images.  The user may refresh or reload the page before everything (including thumbnail refresh) completes, exacerbating the already high load on the server.

This PR introduces a UI blocking dialog until thumbnails are reloaded:

Screenshot 2024-03-13 at 15 03 59

Further work will attempt to reduce the time these tasks take, and to block UI interaction for other long-running tasks during which the user should not interact further.


will-moore:
In testing, I noticed that "Save to All" functionality appears to ignore LUTs. The LUTs are used to generate thumbnails but are not actually saved, leading to a case where the thumbnails (with LUTs) are out of sync with the image rendering settings (without LUTs):
However, this isn't coming from this PR - same behaviour before. 

Screenshot 2024-04-15 at 15 09 23

knabar:
> This looks good and is working fine in testing.
> 
> I wonder if we want to consolidate this with `function gs_modalJson (url, data, callback) {` which is used for the "Save" rendering settings button and gives a different look:
> 
> Having the 2 different blocking UIs for adjacent buttons is slightly unexpected, but certainly not a blocker.

Thanks for pointing that out - I'll dig in a bit and see if I can clean it up.

knabar:
> In testing, I noticed that "Save to All" functionality appears to ignore LUTs. The LUTs are used to generate thumbnails but are not actually saved, leading to a case where the thumbnails (with LUTs) are out of sync with the image rendering settings (without LUTs): However, this isn't coming from this PR - same behaviour before.

Opened a new issue #552

knabar:
> I wonder if we want to consolidate this with `function gs_modalJson (url, data, callback) {` which is used for the "Save" rendering settings button and gives a different look:

`gs_modalJson` was only used in one location, it's now replaced with the new functionality.

will-moore (approved):
Code removal / update looks good and functionality is working, Thanks

PR #549, opened 2024-04-12
Remove webgateway cache

knabar:
This PR removes the webgateway cache functionality for rendered images and refactors the `tempfile` functionality into its own new file.

The motivation behind removing the webgateway cache is the minimal performance improvements from having the cache versus the large overhead introduced with the required cache invalidation.  Tasks like applying rendering settings to a large number of images takes much longer than needed due to cache invalidation.


omeroweb/webgateway/views.py
```diff
@@ -1246,17 +1220,14 @@ def render_ome_tiff(request, ctx, cid, conn=None, **kwargs):
...
-            try:
-                tiff_data = imgs[0].exportOmeTiff()
-            except Exception:
-                logger.debug("Failed to export image (2)", exc_info=True)
-                tiff_data = None
-            if tiff_data is None:
-                webgateway_tempfile.abort(fpath)
-                raise Http404
-            webgateway_cache.setOmeTiffImage(request, server_id, imgs[0], tiff_data)
+            webgateway_tempfile.abort(fpath)
```
will-moore, line 1229: not sure that this line does anything now?

will-moore: Ah - I missed the unchanged `fpath, rpath, fobj = webgateway_tempfile.new(...) ` above, which makes more sense and explains `.abort()`

omeroweb/webgateway/webgateway_tempfile.py
```diff
@@ -0,0 +1,186 @@
+import logging
```
sbesson, line 1: This file probably requires the boilerplate header with the licensing etc

will-moore (approved):
This was included in the merge-ci deployment today - tested all the various rendering methods in viewers, thumbnails, plate JSON etc all working fine. Copy, Paste and Saving of rendering settings also tested. 👍

PR #554, opened 2024-05-14
Optimize obj_id_bitmask by using its own query

chris-allan:
Having the column oriented responses from OMERO.tables transposed into rows is quite the performance hit for the bitmask endpoint.  As the endpoint also works on one column the batching and various other semantics of perform_table_query() do not apply and add needless overhead.

Consequently, the implementation has been rewritten to use its own query and the bit packing has been further optimized based on new input.

/cc @erindiel, @sbesson, @kkoz, @knabar


omeroweb/webgateway/views.py
```diff
@@ -3192,42 +3192,41 @@ def obj_id_bitmask(request, fileid, conn=None, query=None, **kwargs):
...
+
     try:
-        data = rowsToByteArray(rsp_data["data"]["rows"])
-        return HttpResponse(bytes(data), content_type="application/octet-stream")
+        column_names = [column.name for column in table.getHeaders()]
+        if col_name not in column_names:
+            return dict(error="Unknown column %s" % col_name)
+        row_numbers = table.getWhereList(query, None, 0, 0, 1)
+        (column,) = table.slice([column_names.index(col_name)], row_numbers).columns
+        return HttpResponse(
```
knabar, line 3208: For clarity, I'd only wrap lines 3208-3210 in the `try`..`except`, since the preceding lines can't raise a `ValueError` (or if they do, the error message returned is misleading)

will-moore (approved):
Code changes look good.
Removal of `def rowsToByteArray()` could be considered a breaking API change, although not likely to being used externally.

Bitmask table tests are still passing at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/94/testReport/OmeroWeb.test.integration.test_table/TestOmeroTables/

chris-allan:
To me the undocumented previous function was internal API. Happy to keep it around and mark it deprecated if you think that's justified, @will-moore.

will-moore:
No, I think it's fine. Just flagging it in case anyone else disagreed. Thx

knabar (approved):

kkoz (approved):

will-moore:
I'm afraid we have 2 failing tests again - see https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/99/testReport/OmeroWeb.test.integration.test_table/TestOmeroTables/test_table_bitmask_query_result0_/history/

chris-allan:
51d8ba745b5f5ec248f29c258a7d14150017e5ce should fix the integration test failures; the previous implementation was kind of weird and the behaviour we're now mimicing was undocumented but it's been like that for several years and the integration tests depend on it. We can decide in the future if we want to make more drastic changes and properly fail if a column is unknown.

79c23056136bc929e4ebd57d24600780c92921c5 is a little more sinister as the previous implementation would have had weird behaviour around negative values. We aren't specifically testing for this sort of strange data layout but the implementation now mimics what was there previously and again we can decide in the future if we want to explicitly disallow negative indexes.

knabar:
@chris-allan Thank you for tracking down these issues.

Once the new API endpoints are available, would we consider deprecating the bitmask call altogether?

chris-allan:
> Once the new API endpoints are available, would we consider deprecating the bitmask call altogether?

Potentially, yes. Once they're in place I would anticipate very similar performance characteristics to bitmask with much more appropriate functionality.

sbesson (approved):

chris-allan:
Test cases now passing:

 * https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/110/

PR #555, opened 2024-05-15
Archive status

chris-allan:
Adds archive status to the tree metadata, visually in the tree and to image data JSON.

image


jburel:
This PR leads to some test failures see https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/64/

will-moore:
Tested on merge ci with
```$ omero obj update Image:1673 archived=True```
And this displays correctly in webclient as shown above.

My only question is whether users need some clue as to what the red dot means. I guess this might not be necessary if all users have been trained accordingly, (which may well happen in places that are using archiving, since users should know to expect different behaviour from those images?), but it seems that there could be some who are confused by the red dot.
Possibly a tooltip on the red-dot icon (might be a bit painful digging into how to do this in jstree) or in the right panel `core_metadata.html` add a row `Archived: true` (only if True - don't need to add `Archived: false` for all other images)
or something else?

chris-allan:
> My only question is whether users need some clue as to what the red dot means. I guess this might not be necessary if all users have been trained accordingly, (which may well happen in places that are using archiving, since users should know to expect different behaviour from those images?), but it seems that there could be some who are confused by the red dot.
Possibly a tooltip on the red-dot icon (might be a bit painful digging into how to do this in jstree) or in the right panel core_metadata.html add a row Archived: true (only if True - don't need to add Archived: false for all other images)
or something else?

Likely. I only used `image_locked16.png` because it was available. I expect the red-dot is not particularly informative and this deserves a dedicated icon and a tooltip. Can also add something to `core_metadata.html` as a demonstration of what's possible.

chris-allan:
Added right-hand panel visual:

image

chris-allan:
Added a tooltip to the tree (visuals are from Windows and will look slightly different depending on operating system):

image

lunson:
Options for iconography
icon_archived

and the image files for each of the options:
[icons_archive.zip](https://github.com/user-attachments/files/15535584/icons_archive.zip)

chris-allan:
Display of archive status in the tag tree:

image

will-moore:
Tests passing today 👍 https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/82/testReport/

knabar:
New icon looks good:
Screenshot 2024-07-29 at 10 47 18

PR #561, opened 2024-06-17
cap numpy to less than 2

jburel:
See https://github.com/ome/omero-py/pull/414


chris-allan (approved):

sbesson (approved):
Proposing to get in asap so that we can unblock other work /cc @chris-allan @knabar

chris-allan:
Was what was required to get ome/omero-web#554 builds working again.

PR #568, opened 2024-07-19
Perceptibly uniform luts for OMERO.web

Tom-TBT:
This PR adds the preview of LUTs introduced in https://github.com/ome/openmicroscopy/pull/6398

The preview display for other OMERO components depends on this PR:
* https://github.com/ome/omero-figure/pull/579
* https://github.com/ome/omero-iviewer/pull/481


sbesson:
As mentioned yesterday at the weekly OME meeting, one question is whether we can reduce the number of coupled PRs required to add new LUTs to the infrastructure in the future.

In the OMERO.web ecosystem, it's not clear to me why the png image as well as the list of LUTs needs to be updated for every downstream application rather than exposing them as APIs.

In terms of testing, several combinations probably will need to be tested including OMERO.server with/without https://github.com/ome/openmicroscopy/pull/6398 against OMERO.web with/without this PR to ensure the client responds accordingly under all scenarios. I assume the new LUTs tables should only appear in the drop-down menu if available server-side and defined in OMERO.web. In all other scenarios, the previous set of LUTs should be displayed and no exception sent to the user.

jburel:
It should not be necessary to upgrade the various apps when adding a new LUT
The LUTs are considered as "script" with a specific mimetype
Java clients retrieve the LUT using method like https://github.com/ome/omero-gateway-java/blob/master/src/main/java/omero/gateway/facility/BrowseFacility.java#L1614

Tom-TBT:
The whole point of the PR for the 4 apps (iviewer, figure, insight and web) is the preview of the LUT during selection. Otherwise, you get a blank space next to the LUT name. 

So I think the question is how to load that LUT preview in OMERO.web and OMERO.insight from the server. Then like you say @sbesson, we can expose from OMERO.web API the previews for the downstream applications.

After such change, any LUT added in OMERO.server will show up nicely in all apps without needing an additional update.

will-moore:
@Tom-TBT did you write a new script to create your `luts_10.png` or did you find an existing script? I can't remember if we documented that anywhere?

Obviously it would be nice if a single change on the server would automatically update the LUTs in various clients. It's just a question of how much work it is to add the API to the server compared with the occasional update to various clients?

knabar:
Can this PR be closed?

Tom-TBT:
These changes are substituted with a dynamic version of LUT png generation:
https://github.com/ome/omero-web/pull/581

PR #574, opened 2024-08-22
Fix word wrap for names

### LINKED ISSUE #572: Display of long names - word wrap
Similar to issue with file paths:
https://github.com/ome/omero-web/issues/437

Reported at https://forum.image.sc/t/image-names-not-wrapping/100493

Seems like ``word-break: break-word;` should work
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

will-moore:
Fixes #572 
See report at https://forum.image.sc/t/image-names-not-wrapping/100493

To test, set various long names without spaces.
These should now wrap correctly instead of disappearing off the screen.

### DIFF
--- a/omeroweb/webclient/static/webclient/css/layout.css
+++ b/omeroweb/webclient/static/webclient/css/layout.css
@@ -672,6 +672,7 @@
 			width:85%;
 			display:inline-block;
 			padding:5px 0;
+            word-wrap: break-word;
 	    }
     
 	        .data_heading span {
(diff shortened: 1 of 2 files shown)


pwalczysko:
Yes, working fine on merge-ci.

Screenshot 2024-08-30 at 13 38 23
Screenshot 2024-08-30 at 13 37 31


the interesting feature is that a dash ``-`` character seems to be preferred for inserting the break. I think this is acceptable (the wrapping happens even if no dash is present in name). Just a heads up.

pwalczysko (approved):

PR #576, opened 2024-08-27
webclient page scaleable, login page device-width

will-moore:
I noticed that the webclient login page is really not very mobile-phone friendly as the whole page shrinks to be quite small.

At the same time, the main webclient page is completely unusable on a phone as it is NOT scaleable and you can't access the centre & right areas of the page.
This is because it has the `user-scalable=no` flag on the `<viewport>`
which is "against Web Content Accessibility Guidelines" see https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag

Using Chrome dev tools to emulate a phone E.g. a `Google Pixel 4`.

Before this PR: login page is tiny and webclient page is unusable:

Screenshot 2024-08-27 at 20 42 12

Screenshot 2024-08-27 at 20 42 46

After these changes, the login page is a useful size (but still scalable) and the webclient page is usable as you can zoom (rescale) and pan around the page to access all functionality:

Screenshot 2024-08-27 at 20 45 30

Screenshot 2024-08-27 at 20 44 24

Screenshot 2024-08-27 at 20 44 55

### DIFF
--- a/omeroweb/webclient/templates/webclient/login.html
+++ b/omeroweb/webclient/templates/webclient/login.html
@@ -68,7 +68,8 @@
 
 {% block head %}
     {{ block.super }}
-	
+	<meta name="viewport" content="width=device-width, initial-scale=1" />
+
 	<script type="text/javascript" charset="utf-8">
 			$(function(){ $("label").inFieldLabels(); });
 		</script>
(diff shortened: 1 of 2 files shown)


knabar (approved):
Looks good

PR #577, opened 2024-08-27
Refactors #555 to allow for backward compatibility

chris-allan:
There were certain downstream omero-web plugins which were using `_marshal_image()`.  This refactors the changes of #555 such that the prototype of `_marshal_image()` is maintained and a new `_marshal_image_map()` is added to give us options going forward for scalability.

Alternative to #575.

/cc @Tom-TBT

### DIFF
--- a/omeroweb/webclient/tree.py
+++ b/omeroweb/webclient/tree.py
@@ -531,7 +531,6 @@ def _marshal_image(
     """Given an Image row (list) marshals it into a dictionary.  Order
     and type of columns in row is:
       * id (rlong)
-      * archived (boolean)
       * name (rstring)
       * details.owner.id (rlong)
       * details.permissions (dict)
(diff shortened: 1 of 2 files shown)


chris-allan:
I've opened this as a draft so we don't end up with build conflicts with #575 and can use this for discussion.

omeroweb/webclient/tree.py
```diff
@@ -1536,23 +1576,12 @@ def marshal_tagged(
...
-        e = unwrap(e)
-        row = [
-            e[0]["id"],
-            e[0]["archived"],
-            e[0]["name"],
-            e[0]["ownerId"],
-            e[0]["image_details_permissions"],
-            e[0]["filesetId"],
-        ]
+        data = unwrap(e)
```
sbesson, line 1579: `data = unwrap(e)[0]`

omeroweb/webclient/tree.py
```diff
@@ -1536,23 +1576,12 @@ def marshal_tagged(
...
-        if load_pixels:
-            d = [e[0]["sizeX"], e[0]["sizeY"], e[0]["sizeZ"], e[0]["sizeT"]]
-            kwargs["row_pixels"] = d
         if date:
-            kwargs["acqDate"] = e[0]["acqDate"]
-            kwargs["date"] = e[0]["date"]
-        images.append(_marshal_image(conn, row, **kwargs))
+            kwargs["acqDate"] = data["acqDate"]
+            kwargs["date"] = data["date"]
+        images.append(_marshal_image(conn, data, **kwargs))
```
sbesson, line 1584: Same as above, should this use the new `_marshal_image_map` API?

sbesson (commented):
A couple of comments. From an API construction, I think this proposal gets us into a good place:
- a new `_marshal_image_map` is introduced that takes a dictionary rather than an array of fixed length. This means new properties can be introduced in a backwards-compatible way in the future
- the signature of the existing `_marshal_image` is restored and its implementation is forwarded to the new `_marshal_image_map`. This should fix the behavior of existing downstream apps that were using this internal method
- I assume the high-level `marshal_images` and `marshal_tagged` API should use the `_marshal_image_map` API moving forward?

In terms of testing, I expect the changes introduced in https://github.com/ome/openmicroscopy/pull/6389 should be unchanged as the `omeroweb.webclient.tree.marshal_images` API should create a dictionary including the `archived` key.
Should the legacy `_marshal_image` API be tested and/or deprecated?

omeroweb/webclient/tree.py
```diff
@@ -550,22 +549,74 @@ def _marshal_image(
...
+    @type row L{dict}
+    """
     image = dict()
-    image["id"] = unwrap(image_id)
-    image["archived"] = unwrap(archived) is True
-    image["name"] = unwrap_to_str(name)
-    image["ownerId"] = unwrap(owner_id)
-    image["permsCss"] = parse_permissions_css(permissions, unwrap(owner_id), conn)
-    fileset_id_val = unwrap(fileset_id)
+    image["id"] = unwrap(data["imageId"])
```
sbesson, line 602:
```suggestion
    image["id"] = unwrap(data["id"])
```

sbesson:
With OMERO.web 5.27.0 and OMERO.autotag 4.0.1, I can reproduce the original issue as toggling the `Auto-tag` from the centre panel triggers a 500 error.

I deployed a local build of this PR with the additional change proposed in https://github.com/ome/omero-web/pull/577#discussion_r1734342465, I can confirm that the auto-tag server issue is fixed. Setting the `archived` flag to true via the CLI also produces the expected behavior as per #555 i.e. an archived icon and `Archived: True` in the right-hand panel.

I would be in favor of closing #575 and moving this out of draft for inclusion in the nightly CI builds in preparation of a round of functional testing.

knabar:
Do we also want to address the issue of the "underscore" method that indicates it is private, since we know it is not used as such?

For example, remove the leading underscore from the methods, but keep the underscore reference for compatibility:

```
def marshal_image(...):
    ...

_marshal_image = marshal_image

def marshal_image_map(...):
    ...
```

This is admittedly a bit redundant, but might deter future signature changes.

chris-allan:
> Should the legacy _marshal_image API be tested and/or deprecated?

I think at least tested. Once the current integration tests are passing I'll add some more unit tests to this project on this PR.

chris-allan:
> Do we also want to address the issue of the "underscore" method that indicates it is private, since we know it is not used as such?
> 
> For example, remove the leading underscore from the methods, but keep the underscore reference for compatibility:
> 
> ```
> def marshal_image(...):
>     ...
> 
> _marshal_image = marshal_image
> 
> def marshal_image_map(...):
>     ...
> ```
> 
> This is admittedly a bit redundant, but might deter future signature changes.

I don't think so. Or at least not here. What was clear from #575, especially ome/omero-mapr#84, is that almost every leading underscore marshal method from the tree package is used in that plugin and has been for nearly a decade. The whole module needs review and an assessment of what we we want to do with it as there are clearly critical plugins that are tightly coupled to nearly its entire API.

will-moore (approved):
Changes look good and make sense, preserving previous API. Auto-tag working locally and the archived status is shown in webclient correctly.

chris-allan:
Unit tests have been added to cover both `_marshal_image()` and `_marshal_image_map()`.

omeroweb/webclient/tree.py
```diff
@@ -753,31 +804,20 @@ def marshal_images(
...
-            kwargs["date"] = e["date"]
+            kwargs["acqDate"] = data["acqDate"]
+            kwargs["date"] = data["date"]
 
         # While marshalling the images, determine if there are any
         # images mentioned in shares that are not in the results
         # because they have been deleted
-        if share_id is not None and image_rids and e["id"] in image_rids:
+        if share_id is not None and image_rids and data["id"] in image_rids:
             image_rids.remove(e["id"])
```
sbesson, line 817:
```suggestion
            image_rids.remove(data["id"])
```

should fix the failing integration test - https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/158/testReport/OmeroWeb.test.integration.test_tree/TestTree/test_marshal_images_share/

sbesson (approved):
Retested the last commit in a deployment with the current versions of OMERO.autotag and OMERO.mapr. Confirmed that the archived status is still displayed as expected in the webclient (left and right hand panels) and both applications work as expected.

The nightly CI integration tests have been passing and the new collection of unit tests add coverage for all supported arguments of the internal `_marshal_image` and `_marshal_image_map` APIs.

Proposing to get this released asap to reduce the impact on people who upgraded their version of OMERO.web in a deployment with any of the impacted applications. Given these APIs are marked as internal, either 5.27.1 or 5.28.0 would be acceptable version increment from my side

PR #583, opened 2024-09-26
Allow users to click a group name to quick-switch to that group

DavidStirling:
This PR slightly tweaks this group selection menu:

Screenshot 2024-09-26 at 10 55 44

The current workflow for selecting a group seems to be to open the menu, hover the group name and then navigate to the right submenu to choose which user to show data for.

In practice this seems a bit clunky, particularly when using a touchpad. It's easy to accidentally touch another group when trying to move the cursor to the right submenu.

This PR makes it so that clicking on the group name in the left menu behaves as if the user clicked "All Members" on the right submenu. I'd hope that this shortcut makes switching group a little faster and easier.

Based on the current code I wonder if at some point clicking the group name was meant to pin the right menu open and this is simply broken? I don't know the history here.

One question is how to handle private groups which don't normally display the "All Members" option. When tested in practice it seems that having `experimenter=-1` in the URL for these groups doesn't actually change what the user sees in the resulting tree, so I hope that's fine.

### DIFF
--- a/omeroweb/webclient/templates/webclient/base/includes/group_user_content.html
+++ b/omeroweb/webclient/templates/webclient/base/includes/group_user_content.html
@@ -28,7 +28,11 @@
 
 {% for grp in myGroups %}
     <li {% if grp.id == eContext.context.groupId %}class="current_group"{% endif %}>
-        <a href="#" data-gid="{{ grp.id }}">
+        {% if grp.id == active_group.id %}
+            <a href="{{ current_url }}?experimenter=-1" data-gid="{{ grp.id }}">
+        {% else %}
+            <a href="{% url 'change_active_group' %}?active_group={{grp.id}}&url={{ current_url }}?experimenter=-1" data-gid="{{ grp.id }}">
+        {% endif %}
             <img
             {% if grp.getDetails.getPermissions.isGroupWrite %} src="{% static 'webclient/image/group_green16.png' %}"
             {% else %}
--- a/omeroweb/webclient/templates/webclient/base/includes/group_user_content.html
+++ b/omeroweb/webclient/templates/webclient/base/includes/group_user_content.html
@@ -28,10 +28,10 @@
 
 {% for grp in myGroups %}
     <li {% if grp.id == eContext.context.groupId %}class="current_group"{% endif %}>
-        {% if grp.id == active_group.id %}
-            <a href="{{ current_url }}?experimenter=-1" data-gid="{{ grp.id }}">
+        {% if grp.id == ome.active_group %}
+            <a href="#" data-gid="{{ grp.id }}">
         {% else %}
-            <a href="{% url 'change_active_group' %}?active_group={{grp.id}}&url={{ current_url }}?experimenter=-1" data-gid="{{ grp.id }}">
+            <a href="{% url 'change_active_group' %}?active_group={{grp.id}}" data-gid="{{ grp.id }}">
         {% endif %}
             <img
             {% if grp.getDetails.getPermissions.isGroupWrite %} src="{% static 'webclient/image/group_green16.png' %}"
@@ -54,7 +54,7 @@
 
     {% for grp in groups %}
         <li {% if grp.id == eContext.context.groupId %}class="current_group"{% endif %}>
-            <a href="#" data-gid="{{ grp.id }}">
+            <a href="{% url 'change_active_group' %}?active_group={{grp.id}}&url={{ current_url }}?experimenter=-1" data-gid="{{ grp.id }}">
                 <img
                 {% if grp.getDetails.getPermissions.isGroupWrite %} src="{% static 'webclient/image/group_green16.png' %}"
                 {% else %}
--- a/omeroweb/webclient/templates/webclient/base/includes/group_user_dropdown2.html
+++ b/omeroweb/webclient/templates/webclient/base/includes/group_user_dropdown2.html
@@ -53,11 +53,9 @@
 
                         // When hover or click on a Group, show Users...
                         $( "#groupList" ).on( "mouseenter click", "li a", function(event) {
-                            event.preventDefault();
                             var gid = $(this).attr("data-gid");
                             $usersList.hide();
                             $("#groupMembers-" + gid).show();
-                            return false;
                         });
                     });
                 }


knabar:
I'm also not aware of the history of this menu, but noticed that the context variables used (`eContext.context.groupId` and `active_group.id`) seem to be unset.  The latter now seems to be `ome.active_group`, so there may have been some incomplete refactoring at some point?

Right now, it seems to be that only the `{% else %}` branch of this change will ever run.

DavidStirling:
The condition statement used here was based on the function for generating URLs in the submenu. It looks like you're right and that the `active_group=<id>` param is included even if that's the user's current group. This also applies to the current release.

will-moore:
I played with this a bit and I'm not sure that I like the switch to "All Members" by default. For most users wanting to switch groups, they are likely wanting to see their own data in a different group.
I tried removing the `experimenter=-1` for the group link to be:
```
<a href="{% url 'change_active_group' %}?active_group={{grp.id}}" data-gid="{{ grp.id }}">
```
This is now quite nice as it switches group without changing the user.
So, if I'm looking at my data in Group 1 and I switch to Group 2, I see my data there.
If I'm looking at "John's" data in Group 1 and I switch to Group 2, I am still looking at "John's" data there.
If John isn't a member of Group 2, then it automatically defaults to showing my data there.

It's always reasonably easy to find the "All Members" option in the menu when you're switching group, but it's not always easy to find yourself or another user that you may have just selected. So, keeping the current user when switching is much nicer.

It seems like we're not missing any nice functionality by `{% if grp.id == active_group.id %}` failing everywhere (it just avoids a silent "change_active_group" step) so it might be nice to simply clean up and remove all those if/else blocks?
Otherwise we could use `{% if grp.id == ome.active_group %}` in those places.
However, for the group link in this PR, I don't know that I'd want the current group in the menu to reload the page with "All Members" (kinda unexpected behaviour). Better just to have it do nothing with `href="#"`.

sbesson:
> For most users wanting to switch groups, they are likely wanting to see their own data in a different group.

I am not sure I agree with this statement. This will be highly dependent on the type of user, how the groups are set-up and how the data is managed.

> It's always reasonably easy to find the "All Members" option in the menu when you're switching group, but it's not always easy to find yourself or another user that you may have just selected. 

Absolutely, especially as the number of users in a given group grows, this can be particularly cumbersome.

If there a way to use the current user context in the proposed functionality to address both scenarios i.e.:
- if `All Members` is selected in the current group, also select `All Members` when clicking on a new group
- if any other user is selected in the current group, try and select the same user when clicking on a new group and fallback on the authenticated user

will-moore:
@sbesson That is already the behaviour you'll get with my proposed change. Clicking on a group without choosing a User simply switches the group without switching the user. If you're already viewing "All Members" then you will still be seeing "All Members" in the new group. You don't need to know the current user for this.

pwalczysko:
> It's easy to accidentally touch another group when trying to move the cursor to the right submenu.

Sorry, are the proposed changes really solving the problem first highlighted by @DavidStirling ? I fully agree that this is a problem, i.e. the danger is in choosing a wrong group, happens many times during the trainings. So when I click onto the groupname in top left, then I have to go through two submenus - the group submenu and the users-in-that-group submenu.
This is often a flop, as you are prone to `accidentally touch another group ` as @DavidStirling puts it - indeed, you do not need a trackpad, even with a superior mouse you will still make the error.

Then what you @will-moore propose is to let me just click on the group name in the first submenu, and not to have to navigate to the second submenu ? Which would have then as a result to navigate to the correct user in another group (where correct==the same as in the present group). 
But, in praxis, at least in my hands, @DavidStirling 's workflow usually attempts to "stay in the same group, but select another user in that group". Hence the suggested solution will not wfm (not claiming that this is a blocker for this PR or your helpful suggestion @will-moore ). 
Would you have some improvement for my case too please @will-moore  ? ^^^

pwalczysko:
Maybe to make my point clearer, I actually also dispute the relevance of

> For most users wanting to switch groups, they are likely wanting to see their own data in a different group.

The problem here is that the menu in its present state is a gate for both users who want to switch groups, as well as the ones who "just want to switch to another user in the same group".  As mine is the latter case, the solutions proposed by now are not doing it for me I am afraid. I was always told that the problem is because I have an ``edge case`` of having both users A, B in both groups 1, 2 and hence attempted switch to (starting from 1A ) to  1B often ends at 2B.  After some years with production systems, I would dispute the "edge-caseness" of my case :)

jburel:
Discussed today
Use cases to check:
* If a new group is selected (``group-2``) and the selected user (``user-1``) in the previous group (``group-1``) belongs to the new group then select that user i.e. ``user-1 in group-2``
*  If new group selected (``group-2``) and the selected user (``user-1``) in previous group (group-1) **does not** belong to the new group then select the logged in user. If the logged in user is **not** a member of the new group e.g. admin, default to ``All members``
* * If a new group is selected (``group-2``) and ``All members` is selected in the previous group (``group-1``)  then select ``All members``  i.e. ````All members` in group-2``

will-moore:
To address the accidental selection of adjacent groups while trying to select users, we could try using a jQuery plugin designed to fix this: https://github.com/kamens/jQuery-menu-aim/tree/master

will-moore:
The use-cases I checked above were with these changes:
NB: this includes changes to the Admin's full list of groups to handle the case where you switch to a group that you're not a member of (which wasn't part of the original PR).
In addition:

 - [x] if you choose one of your own groups that you're already viewing, (`grp.id == active_group.id`) then nothing happens (no group or user switch).

```
--- a/omeroweb/webclient/templates/webclient/base/includes/group_user_content.html
+++ b/omeroweb/webclient/templates/webclient/base/includes/group_user_content.html
@@ -28,10 +28,10 @@
 
 {% for grp in myGroups %}
     <li {% if grp.id == eContext.context.groupId %}class="current_group"{% endif %}>
-        {% if grp.id == active_group.id %}
-            <a href="{{ current_url }}?experimenter=-1" data-gid="{{ grp.id }}">
+        {% if grp.id == ome.active_group %}
+            <a href="#" data-gid="{{ grp.id }}">
         {% else %}
-            <a href="{% url 'change_active_group' %}?active_group={{grp.id}}&url={{ current_url }}?experimenter=-1" data-gid="{{ grp.id }}">
+            <a href="{% url 'change_active_group' %}?active_group={{grp.id}}" data-gid="{{ grp.id }}">
         {% endif %}
             <img
             {% if grp.getDetails.getPermissions.isGroupWrite %} src="{% static 'webclient/image/group_green16.png' %}"
@@ -44,7 +44,7 @@
                     {% endif %}
..

@@ -54,7 +54,7 @@
 
     {% for grp in groups %}
         <li {% if grp.id == eContext.context.groupId %}class="current_group"{% endif %}>
-            <a href="#" data-gid="{{ grp.id }}">
+            <a href="{% url 'change_active_group' %}?active_group={{grp.id}}&url={{ current_url }}?experimenter=-1" data-gid="{{ grp.id }}">
                 <img
                 {% if grp.getDetails.getPermissions.isGroupWrite %} src="{% static 'webclient/image/group_green16.png' %}"
                 {% else %}
```

pwalczysko:
> To address the accidental selection of adjacent groups while trying to select users, we could try using a jQuery plugin designed to fix this: https://github.com/kamens/jQuery-menu-aim/tree/master

Tested this. I think this would do the job, I also like the monkey example under ``example/example.html``. Not sure about the repo last conrib 11 years ago though.

knabar:
Skipping this for 5.28.0 since we want to make some further changes.

DavidStirling:
Sounds fair, do we have a consensus on what we'd like the final behaviour to be? I'd lean towards not setting experimenter at all so that we carry over whatever the user last selected.

pwalczysko:
> I'd lean towards not setting experimenter at all so that we carry over whatever the user last selected.

Can you maybe expand on the usecase ? This sounds to me like (supposing a switch between group-1 and group-2 should happen): 

- user-1 logs in into group-1
- user-1 views user-2's data in group-1
- user-1 wants to switch to viewing user-2's data in group-2 in one click

This ^^^ does not seem like something which user-1 wants to do very often. Maybe there is a usecase with All Members which I am missing ? 
The behaviour suggested by @will-moore is rather that the one-click action should always switch the group and get user-1 to their own data in group-2.

DavidStirling:
> This ^^^ does not seem like something which user-1 wants to do very often. Maybe there is a usecase with All Members which I am missing ? The behaviour suggested by @will-moore is rather that the one-click action should always switch the group and get user-1 to their own data in group-2.

I would interpret the shortcut as not being opinionated about user selection that would otherwise be made in the right panel, so switching groups with it doesn't try to specify the active user at all.

I imagine that if user 1 has specifically "asked" to see user 2's data in Group 1, they might expect to still be looking at user 2's data if switching to Group 2 using the shortcut. Otherwise they could use the non-shortcut selection to see a different user's data. This is especially true if they've selected "all members" rather than a specific user.

I may be misunderstanding the suggestions above, so please do correct me if I'm wrong about how that'd work.

pwalczysko:
> so switching groups with it doesn't try to specify the active user at all.

And thus if user-2 was selected, it will be selected in group-2 as well, do I get it right ?

> Otherwise they could use the non-shortcut selection to see a different user's data. This is especially true if they've selected "all members" rather than a specific user.

That makes sense to me. What do you think @will-moore ?

> To expand on the "accidentally hitting the wrong group" issue - I don't see a simple way of preventing this entirely,

Coming back to this comment ^^^ - I do not dispute usefulness of the PR, it is just that my usecase is a bit different - if we cannot prevent this entirely, then lets just mitigate it - the mitigation suggested in this PR is not addressing my usecase. Happy if it is done in another PR, but before that, we just want to see the big picture I believe.

DavidStirling:
Applied, thanks!

will-moore (approved):
Testing the checklist of behaviours above with this PR now deployed on merge-ci. All working as expected. 
LTGT 👍

pwalczysko (approved):

PR #588, opened 2024-10-16
rf(py314): Replace deprecated pkgutil.find_loader

effigies:
This PR removes [pkgutil.find_loader()][] and replaces it with [importlib.util.find_spec()][]. `find_loader` was deprecated in Python 3.12 and will be removed in 3.14. `find_spec` has been present since Python 3.4.

Both functions return `None` if the module loader cannot be found. For its use in this project, this is sufficient and no translation of the return value is needed.

[pkgutil.find_loader()]: https://docs.python.org/3/library/pkgutil.html#pkgutil.get_loader
[importlib.util.find_spec()]: https://docs.python.org/3/library/importlib.html#importlib.util.find_spec

### DIFF
--- a/omeroweb/urls.py
+++ b/omeroweb/urls.py
@@ -24,7 +24,7 @@
 #
 
 import logging
-import pkgutil
+import importlib.util
 from django.conf import settings
 from django.apps import AppConfig
 from django.conf.urls import include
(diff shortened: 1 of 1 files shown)


will-moore:
Hi @effigies, thanks for the PR.
Before we can accept a contribution to any of our GPL repos, we need a CLA form from you as described at https://ome-contributing.readthedocs.io/en/latest/cla.html 
Many thanks!
 Will

effigies:
I hereby release the contribution into the public domain. You may include it as-is without attribution, or make your own change with similar effect, to avoid counting me as a contributor with any kind of copyright claim.

I might get around to signing that if I make a real contribution, but I do not believe any creative effort was involved that would justify a copyright claim, let alone license.

Feel free to close this if the above is insufficient.

knabar:
Closing in favor of https://github.com/ome/omero-web/pull/589, which is based on a revert.

PR #594, opened 2024-11-06
Add minimum batch size to avoid infinite loop due to zero batch size

### LINKED ISSUE #593: Infinite loop when downloading tables with <10 rows.
**Bug**

Trying to download an omero table with < 10 rows with the "Whole Table" download button (top right) causes an infinite loop and a wrong and always increasing download size (here `3MB`  for a real total size of ~ `79.59 KB` ) to be displayed:
image

The "Show current page as: CSV | JSON" links seem be work correctly.

**To Reproduce**

Open any table with less than 10 rows in omero web and click the "Whole Table" download button (top right).
image

Quick examples at:
- https://idr.openmicroscopy.org/webclient/omero_table/39275220/ (1 row)
- https://idr.openmicroscopy.org/webclient/omero_table/34961908/ (5 rows)

With  rows >= 10 everything works:
- https://idr.openmicroscopy.org/webclient/omero_table/15349581/ (10 rows)
- https://idr.openmicroscopy.org/webclient/omero_table/43413223/ (27 rows)

**Cause and fix?**

In [omero_table.html](https://github.com/ome/omero-web/blob/master/omeroweb/webclient/templates/webclient/annotations/omero_table.html) in the `download_csv` function in this [line](
https://github.com/ome/omero-web/blob/213bb61adc23aed4a750f2a83dc94b7efdb22cee/omeroweb/webclient/templates/webclient/annotations/omero_table.html#L219C36-L219C57) the `batchSize`  evaluates to `0` for  values of `rowCount` < 10.
Then, when `count` is incremented [here](https://github.com/ome/omero-web/blob/213bb61adc23aed4a750f2a83dc94b7efdb22cee/omeroweb/webclient/templates/webclient/annotations/omero_table.html#L255), since `batchSize`  is  `0` the `count` remains `0` causing the infinite loop.

I think this could be solved by forcing the `batchSize` to be at least `1` e.g. with:
`const batchSize = Math.min(Math.max(parseInt(rowCount/10), 1), MAX_BATCH_ROWS);`
If it is OK, I could open a PR with this change.

knabar:
Fixes #593 

Avoids infinite loops for tables with fewer than 10 rows (resulting in a zero batch size).

### DIFF
--- a/omeroweb/webclient/templates/webclient/annotations/omero_table.html
+++ b/omeroweb/webclient/templates/webclient/annotations/omero_table.html
@@ -213,10 +213,11 @@
         enableDownloadButtons(false);
         const rowCount = filter ? parseInt("{{ meta.totalCount }}") : parseInt("{{ meta.rowCount }}");
         const tableName = "{{ data.name }}.csv";
-        // Use 10 batches, or max 3000 rows per batch
+        // Use 10 batches, or max 3000 rows per batch, with a minimum of 10
         const MAX_BATCH_ROWS = 3000;
+        const MIN_BATCH_ROWS = 10;
         const MAX_FAILED_CALLS = 5;
-        const batchSize = Math.min(parseInt(rowCount/10), MAX_BATCH_ROWS);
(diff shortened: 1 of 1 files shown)


will-moore (approved):
Simple fix. Code looks good. 👍

PR #595, opened 2024-11-12
linkify supports urls that include {}[] characters

sukunis:
Test by adding URL:
https://openbis-eln-lims.ethz.ch/openbis/webapp/eln-lims/?menuUniqueId={%22type%22:%22SAMPLE%22,%22id%22:%2220150127162826183-170%22}&viewName=showViewSamplePageFromPermId&viewData=%2220150127162826183-170%22

grafik

### DIFF
--- a/omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
+++ b/omeroweb/webclient/static/webclient/javascript/ome.webclient.actions.js
@@ -61,7 +61,7 @@ OME.getURLParameter = function(key) {
 };
 
 var linkify = function(input) {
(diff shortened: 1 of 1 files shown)


knabar:
According to https://www.ietf.org/rfc/rfc3986.txt, square brackets are only allowed for the host part (to deal with IPv4/v6 addresses) and must be encoded otherwise. Curly brackets are considered unsafe and should be encoded.

Obviously these rules are not followed in common usage, but we should be sure not to introduce any potential issues by automatically linkifying those URIs.

PR #596, opened 2024-11-12
Linkify can parse anchor tag

sukunis:
To shorten links that are added in comments: linkify accepts anchor tag input. 

Tested with <a href="https://www.example.com">test link</a>

grafik

Result:
grafik


knabar:
At first glance this would allow linkifying `<a>` tags with `javascript:` hrefs and possibly other undesired links.  At a minimum the `href` property would have to be checked against the usual linkify regex, but that might not be sufficient.

PR #597, opened 2024-11-14
Create release

jburel:
This PR adds an action that will create a release when a tag is pushed
This will allow to check for upgrade in QA.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,23 @@
+---
+
+name: Create Release
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+jobs:
+  build:
+    name: Create Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Create Release
+        id: create_release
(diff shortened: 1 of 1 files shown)


jburel:
The "hit" endpoint checks if an upgrade is required. This only works for OMERO.server
I have adjusted the code so it also works for omero-web, insight
For that I checked what is the latest value available from  GitHub using ``https://github.com/ome/omero-web/releases/latest``

sbesson:
Thanks, you might be able to use the REST API directly as in https://api.github.com/repos/ome/omero-web/releases/latest

I'll leave @knabar to evaluate as he's going to be primarily involved here but I am not convinced that adding yet another GitHub workflow is justified here. Instead we could just update the release process and manually create a release after pushing the tag, this takes a few clicks and <30s using the current UI. Another potential caveat of the automated workflow is that we might not necessarily want to create GitHub releases for every tag and notably for release candidates if this API is going to be consumed by the OME upgrade check logic.

jburel:
openmicroscopy and insight create a release. 
That's why I picked that approach so I had the same piece of code to find the information
but if we don't want to create a release I will have to modify the code again i.e.
```
    r = requests.get(url)
    ver = regex.match(r.url).group(1)
```

jburel:
I have now modified the code to use the REST API i.e. https://api.github.com/repos/ome/omero-web/tags
I will delete the release that I created manually

PR #599, opened 2024-11-20
Update PyPI publish workflow to use build instead of sdist

knabar:
`sdist` produces a package name containing `omero-web` instead of `omero_web`, triggering a deprecation warning (see [PEP 625](https://peps.python.org/pep-0625/)).

Other packages have been updated similarly: [omero-py](https://github.com/ome/omero-py/pull/406), [omero-scripts](https://github.com/ome/omero-scripts/pull/220)

### DIFF
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -13,8 +13,8 @@ jobs:
           python-version: '3.9'
       - name: Build a binary wheel and a source tarball
         run: |
-          python -mpip install wheel
-          python setup.py sdist bdist_wheel
+          python -mpip install build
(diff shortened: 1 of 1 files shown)


jburel (approved):

sbesson (approved):
This is consistent with what's being applied to `omero-py` and `omero-scripts`.

@jburel I assume other Python projects will need to be adjusted as they are getting released to newer versions of PyPI - https://github.com/search?q=org%3Aome+sdist+language%3Ayaml&type=code

jburel:
Yes so far I have been updating them while doing a new release
but I need to go over the remaining one that an email is sent

PR #603, opened 2025-01-07
Don't use group -1 for luts_png

will-moore:
Issues that need addressing:
 - We can't rely on cache being enabled for omero-web
 - Dynamic generation of all LUTs takes a while and loads a lot of files from the server

Proposed solution:
 - We use the a new file `webgateway/static/json/luts.json` as the source for LUTs to build the dynamic `/webgateway/luts_png/` (the `/webgateway/luts_png/` still corresponds to the dynamic `/luts/` JSON that comes from the server. Any LUTs from the server that are not in the static `luts.json` will be shown as white in `/webgateway/luts_png/`.
 - If you use `/webgateway/luts/rgb=true` then the JSON produced will include the `rgb` values for each LUT. This uses the static `luts.json`, but if there are new LUTs coming from the server (NOT in the `luts.json`) then we load those LUT files from the server.
 - The workflow for updating the LUTs in the static `luts.json` is to simply take the JSON output from `/webgateway/luts/rgb=true` and save it into `webgateway/static/json/luts.json`.

NB: The static `luts.json` in this PR doesn't yet have the "new" LUTs recently added. This enables us to test a few things. Then, we can update the `luts.json` with recent LUTs *before* merging this PR...

To test:

 - Go to `/webgateway/luts/` - this will be unchanged from before. Compare to `/webgateway/luts/?rgb=true` which has `rgb` for *every* LUT - each is an array of shape `(256, 3)`. The [recently added LUTs](https://github.com/ome/openmicroscopy/pull/6398/files) that are not yet in the static `luts.json` (e.g. `cividis.lut`) are being generated on the fly from the server.
 - Go to `/webgateway/luts_png/`. This is now being generated from the rgb values in the static `luts.json`, instead of loading LUT files from the server. This is very fast, so caching functionality is no-longer needed (has been removed). You will see white gaps for LUTs that are not yet cached in the static `luts.json`. ]
 - Check README for instructions on updating the static `luts.json`


will-moore:
cc @Tom-TBT

Tom-TBT:
Thank you for the fix Will, sorry for that.

will-moore:
We are seeing significant performance improvements with this change if the user's default group has a large number of users, so it seems to be worth including.

Tom-TBT:
Interesting. I'm curious, can you explain why the performance is affected here?
Without setting group to -1, does `getObject` search first for all users in the group before searching outside of the group?

will-moore:
@Tom-TBT The issue of slow cross-group queries when a user is in a big group are very low-level OMERO internals. I'm not sure if it's documented anywhere...

Another issue we're finding is that we can't assume that omero-web installations will have caching enabled.

The default behaviour is for caching to be disabled. Default is a dummy cache:

https://github.com/ome/omero-web/blob/96d113830f31142cb272c5a486b048cd185c51bd/omeroweb/settings.py#L456

and configuring e.g. redis is described as optional:

https://omero.readthedocs.io/en/stable/sysadmins/unix/install-web/walkthrough/omeroweb-install-rockylinux9-ice3.6.html#running-omero-web

So, what is best to do? Options:

 - Update `omeroweb/settings.py` to use some valid cache config by default. E.g. https://docs.djangoproject.com/en/5.1/topics/cache/#filesystem-caching where we'd need to figure out a suitable `LOCATION` and there's warnings about locations within `MEDIA_ROOT , STATIC_ROOT, or STATICFILES_FINDERS`.
or `django.core.cache.backends.locmem.LocMemCache` - not ideal for production but could be fine as a default for occasional use such as luts?
 - Update our docs (and code) to try and get all users to enable caching? Seems like a big task just for this usage of caching, but maybe we want to take more advantage of caching if we know it's available?
 - As an alternative to caching `luts_png` in Django, we could resort to saving it within OMERO. E.g. a shared OriginalFile with `cache_key` as the name? Not as nice a caching in Django, but certainly better than no caching at-all.
 - Any other options? Could ask admins to run some `luts_png` generation on the server after updating LUTS, but not very nice!

will-moore:
Tested setting the group to the `user` group with this addition to the script above:

```
user_group_id = conn.getAdminService().getSecurityRoles().userGroupId
conn.SERVICE_OPTS.setOmeroGroup(user_group_id)
```
And logged-in to nightshade as `test-user` who's default group has many users.

However, this had no difference in performance compared with leaving the group as the user's default group. Only setting `group: -1` causes a slow-down.

will-moore:
cc @jburel @Tom-TBT I've updated the description to correspond to the proposed solution now in this PR, with options TBD...

pwalczysko:
> NB: for testing this PR, use /webgateway/luts_png/?cached=false to disable the cache.

Could you please expand on how to test ?

will-moore:
Discussed at web meeting today...
It would be nice to replace usage of `LUTS_IN_PNG` list and static `luts_10.png` with a JSON object that combines both the LUT names alongside the 256 rgb values of the LUT.

README.rst
```diff
@@ -51,6 +51,18 @@ Usage
 For running omero-web in production with NGINX, see See: `OMERO.web install`_ documentation.
 To run in development mode, see below.
 
+LUTs caching
+------------
+
+The look-up tables png available at `/webgateway/luts_png/` is generated from `rgb` values
+cached in https://github.com/ome/omero-web/blob/master/omeroweb/webgateway/static/webgateway/json/luts.json.
+The LUTs in the `/luts_png/` will always correspond to the LUTs on the server as available in JSON
```
jburel, line 59: This is confusing.

README.rst
```diff
@@ -51,6 +51,18 @@ Usage
...
+from `/webgateway/luts/`.
+If new LUTs are added to the server and are not found in the `luts.json` then the `/luts_png/` will
```
jburel, line 61: again confusing. 
luts.json is a visual representation i.e. name  and associated png

will-moore: No, `luts.json` is a json file (see this PR).

jburel: I know, I should have said "holds a visual representation but ``If new LUTs are added to the server and are not found in the `luts.json`` is not clear

README.rst
```diff
@@ -51,6 +51,30 @@ Usage
 For running omero-web in production with NGINX, see See: `OMERO.web install`_ documentation.
 To run in development mode, see below.
 
+LUTs caching
+------------
+
+The OMERO server ships with a set of look-up tables (LUTs) for rendering images. Users can also
+add their own LUTs to the server. The LUTs available on the server can be retrieved from the
+`/webgateway/luts/` endpoint as JSON data.
```
sbesson, line 59: Although it certainly does not hurt to have this documentation here, it feels a bit at odds with the rest of the README.

Should https://omero.readthedocs.io/en/stable/developers/Web/WebGateway.html be extended to cover the LUT endpoints of the OMERO.web gateway?

sbesson (commented):
Immediate thoughts while reading this proposal is that it brings us back to the coupling issues originally raised in https://github.com/ome/omero-web/pull/568. When new LUTs get added to the server, the following must happen
1- a release of OMERO.web with an updated `luts.json`
2- possibly a release of all web apps of the ecosystem to depend on the newest version of OMERO.web

Has there been any consideration about hybrid solution where the cached JSON would be retrieved and the missing LUTs would be dynamically fetched from the server when retrieving the PNG representation?

jburel:
The Upgrade of LUTS does not happen very often (first time)
The dynamic change was explored by @will-moore but due to the rarity of the upgrade and do not see the synch release a major burden

omeroweb/webgateway/views.py
```diff
@@ -2088,26 +2087,47 @@ def listLuts_json(request, conn=None, **kwargs):
...
-                "path": lut.path.val,
-                "name": lut.name.val,
-                "size": unwrap(lut.size),
-                "png_index": idx,
-                "png_index_new": i,
-            }
-        )
+        # rgb = load_lut_to_rgb(conn, lut.id.val)
+        lut_data = {
+            "id": lut.id.val,
```
sbesson, line 2107: The LUT ID might vary from server to server. This is not a problem per se especially as you are using `path+name` for normalizing but this means there is a mismatch between the cached JSON file and this endpoint - which can be seen for instance by comparing https://merge-ci.openmicroscopy.org/web/webgateway/luts/ with https://merge-ci.openmicroscopy.org/web/static/webgateway/json/luts.json

Should `id` not omitted in the cache?

sbesson (commented):
I appreciate the LUT upgrade is not frequent and there is so much time we want to invest into the dynamic aspect. Especially if the next OMERO.web release will also include the upgraded JSON with the LUTs.

I was primaril asking as the dynamic loading is implemented in the `/webgateway/luts&rgb=true` and this endpoint could possibly be used rather than loading the cache in `/webgateway/luts_png`.

Performance-wise, comparing the timings of https://merge-ci.openmicroscopy.org/web/webgateway/luts/?rgb=true to https://merge-ci.openmicroscopy.org/web/static/webgateway/json/luts.json, the former call responds with a couple of 100 additional milliseconds (~574ms vs. 257ms). So there is an open question of which metrics is acceptable.

At the API level, the new `rgb=true` parameters in the `/webgateway/luts/` endpoint are backwards compatible. However, as described in the PR, the behavior of `/webgateway/luts_png/` is completely modified. At minimum, the docstring should be updated to reflect the new expectation. This also raises the question of whether these changes are significant enough that they should be considered as backwards incompatible .
Testing wise, the previous implementation suggests updates to the LUT name/path would be reflected in these endpoints. Is this also true in the new implementation and should this be tested?

will-moore:
Thanks for the in-depth reviews...

> Has there been any consideration about hybrid solution where the cached JSON would be retrieved and only the missing LUTs would be dynamically fetched from the server when retrieving the PNG representation?

I actually think this would make a lot of sense. The performance issues we saw previously were because we were fetching all 47 LUTs from the server, whereas if you're only fetching 1 or 2 then this won't be a problem. This would mean the behaviour of `/webgateway/luts_png/` would be unchanged (always renders all LUTs that are on the server).

The `574ms` timing for https://merge-ci.openmicroscopy.org/web/webgateway/luts/?rgb=true above includes loading ~8 LUTs from the server (which are not yet in the static luts.json).

We could consider re-enabling the caching of the `/luts_png/`, for those rare times when LUTs are added (but it wouldn't be essential for users to enable caching).

Updates to LUTs name/path will be reflected in the `/webgateway/luts` json. Testing would be nice but is not so easy and if there is an issue then it's likely due to the OMERO api since all we're going here is converting that output to JSON.

omeroweb/webgateway/views.py
will-moore: Yes `id`s will change, but are harmless. If we want to allow the easy copying of `/webgateway/luts/?rgb=true` into the static `luts.json` but ALSO exclude ids, then maybe we need another parameter `?ids=false`. Easy to do if worth it?

README.rst
```diff
@@ -51,6 +51,30 @@ Usage
 For running omero-web in production with NGINX, see See: `OMERO.web install`_ documentation.
 To run in development mode, see below.
 
+LUTs caching
+------------
+
+The OMERO server ships with a set of look-up tables (LUTs) for rendering images. Users can also
```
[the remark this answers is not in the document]
will-moore: Fixed in 961d30774

jburel:
> Updates to LUTs name/path will be reflected in the /webgateway/luts json. Testing would be nice but is not so easy and if there is an issue then it's likely due to the OMERO api since all we're going here is converting that output to JSON.

A test focusing on the OMERO API call will help in that case

will-moore:
I think adding tests for the OMERO API (if they don't already exist) is probably outside the scope of this PR.
I'm not exactly sure how a test would check that the LUTs listed by the script service match what's on the server?

To focus on what needs to be done on this PR:
 - should I update the `webgateway/luts_png/` to load the non-cached LUTs from the server (as we do for `/luts/?rgb=true` JSON)?
 - should I add back Django caching for `webgateway/luts_png/` as before this PR (for the times when it could be useful)?

will-moore:
To test the performance of loading 8 LUTs from the server, compare response times for `/webgateway/luts_png/?cached=false` (no LUTs loaded from server) with `/webgateway/luts_png/?cached=false&new=true`.

NB: we always use `?cached=false` so we don't get cached response. Without that, it should be even faster.

will-moore:
Tested on merge-ci by alternatively running these 2 commands in the webclient Devtools Console:

```
fetch("https://merge-ci.openmicroscopy.org/web/webgateway/luts_png/?cached=false")
fetch("https://merge-ci.openmicroscopy.org/web/webgateway/luts_png/?cached=false&new=true")
```
The `&new=true` call, when we are loading 8 LUTs from the server takes about 400-500ms server response time (not including the SSL time, connection time etc) compared to about `200ms` for the other call, when we're not loading LUTs from the server.

To err on the side of safety, I suggest:
 - By default we DON'T load LUTs from the server.
 - If a LUT is added to the server, it won't initially show up in the `webgateway/luts_png/` (white placeholder instead).
 - If the server has cache enabled, then a single visit to `webgateway/luts_png/?cached=false&new=true` (E.g by an Admin) will be sufficient to load the LUTs once and populate the Django cache with the complete png, so that all other users will benefit.

This is actually the current behaviour. So if this sounds good, I can update the README to explain this workflow then we should be good to go?

will-moore:
I'll add the "new" LUTs into the static file now...

sbesson (commented):
This PR tries to fix the LUT retrieval in two separate deployments configuration:

1. if no OMERO.web cache backend is configured (via `omero.web.caches`), the implementation will always load the static JSON file and return a preview which might include white gaps if additional LUTs are present server-side unless `new=true` is passed. This is consistent with the idea to return quickly
2. if an OMERO.web cache backend is configured e.g. Redis
   - if a cache entry with a key hashed from the list of LUTs exists and `cached=False` is not set, it is returned as a response
   - otherwise, as above the static JSON file is read and used to generate a preview image. White gaps might be present if there are more LUTs server-side than in the static JSON and `new=true` is not passed
   -  the preview image is then stored in the cache

My primary issue is associated with the second scenario and the impact of the double cache and the issues associated with troubleshooting and invalidating such a cache. One example of complex scenario:
- LUTs are added server-side e.g. via a new OMERO.server release 
- OMERO.web is either not yet released or not deployed with an matching LUTs JSON file
- the first call to the LUT preview endpoint finds no cache associated with the new hash. The implementation regenerates a preview image including white gaps and stored in the cache
- every subsequent call to the endpoint will read the cached preview image with white gaps even if OMERO.web is upgraded with a new JSON including the new LUTs
- the only way to fix this situation would be to invalidate the cache

Is there a way to detect such a scenario and force the LUT loading as a one-off?

In general, the docstrings should be updated to describe the new logic has changed. The README addition is useful but as this is documenting an occasional workflow including possibly server-side change, I think it would be more relevant under https://omero.readthedocs.io/en/stable/.

will-moore:
@sbesson That commit 3bfb3fc should address your points.

sbesson (commented):
Retested in the context of a deployment including:
- OMERO.server 5.6.14 i.e. with the new LUTs introduced in https://github.com/ome/openmicroscopy/pull/6398
- OMERO.web built from this PR with [3175ede](https://github.com/ome/omero-web/pull/603/commits/3175ede18d6fb618a54b4d3aaef8cbf85bdde3db) reverted to include a partial JSON and configured to use Redis as the cache & sessions backend
- the Redis cache cleaned before testing

The current behavior is the following:
- loading the channel colors e.g. from the Preview tab includes the new LUTs but displays them with a white preview. The server logs confirm that no original files are read on disk and `redis-cli` does not include a LUT hash key
- calling `/webgateway/luts_png?cached=false` manually loads the new LUTs as can be seen from the server logs, stores the png in the Redis cache using a LUT hash and fixes the display in OMERO.web 

Trying to summarize the behavior of OMERO.web for the last versions:

- OMERO.web 5.27.2 and earlier: the LUT preview is statically loaded from a PNG file. New LUTs added to the server are not displayed as new rows until a new OMERO.web is released with a new version of the static PNG file.
- OMERO.web 5.28.x: if applicable, the LUT preview is loaded from Redis. Otherwise the LUT preview is dynamically generated from the server and cached in Redis is configured. When no cache is configured, this means the LUTs are parsed every time.
- OMERO.web 5.29.0 (this PR): if applicable, the LUT preview is loaded from Redis.  Otherwise the LUT preview is dynamically generated using the list of LUTs from the server and the rendering value from a static JSON file. Missing LUTs should only be loaded from the server unless `cached=false` is passed explicitly as a query parameter otherwise the rows will display as blank.

Overall, the proposed behavior certainly addresses one of the issue found in OMERO.web 5.28.x without Django cache and the repeated server calls to load LUTs when hitting to `/webgateway/luts_png`. However two comments:
- in the general sense, it feels we are moving back to a pre OMERO.web 5.28.0 behavior where the dynamicity is removed and the static JSON file will need be updated and released following the introduction of new LUTs server-side 
- for deployments using a Django cache, the LUT caching will only happen if `/webgateway/luts_png?cached=false` is called manually. Is the expectation that the users or the sysadmin would call `/webgateway/luts_png?cached=false` manually on upgrade? If not, it really feels like in practice, the response is the same as deployments without cache. I don't see this as a huge problem as the LUTs are not upgraded so often currently but  this makes me wonder whether we should delete the caching logic entirely.

I have not tested the impact of `omero_group=None`, what is the ideal functional scenario?

will-moore:
The `omero_group=None` was the original change in this PR and was because `conn.getObject("OriginalFile", lut.getId()._val)` was returning `None` on "nightshade" server with group:-1 (wasn't causing an issue on merge-ci). I think that was fixed on nightshade by updating `user` group permissions.

However, I have recently seen the same issue on my Docker OMERO server, reported at https://github.com/ome/omero-figure/pull/605#discussion_r1970134502

I tested originally with this script:
```
from omero.gateway import BlitzGateway
from omero.cli import cli_login

with cli_login() as cli:
    conn = BlitzGateway(client_obj=cli._client)

    # with this line removed, the code below works without error
    conn.SERVICE_OPTS.setOmeroGroup(-1)

    scriptService = conn.getScriptService()
    luts = scriptService.getScriptsByMimetype("text/x-lut")
    luts.sort(key=lambda x: x.name.val)

    for i, lut in enumerate(luts):
        orig_file = conn.getObject("OriginalFile", lut.getId()._val)
        if orig_file is not None:
            print(f"{lut.id.val} {orig_file.getPath()} {orig_file.getName()}")
        else:
            print(f"NOT FOUND {i}: {lut.getName()}")
        lut_data = bytearray()
        # Collect the LUT data in byte form
        for chunk in orig_file.getFileInChunks():
            lut_data.extend(chunk)
```

will-moore:
My feeling is that it's really trivial to simply go to `/webgateway/luts_png?cached=false` after updating LUTs. And it encourages Admins to enable caching which is the direction we're moving?

Either way, we're in a much better place than before, and LUTs updates are rare, so I'm OK either way.

jburel:
The dynamic aspect might be reduced but I agree with @will-moore that the situation has improved: we have moved away from hardcoded list in app e.g. figure and assumption that the display matches the order of the lut.
We are significant amount of time on a feature that has occurred 1 in probably 10 years.

sbesson (approved):
Discussed with @will-moore @jburel and @knabar
- the changes should address the performance degradation observed with the new LUT PNG endpoint when no Django cache is configured
- overall, our handling of new LUTs is more robust than it was before. With the latest changes, the new implementation effectively balances user considerations of exposing all LUTs available server-side versus peformance considerations, trying to avoid the reading of original file on every webgateway call.

With the update of the JSON to match the LUTs shipped in the current OMERO.server release, there are really 2 scenarios to consider moving forward:
- inclusion of new look-up tables in the OMERO binaries. In that case, the server changes should be mirrored in a concurrent OMERO.web release. The README additions here include the required steps at the OMERO.web level but this might be something that could be documented in a separate location since it touches several components
- addition of new look-up tables by an admin to an OMERO deployment. In that case, there is no expectation the JSON in OMERO.web will reflect the custom tables and these will render blank in the UI unless this is in the Django cache

The suggestion from @knabar was to address this via recommendations in a new section of https://omero.readthedocs.io/en/stable/sysadmins/troubleshooting.html which seems like a good way forward

will-moore:
Just looking where we could add LUTs docs for the server. I don't see anywhere obvious. Maybe under https://omero.readthedocs.io/en/stable/sysadmins/index.html#managing-omero ?

jburel:
https://omero.readthedocs.io/en/stable/sysadmins/troubleshooting.html

will-moore:
Yes, I can put into troubleshooting: "If you don't see new LUTs in iviewer or figure, here's how to fix it -> https://github.com/ome/omero-web/?tab=readme-ov-file#luts-caching

But do we want to put any docs "Here's how to add LUTs to your OMERO server"? for those who may not know?

sbesson:
@will-moore at least from my side, I don't expect OMERO system administrators to be the primary constituency for the custom lookup tables documentation. Minimally, this group might be interested in ensuring custom LUTs are preserved on server upgrade similarly to https://omero.readthedocs.io/en/stable/sysadmins/server-upgrade.html#merge-script-changes.

The concept that is the closest to LUTs in terms of installation, extensibility and management is probably OMERO scripts. https://omero.readthedocs.io/en/stable/developers/scripts/index.html is a section which deals with writing and adding scripts. I don't know if an OMERO guide has been created to that purpose as this might be another location for this type of workflow.

PR #604, opened 2025-01-17
Right panel perf

will-moore:
Improves performance of loading right panel for single Objects (Project, Dataset, Image, Screen, Plate, Well, PlateAcquisition)

I looked into the timing for `getObject()` with `group: -1` and found that is slower when the user is logged-in to a large group (even if that group is private).

E.g. about `1 second` in the code below.

It's actually faster to:
 - perform a preliminary `queryService.get()` to get the object, get it's group ID...
 - and then use that group ID for `getObject()`
 - These 2 queries take about `0.02 seconds` in the code below.

```
import argparse
import sys
from datetime import datetime

from omero.cli import cli_login
from omero.gateway import BlitzGateway

def main(argv):
    parser = argparse.ArgumentParser()
    parser.add_argument('dataset', type=int, help='Dataset ID')
    args = parser.parse_args(argv)

    with cli_login() as cli:
        conn = BlitzGateway(client_obj=cli._client)
        dataset_id = args.dataset
        for test in range(10):
            start = datetime.now()
            if test % 2 == 0:
                # Use group: -1 via get(), then set group for conn.getObject()
                group_id = conn.getQueryService().get("Dataset", dataset_id, {'group': '-1'}).details.group.id.val
            else:
                group_id = -1
            conn.SERVICE_OPTS.setOmeroGroup(group_id,)
            dataset = conn.getObject("Dataset", dataset_id)
            print(datetime.now() - start, "group_id", group_id)

if __name__ == '__main__':
    main(sys.argv[1:])
```
prints:
```
0:00:00.031096 group_id 2
0:00:01.097150 group_id -1
0:00:00.028419 group_id 2
0:00:00.999497 group_id -1
0:00:00.019667 group_id 2
0:00:01.045434 group_id -1
0:00:00.021599 group_id 2
0:00:01.070055 group_id -1
0:00:00.023816 group_id 2
0:00:00.942411 group_id -1
```

### DIFF
--- a/omeroweb/webclient/templates/webclient/annotations/annotations_share.html
+++ b/omeroweb/webclient/templates/webclient/annotations/annotations_share.html
@@ -242,17 +242,7 @@
                         {% if manager.share.isExpired %}
                             This {{ manager.share.getShareType|lower }} has expired and you no longer can make any comments.
                         {% else %}
-                        <form id="add_share_comment_form" action="{% url 'annotate_comment' %}" method="post">{% csrf_token %}
-                        <table>
-                            <tr class="hiddenField"><td>{{ form_comment.share }}</td></tr>
-                            <tr>
-                                <td>{{ form_comment.comment }}</td>
-                            </tr>
-                            <tr>
-                                <td><input type="submit" value="{% trans 'Add Comment' %}" /></td>
-                            </tr>
-                        </table>
-                        </form>
+                            Commenting is no-longer supported.
                         {% endif %}
                     </th>
                 </tr>
--- a/omeroweb/webclient/controller/container.py
+++ b/omeroweb/webclient/controller/container.py
@@ -132,8 +132,9 @@ class BaseContainer(BaseController):
         rsp_obj = None
         try:
             if self.conn.SERVICE_OPTS.getOmeroGroup() == -1:
-                obj = self.conn.getQueryService().get(obj_type, int(obj_id),
-                    {'group': '-1'})
+                obj = self.conn.getQueryService().get(
+                    obj_type, int(obj_id), {"group": "-1"}
+                )
                 group_id = obj.getDetails().group.id.val
                 self.conn.SERVICE_OPTS.setOmeroGroup(group_id)
             rsp_obj = self.conn.getObject(obj_type, obj_id)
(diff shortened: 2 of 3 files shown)


will-moore:
@joshmoore. Do you know if this PR caused any improvement in the right panel loading times?

joshmoore:
The numbers with and without the PR were similar, with the 10ish calls of 300ms or so.

joshmoore:
By the way, the workaround we finally settled on was to move all pre-created conference users into a different, backup LDAP group and then on login have them moved back to the main conference group automatically.

cc: @pwalczysko

will-moore:
@pwalczysko "this is still a problem" - Need to find a server like learning where we can test this..

pwalczysko:
Sorry to bring up another topic (but quite important imho in this context): The biggest problem performance-wise in the "many users in the group" setup is actually less the RHP (although it would help to improve this dramatically). The main problem on the user-experience front is rather the slowness of opening images in full viewer (iviewer in our case).

pwalczysko (approved):

pwalczysko (changes requested):
Sorry, was too fast with the approval. Lets wait till the GHA finishes before merging.

pwalczysko (approved):
Now lgtm

omeroweb/webclient/templates/webclient/annotations/annotations_share.html
```diff
@@ -242,17 +242,7 @@ <h1>{% trans "Annotations" %}</h1>
                         {% if manager.share.isExpired %}
                             This {{ manager.share.getShareType|lower }} has expired and you no longer can make any comments.
                         {% else %}
-                        <form id="add_share_comment_form" action="{% url 'annotate_comment' %}" method="post">{% csrf_token %}
```
sbesson, line 245: What is the relationship between these changes and the proposed performance improvements?

will-moore: There was a lot of code for loading the objects for `form_comment = CommentAnnotationForm(initial=initial)` when that `add_share_comment_form` is not supported any more.

sbesson: Understood, so minimally we are trying to minimize the number of cross-group calls to `conn.getObject[s]` when loading the right-hand panel, is that correct?

will-moore: Yep, correct.

PR #606, opened 2025-02-05
Fix issues related to images with very high channel count

knabar:
This PR will address a number of issues with images that have a very high channel count and therefore produce URLs that exceed the length limitations.

Currently addressed:
- Right-hand preview panel `render_image` URLs (inactive channels are removed from URL). Note that this does not address requests with more than the allowed number of active channels.
- Fix `copy_image_rdef_json` in `webgateway/views.py` to fetch arguments from both `GET` and `POST` as given. Also allow just storing rendering defs via `POST`, which was only possible via `GET` previously.  Allowing `POST` for all calls allows passing in settings for very large numbers of channels.

### DIFF
--- a/omeroweb/webgateway/static/webgateway/js/ome.viewport.js
+++ b/omeroweb/webgateway/static/webgateway/js/ome.viewport.js
@@ -1064,7 +1064,7 @@ jQuery._WeblitzViewport = function (container, server, options) {
   /**
    * @return {String} The current query with state information.
    */
-  this.getQuery = function (include_slider_pos, include_xy_pos, include_zoom) {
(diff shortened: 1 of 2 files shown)


sbesson:
Does this incidentally resolve https://github.com/ome/omero-iviewer/issues/321?

knabar:
> Does this incidentally resolve [ome/omero-iviewer#321](https://github.com/ome/omero-iviewer/issues/321)?

I think it will still require a small adjustment in `omero-iviewer`, but at least it should be possible to fix now.

will-moore (approved):
Functionality is working as expected and code looks good 👍 
Missing channels are treated the same as inactive channels by `/render_image/` so we get the behaviour we want.

PR #607, opened 2025-02-13
JSON API: add support for login using JSON data

sbesson:
Discovered while using the JSON API and failing to authenticate using `curl --json`. It took me a while to figure out this undocumented limitation especially since other JSON API POST endpoints expect the data to be sent as JSON.

The JSON API login endpoint delegates to the WebGateway LoginView logic which only support form data defined in the login form. This commit proposes to extend the LoginView.post() logic to also support authentication sent at JSON data.

There are a few ways to test this change:
- using `curl --json` with the appropriate headers
- using `requests.post(login_url, json=payload)` e.g. by modifying https://github.com/ome/openmicroscopy/blob/1ae1bfa5b831dc0225df48dbef7871cc6ab044f6/examples/Training/python/Json_Api/Login.py#L70 to use `json` rather than `data`

This change should not affect the ability to log in using the login page or the JSON API with form data.

### DIFF
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -3474,31 +3474,21 @@ class LoginView(View):
         and store that on the request.session OR handling login failures
         """
         error = None
-        username = None
-        username = None
-        server_id = None
-        new_connection = False
-        if len(request.POST) > 0:
-            # Authentication posted using form data
-            form = self.form_class(request.POST.copy())
-            if form.is_valid():
-                username = form.cleaned_data["username"]
-                password = form.cleaned_data["password"]
-                server_id = form.cleaned_data["server"]
-                new_connection = True
-        else:
-            # Authentication posted using JSON data
+        if request.content_type == "application/json":
             try:
                 payload = json.loads(request.body)
-                username = payload.get("username")
-                password = payload.get("password")
-                server_id = payload.get("server")
-                new_connection = True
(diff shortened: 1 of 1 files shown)


knabar (commented):
I had not seen the `len(request.POST) > 0` check before, but it should do what we want here. Code looks good to me.

sbesson:
> I had not seen the len(request.POST) > 0 check before, but it should do what we want here.

I found through debugging this was one of the ways to distinguish between a [HTTPRequest](https://docs.djangoproject.com/en/4.2/ref/request-response/)  form data and JSON data. I am still not 100% happy with this approach which looks fragile and I am considering using the value of `request.content_type` instead. Thoughts?

sbesson:
Thanks @will-moore, I hadn't tested the failing scenarios. After discussion with @knabar, pushed [42790d5](https://github.com/ome/omero-web/pull/607/commits/42790d551783d97dee01e9e83c95dbd98f497eb0) to 1- use the content type to detect JSON POST requests, 2- pass the JSON as a dictionary as an input to `LoginForm`. This significantly reduces the diff and means the login logic is largely unchanged and can use `form`

knabar (approved):

will-moore (approved):
Working fine, thanks.
Tested successful login with `data` and `json` and invalid login with `data` and `json`. All with expected results.

kkoz (approved):
Works as expected

PR #608, opened 2025-02-14
Include reason in JSON response on CSRF failure

sbesson:
If a POST/PUT/DELETE request is rejected by `CsrfViewMiddleware`, the current body of the HTTP response unconditionally suggests the rejection is due to an invalid CSRF token.  However there are other reasons that might cause the rejection of the request such as a missing or malformed Referer header when connecting to HTTPS - see https://docs.djangoproject.com/en/4.2/ref/csrf/ for more details.

One way to test these condition is to do a simple connection test using the JSON API e.g. in Python using https://github.com/ome/openmicroscopy/blob/develop/examples/Training/python/Json_Api/Login.py and test various failures conditions including

- missing `X-CSRFToken` in the header
- invalid value for the `X-CSRFToken` in the header
- missing `Referer` in the header (note OMERO.web must be deployed on HTTPS)
- malformed `Referer` value in the header (note OMERO.web must be deployed on HTTPS)

Without this change the error should be identical in all situation  making it quite hard to troubleshoot the issue without having access to the OMERO.web logs where the underlying caused is logged at WARN level.

```
{'message': "CSRF Error. You need to include valid CSRF tokens for any POST, PUT, PATCH or DELETE operations. You have to include CSRF token in the POST data or add the token to the HTTP header. "}
```

With this change, the response message should include the failure reason:

```
{'message': 'CSRF Failed: Referer checking failed - no Referer. '}

### DIFF
--- a/omeroweb/feedback/views.py
+++ b/omeroweb/feedback/views.py
@@ -177,13 +177,7 @@ def csrf_failure(request, reason=""):
     Always return Json response
     since this is accepted by browser and API users
     """
-    error = (
-        f"CSRF Error. {reason} You need to include valid CSRF tokens for any"
-        " POST, PUT, PATCH or DELETE operations."
-        " You have to include CSRF token in the POST data or"
-        " add the token to the HTTP header."
-    )
-    return JsonResponse({"message": error}, status=403)
+    return JsonResponse({"message": f"CSRF Failed: {reason}"}, status=403)
 
 
 def handler500(request):
--- a/omeroweb/feedback/views.py
+++ b/omeroweb/feedback/views.py
@@ -178,7 +178,7 @@ def csrf_failure(request, reason=""):
     since this is accepted by browser and API users
     """
     error = (
-        "CSRF Error. You need to include valid CSRF tokens for any"
+        f"CSRF Error. {reason} You need to include valid CSRF tokens for any"
         " POST, PUT, PATCH or DELETE operations."
         " You have to include CSRF token in the POST data or"
         " add the token to the HTTP header."

### CODE REVIEW

**Review by knabar — COMMENTED** [review]
Agree that this is a big improvement for debugging failures. Should the error message now be reduced to just `"CSRF Error: {reason}"`, since the extra information currently given about including the token and where it needs to be included might not be the cause of the error? Or only include it if `reason` is blank (if that can even happen)?

**Comment by will-moore** [comment]
Testing locally (without https)...

Without any CRSF token...
```
login_rsp {'message': 'CSRF Error. CSRF token missing. You need to include valid CSRF tokens for any POST, PUT, PATCH or DELETE operations. You have to include CSRF token in the POST data or add the token to the HTTP header.'}
```
With `token + `"foo"` or other invalid token
```
login_rsp {'message': "CSRF Error. CSRF token from the 'X-Csrftoken' HTTP header has incorrect length. You need to include valid CSRF tokens for any POST, PUT, PATCH or DELETE operations. You have to include CSRF token in the POST data or add the token to the HTTP header."}
```
With an old invalid token
```
login_rsp {'message': "CSRF Error. CSRF token from the 'X-Csrftoken' HTTP header incorrect. You need to include valid CSRF tokens for any POST, PUT, PATCH or DELETE operations. You have to include CSRF token in the POST data or add the token to the HTTP header."}
```

Missing `Referer` caused no issues (using http). Login success!

**Comment by sbesson** [comment]
Re https://github.com/ome/omero-web/pull/608#issuecomment-2662994592, I was able to reproduce the issue while poiting to the OME CI Web deployment. Looking at the actually status code, it is not 403 but 405 i.e. `Method Not Allowed`. I suspect this is related to the deployment configuration with a NGINX proxy which might downgrade the `POST` requests to `GET`

**Review by knabar — APPROVED** [review]

**Review by will-moore — APPROVED** [review]
Tested again with last commit. Eg. `{'message': 'CSRF Failed: CSRF token from POST has incorrect length.'}` looks good.

**Comment by kkoz** [comment]
Tested with HTTPS. Used a valid `sessionid` and the `/webclient/saveImgRDef` endpoint.
Without `Referer`
```
{
    "message": "CSRF Failed: Referer checking failed - no Referer."
}
```
Without `X-CSRFToken`:
```
{
    "message": "CSRF Failed: CSRF token missing."
}
```
With `X-CSRFToken` but without `csrftoken` cookie:
```
{
    "message": "CSRF Failed: CSRF cookie not set."
}
```
With `X-CSRFToken` set to invalid value:
```
{
    "message": "CSRF Failed: CSRF token from the 'X-Csrftoken' HTTP header has incorrect length."
}
```
With `X-CSRFToken` correct but `csrftoken` cookie invalid value:
```
{
    "message": "CSRF Failed: CSRF cookie has incorrect length."
}
```
With `X-CSRFToken` and `csrftoken` correct but not matching:
```
{
    "message": "CSRF Failed: CSRF token from the 'X-Csrftoken' HTTP header incorrect."
}
```
With matching `X-CSRFToken` header and `csrftoken` cookie:
`true`

kkoz (approved):

Issue #483, opened 2023-06-26
New Tag feature : Adding colors


Rdornier:
Having different colors for tags could be a very good option to make the webclient more user-friendly (like in Trello cards).
The colors could visually differenciate processes, categories, workflows... For example, blue tags could refer to the biological sample, green ones to the microscope acquisition and orange ones to proessing.

Forum thread : https://forum.image.sc/t/omero-tags-queries/82584

will-moore:
Items of work (if we decide to use LongAnnotations on a Tag to store a colour value)

 - Need some way to add a colour annotation to a Tag and make sure only 1 is added. This could be done in the right panel when the Tag is selected, since that will allow updating of the colour after creation. Users might expect to be able to do this in the Tag creation dialog but that will make the dialog more complex and the API end-points for Tag creation (which are already a bit complex with tags to add and remove) will also have to include colour for new Tags. So, simpler to only support editing of colour *after* creation. This will need a new URL endpoint. The Colour Annotations will need a custom namespace to indicate that they are "Colour annotations". Although OMERO permissions allow anyone in a read-annotate group to add Annotations to other's data, it might make sense to restrict this to the *owner* of the Tag.
 - When tags are loaded and we want to show colours, we'll also need to load the Colour annotations:
   - When loading Tags for right panel, which uses `/webclient/api/annotations/?type=tag&image=1`. This loads the Tags via https://github.com/ome/omero-web/blob/ba072f151247df5b62768a0fbf37891675e2802d/omeroweb/webclient/views.py#L1342 and https://github.com/ome/omero-web/blob/ba072f151247df5b62768a0fbf37891675e2802d/omeroweb/webclient/tree.py#L2049. That query will need another join/fetch etc to load Colour Annotations if we are loading Tags (but not when loading other annotations)
   - When loading Tags for the Tree (left panel). That is loaded with e.g. `/webclient/api/tags/?orphaned=true&experimenter_id=-1`  but uses the same OMERO query function as above.
   - Possibly also in the Tagging dialog.
   - When we are filtering a Dataset by Tag - uses same URL endpoint as for right panel
 
 - Maybe also in tagsearch and autotag plugins? cc @Tom-TBT

jo-mueller:
Is this still being worked on? I'm a bit late to the party, but I think a really straightforward way would be to give every `tagset` its own color. The latter, for starters, could be randomly chosen, as adding a dialogue for the color setting would probably be a bit more work.

will-moore:
[...]

PR #613, opened 2025-03-03
externalInfo display

will-moore:
Since we and others are starting to use externalInfo for OME.zarr data it is useful to display this in web. Glencoe already use this.

 - We show a `Zarr` button in the right panel toolbar if the externalInfo is found (and if it has entityType of `com.glencoesoftware.ngff:multiscales`) - see discussion below.
 - Possible to see the full `externalInfo` if needed.

Zarr button:

Screenshot 2026-03-05 at 08 53 19

Click to show:

Screenshot 2026-05-07 at 16 00 43

with details expanded...

Screenshot 2026-05-07 at 16 04 00

cc @knabar @dominikl

### DIFF
--- a/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
+++ b/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
@@ -153,7 +153,6 @@
 
               $("#show_ext_info").on("click", function() {
                 var $panel = $("#externalInfo");
-                console.log("externalInfo", $panel.length);
                 $panel.toggle();
               });
             
--- a/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
+++ b/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
@@ -150,6 +150,12 @@
                 });
 
               {% endif %}
+
+              $("#show_ext_info").on("click", function() {
+                var $panel = $("#externalInfo");
+                console.log("externalInfo", $panel.length);
+                $panel.toggle();
+              });
             
         });
     </script>
@@ -376,6 +382,17 @@
     {% include "webclient/annotations/includes/figure_scripts_menu.html" %}
     {% endif %}
     {% endif %}
+
+    <!-- External Info (Zarr url)-->
+    {% with extinfo=obj.getExternalInfo %}
+      {% if extinfo %}
+        <button id="show_ext_info" class="btn silver btn_extinfo" title="Zarr url" >
+          <span style="width: 25px; margin: 3px 2px 1px 2px; background: transparent; box-shadow: none; font-size: 11px; padding: 0px;">
+            Zarr
+          </span>
+        </button>
+      {% endif %}
+    {% endwith %}
     
 
 
@@ -432,6 +449,13 @@
 {% with extinfo=obj.getExternalInfo %}
   {% if extinfo %}
     <h3>External Info (ID: {{ extinfo.id.val }})</h3>
+
+    <!-- slice 2: to remove "s3" and replace with https -->
+    <a href="https://biongff.github.io/biongff-viewer/?source=https{{ extinfo.lsid.val|slice:"2:" }}" target="_blank" 
+       style="font-size: 13px; text-decoration: underline; display: inline-block;">
+      Open with BioNGFF viewer
+    </a>
+
     <ul style="background: white; margin: 5px 0; padding: 7px; border: 1px solid rgb(221, 221, 221); overflow: auto;">
         <li>lsid: <strong>{{ extinfo.lsid.val }}</strong></li>
         <li>entityType: <strong>{{ extinfo.entityType.val }}</strong></li>
--- a/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
+++ b/omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
@@ -354,24 +354,20 @@
 
     {% endif %}   <!-- endif image or filesetInfo -->
 
-  {% if not share %}
-    {% with files=image.showOriginalFilePaths ext=obj.getExternalInfo %}
-      {% if files or ext %}
-        <!-- show original file paths -->
-        <button id="show_fs_files_btn" class="btn silver btn_fspath"
-          {% if files %}data-showfiles="true"{% endif %}
-          title="{% if files %}Show Fileset ID and file paths on server{% else %}Show External Info{% endif %}">
-          <span></span>
-        </button>
-      {% endif %}
-    {% endwith %}
+    {% if not share %}
+    {% if image.showOriginalFilePaths %}
+      <!-- show original file paths -->
+      <button id="show_fs_files_btn" class="btn silver btn_fspath" title="Show Fileset ID and file paths on server">
+        <span></span>
+      </button>
+    {% endif %}
 
       {% if image %}
       <button id="show_image_hierarchy" class="btn silver btn_hierarchy" title="Show parent Projects & Datasets">
         <span></span>
       </button>
       {% endif %}
-    {% endif %}   <!-- endif not share -->
+    {% endif %}
 
 
     <!-- Publishing Options (Figure Scripts) -->
@@ -431,18 +427,16 @@
 </div>
 
 <!-- This is HIDDEN here, but cloned into the Fileset Info panel when shown -->
-<div id="externalInfo" style="display: none">
+<div id="externalInfo" class="debug" style="display: none; margin-top: 20px; font-size: 13px;">
 <!-- This will return None if obj.getExternalInfo method doesn't exist (omero-py 5.19.6 and earlier) -->
 {% with extinfo=obj.getExternalInfo %}
   {% if extinfo %}
-  <details style="margin: 10px 5px; font-size: 12px;">
-    <summary>External Info</summary>
-    <ul style="background: white; margin: 5px 0; padding: 5px; border: 1px solid rgb(221, 221, 221); overflow: auto;">
-        <li style="font-weight: bold;">{{ extinfo.lsid.val }}</li>
-        <li>entityType: {{ extinfo.entityType.val }}</li>
-        <li>entityId: {{ extinfo.entityId.val }}</li>
+    <h3>External Info (ID: {{ extinfo.id.val }})</h3>
+    <ul style="background: white; margin: 5px 0; padding: 7px; border: 1px solid rgb(221, 221, 221); overflow: auto;">
+        <li>lsid: <strong>{{ extinfo.lsid.val }}</strong></li>
+        <li>entityType: <strong>{{ extinfo.entityType.val }}</strong></li>
+        <li>entityId: <strong>{{ extinfo.entityId.val }}</strong></li>
     </ul>
-  </details>
   {% endif %}
 {% endwith %}
 </div>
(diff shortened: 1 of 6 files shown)


sbesson (commented):
@will-moore thanks for starting this conversation. As noted above, Glencoe is using `ExternalInfo` quite extensively so we are interested in ways to expose this metadata via the OMERO.web API and/or UI.

Starting with with your last question, my immediate feeling is that it would be most beneficial to include this functionality in the OMERO.web JSON API. A few reasons for that:

- as explained in https://omero.readthedocs.io/en/stable/developers/Web.html, the `/webclient` component should be considered as internal only. If the intent is to expose this metadata as JSON for consumption (both internal and external), `api/` is the location of choice
- `omero-marshal` already has support for `ExternalInfo` encoding/decoding so there is no complex cross-component work
- the usage of `ExternalInfo` is not limited to `Image` and can be applied to any object. Glencoe uses this on `Mask` objects to represent labels for instance. Having `api/v0/<images,rois,...>/<id>` loading and exposing the external info in a unified manner would be consistent.

On the query itself, my impression is that `externalInfo` should be treated the same way as other `details`. I don't expect the additional join to be expensive in terms of query time given these are relatively shallow objects with only a few attributes but that should certainly be reviewed as par of the functional testing.

The UI questions is probably the one where I am the less opinionated and I have a suspicion the answer might be "it depends". For Zarr data, we have used `ExternalInfo` as an mechanism to store the location of the data. The closest equivalent would be the legacy OMERO pyramids which are not exposed in the UI. The most relevant existing icon would be the `../../.` in the right-hand panel which lists the Fileset entries. 
However other `ExternalInfo` objects might be more naturally suited either as Image details or as additional attributes

will-moore:
@sbesson This is now deployed on idr-testing. e.g. https://idr-testing.openmicroscopy.org/webclient/?show=image-15160031

joshmoore:
As a quick sidenote before I disappear for a week, I'd be more than happy if we hide the acronym "LSID" from users.

dominikl:
👍 Looks good to me. What does LSID actually stand for, and what is the ExternalInfo actually used for generally @joshmoore ?

joshmoore:
LSID stands for "Life Science Identifier". https://www.lsid.info/ et al. https://zenodo.org/records/46804 is a pretty good read.

The ExternalInfo object wasn't used a lot previously but was intended to hold info of the objects that felt "too detailed" to expose in the main object. It also helps that if you have an opaque identifier that you don't have to check every table to find the object but you can look in the one external info table.

will-moore:
https://en.wikipedia.org/wiki/LSID.
@joshmoore "Hide LSID from users" - what about CLI users? - in https://github.com/ome/omero-cli-zarr/pull/167 we have `--lsid` and in general we try to make the UI reflect the data model.

Do we use another term or just e.g.

```
http://path/to/my/image.zarr
entityType: com.glencoesoftware.ngff.multiscales
entityId: 3
```

joshmoore:
> in general we try to make the UI reflect the data model.

"Attribues"? "Attachments"? :smile:

> what about CLI users? 

I certainly worry about CLI users less.

But if we need to keep it, that's fine, just more than the other terms, "LSID" was a mistake.

will-moore:
As discussed in web meeting, we probably don't want `ExternalInfo` to be so much "in your face" since it's not of interest to most users.
Could look at hiding it under `../../` or elsewhere.

will-moore:
Deployed latest changes to idr-testing and updated screenshot above.

will-moore:
To help testing, you can now set ExternalInfo via the CLI - see https://github.com/ome/omero-py/pull/453

will-moore:
In IDR (and regular OMERO if/when omero-zarr-pixel-buffer becomes adopted) we want it to be nice and easy for a user to get the OME-Zarr URL for an image they are looking at, as they can do for e.g. https://idr.github.io/ome-ngff-samples/.

We probably don't want to show "External Info" or `com.glencoesoftware.ngff.multiscales` but for Images where the extInfo EntityType is `com.glencoesoftware.ngff.multiscales`, we should show the OME-Zarr URL with a handy "copy to clipboard" button.
This could be in addition to the ExternalInfo display as above, which would support all Objects (not just images) and other EntityTypes.

will-moore:
As discussed in web meeting today:
 - Exposing externalInfo to users is not useful in IDR since this is a path/to/zarr rather than a public s3.
 - It is only useful for debugging etc, so let's hide it with a `.debug: {display: none}` css then we can easily show all such UI elements with:

```
$(‘.debug’).show()
```

joshmoore:
> I'm just asking if it will always work to replace s3 with https?

Definitely not in the general case.

omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
```diff
@@ -424,3 +441,25 @@
     <input type="text" size="30">
     <img title="Close" src="{% static 'webgateway/img/close.gif' %}" />
 </div>
+
+<!-- This is HIDDEN here, but cloned into the Fileset Info panel when shown -->
+<div id="externalInfo" class="debug" style="display: none; margin-top: 20px; font-size: 13px;">
+<!-- This will return None if obj.getExternalInfo method doesn't exist (omero-py 5.19.6 and earlier) -->
+{% with extinfo=obj.getExternalInfo %}
```
will-moore: NB: with https://github.com/ome/omero-py/pull/483 this should now be `obj.getDetails().getExternalInfo()`

will-moore:
@dominikl How do you like the current `Zarr` button (as deployed on merge-ci)?

dominikl:
Thanks Will! Looks good to me! I think this would great to have on IDR, makes it so much easier to quickly check where an image is actually coming from! 👍

will-moore:
@knabar As discussed, I've cleaned this up a bit. Still using "Zarr" rather than ExternalInfo to display this to users.
See latest info and screenshot at end of description above.

omeroweb/webclient/templates/webclient/base/base_container.html
```diff
@@ -144,6 +144,8 @@
     <script type="text/javascript" src="{% static "webgateway/js/ome.spwgridview.js"|add:url_suffix %}"></script>
 
     <script>
+        // Tip for developers... E.g. shows ExternalInfo after loading right panel
+        console.log("For dev info, use: $('.debug').show();");
```
knabar: This no longer applies

knabar:
The two places referencing externalInfo can be changed now with #483 merged

> NB: with https://github.com/ome/omero-py/pull/483 this should now be obj.getDetails().getExternalInfo()

Otherwise I think everything looks good!

will-moore:
OK, should be good now. Needed a bit of extra code since you can't do `obj.getDetails().getExternalInfo()` in the django template itself

omeroweb/webclient/templates/webclient/annotations/includes/toolbar.html
```diff
@@ -375,6 +380,17 @@
     {% include "webclient/annotations/includes/figure_scripts_menu.html" %}
     {% endif %}
     {% endif %}
+
+    <!-- External Info (Zarr url)-->
+    {% with extinfo=manager.external_info %}
```
sbesson, line 385: An Image with an associated ExternalInfo object is not necessarily an OME-Zarr dataset. You will want to filter ExternalInfo objects of the correct namespace.

Also should this be called OME-Zarr rather than just Zarr as the group is expected to adhere to the multiscales metadata conventions of the OME-Zarr specification?

will-moore: By `namespace` do you mean `entityType` or `entityId`?

"Should be called OME-Zarr" where? In the UI `Zarr` button? I just think that `OME-Zarr` for this button would be a bit verbose, given the limited space for the toolbar. 
But I'll certainly put "OME-Zarr Info" instead of "Zarr Info" in the panel below.

sbesson: Yes I meant `entityType/entityId`. `OME-Zarr info` would be a good starting point

will-moore: OK, now we only show `Zarr` button if the `entityType` and `entityId` indicate it's a "zarr" externalInfo.

will-moore:
@sbesson As discussed: removed the check on `entityId` to identify externalInfo as a zarr.

Could we include this in the next milestone?

sbesson:
Thanks @will-moore, with the OMERO.web 5.31.1 patch release completed, I think it's reasonable to evaluate this proposal for the next milestone. Could you amend the description of the PR to reflect the latest state of the proposal  and include updated screenshots? Which reviewers would you suggest in addition to @knabar and myself to represent the functional perspective from IDR?

One high-level concern is that this PR exposes the content of a namespaced ExternalInfo object currently specified in a third-party extension of OMERO.server. Although discussions have started to ship this extension and make it a first-class citizen, it is not a core OMERO concept yet. In that sense, modifying OMERO.web to support it natively feels premature and creates a precedent.
Have there been considerations on whether this should be implemented as an extension point and let deployments like IDR define how this metadata should be displayed, including e.g. adding a link to a particular third-party viewer?

will-moore:
Thanks @sbesson, I had added the latest state to the description, including screenshots. But I can tidy up the text and remove older screenshots (maybe I'll move them to a later comment to keep them for comparison).

Is it the display of `entityType` that is of concern, or the check to test whether the `entityType` is `com.glencoesoftware.ngff:multiscales`?

The display of `entityType` is simply displaying a string that is saved to the server. E.g. same as showing the `namespace` for any other Annotation in the webclient.

The test to check whether the `entityType` is `com.glencoesoftware.ngff:multiscales` was added in response to your comment above "You will want to filter ExternalInfo objects of the correct namespace".

Even though omero-zarr-pixel-buffer is not shipped with OMERO yet, it's adoption is significant and those using it would find it very useful to be able to see the `externalInfo` in the webclient. This is not just useful for IDR.
I am quite open to alternative proposals for how to display this info in the webclient.

The way that we display `externalInfo` will certainly evolve as we improve OME.zarr support in OMERO - e.g. adding links to viewers and even making those viewers configurable or making it into an extension point.
But I don't see that the current proposal here is a bad first step and it doesn't feel "premature" or that it will detract from future evolution.

@dominikl is the best person from IDR to review this.

will-moore:
Older screenshot for reference:

(includes temp BioNGFF viewer link used for demo):

Screenshot 2026-03-05 at 08 58 23

Screenshot 2025-03-14 at 17 20 13

dominikl:
👍 Thanks Will! I'd really like to have this PR in asap :-) If you have an ome.zarr which for some reason doesn't work as expected, then it is really painful to inspect the ExternalInfo at the moment. I know, for the user currently it is not of much direct value, but wrt to IDR it will be when we update to the http:// URLs (even now there are probably applications which could take the s3:// URL too).

I don't really see what the problem is with "exposing" something from the ExternalInfo, it is accessible anyway via API, there's nothing hidden/secret there, is it?

knabar:
I opened a PR against this branch at https://github.com/will-moore/omero-web/pull/4 to illustrate how we could separate the display of external info and allowing project or installation specific customizations for Zarr or other formats that may be linked via external info.

sbesson (approved):
Thanks @knabar, the latest changes included in this contribution address the concerns I raised in a previous comment  These changes expose the `ExternalInfo` metadata associated with several objects, primarily `Image` but also `Project` and `Dataset`, in the right-hand panel of the OMERO.web UI. In addition, the templates are made extensible so that an installation can customize their appearance following per the standard [OMERO.web customization](https://omero.readthedocs.io/en/stable/sysadmins/customization.html) principles.

knabar (approved):

PR #621, opened 2025-05-26
/webgateway/original_file_paths: handle images without Fileset

### LINKED ISSUE #420: original_file_paths
/webgateway/original_file_paths/ID/ failing on public nightshade.

```
{
"message": "'NoneType' object has no attribute 'id'",
"stacktrace": "Traceback (most recent call last):\n  File \"/opt/omero/web/venv3/lib64/python3.6/site-packages/omeroweb/webgateway/views.py\", line 1431, in wrap\n    rv = f(request, *args, **kwargs)\n  File \"/opt/omero/web/venv3/lib64/python3.6/site-packages/omeroweb/webgateway/views.py\", line 2689, in original_file_paths\n    fileset_id = image.fileset.id.val\nAttributeError: 'NoneType' object has no attribute 'id'\n"
}
```

This URL is used to load original_file_paths when the `/../..` button is clicked, but it isn't shown when there are no original files, as for Image 3946841.
In fact, it looks like a bot:  `HTTP_FROM = 'googlebot(at)googlebot.com'`

sbesson:
Fixes #420  

For images created via the API (or OMERO 4), this should skip the call to getImportedImageFilePaths() and return an empty dictionary

To test this change, create an image via the API e.g. using of the OMERO scripts, note its ID and make a request to  `/webgateway/original_file_paths/<id>` from your browser (the `../...` button should be hidden from the right hand panel).
The request should error without this PR and return an empty body with this PR included.

### DIFF
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -2822,6 +2822,8 @@ def original_file_paths(request, iid, conn=None, **kwargs):
     image = conn.getObject("Image", iid)
     if image is None:
         raise Http404
+    if image.fileset is None:
+        return {}
     paths = image.getImportedImageFilePaths()
     fileset_id = image.fileset.id.val
     return {


will-moore (approved):
Code looks good and working as expected on merge-ci 👍

Issue #532, opened 2024-02-08
JSON API support for Annotations


will-moore:
See https://github.com/ome/omero-marshal/pull/78#issuecomment-1932882962
and discussion below that.

Endpoints we'd want to support:

 - `api/v0/annotations` for listing all annotations
 - `api/v0/m/objects/<id>/annotations` for listing child Annotations linked to various objects (`projects`, `datasets` etc).
   - For both of these URLs we'd want to filter by Annotation type. This could be done in 2 ways...
       - `api/v0/tagannotations`, `api/v0/fileannotations` etc. which matches omero object names
       - `api/v0/annotations?type=file`.  which is what's used in `webclient/api/annotations` and *could* allow multiple types e.g. `?type=file&type=tag` (although we don't support this in webclient).
   - Also for both of those `/annotations` URLs we'd want `?ns=my_namespace`.
 - `api/v0/annotations/<id>/annotations/` to list the *child* annotations linked to parent annotation:id. Also `api/v0/annotations/<id>/tagannotations/` etc for listing Tags etc?
 - `api/v0/annotations/?annotation=id` to list the *parent* annotations linked to child annotation:id. Also `api/v0/tagannotations/?annotation=id` etc for listing Tags etc.
 - `api/v0/objects/?annotation=id` to list the *parent* objects (`projects`, `datasets` etc) linked to child annotation:id.

Question: do we want to support:
 - `api/v0/annotations/?dataset=123` for listing child Annotations linked to parent Dataset? We *do* support `api/v0/images/?dataset=123` for listing child Images of a Dataset, so it's consistent with that, but in the case of `api/v0/annotations/?annotation=123` we are listing *parent* annotations, not children, so this might be confusing?!

cc @chris-allan @sbesson @joshmoore

sbesson:
> For both of these URLs we'd want to filter by Annotation type `api/v0/tagannotations`

For me the primary concern of using each annotation type in the path is the potential explosion of endpoints both at the top-level as well as each object level.
Personally, I find a `type` query parameters and possibly a `namespace` for the annotation endpoints gives all the required flexibility.

> Also api/v0/annotations/<id>/tagannotations/ etc for listing Tags etc?

I think this depends on the decision made above. If we have `api/v0/annotations?type=<type>` at the top-level, I would use the same pattern for `api/v0/annotations/<id>/annotations/?type=<type>`

> api/v0/objects/?annotation=id to list the parent objects (projects, datasets etc) linked to child annotation:id.

[...]

PR #635, opened 2025-09-26
Marshal annotations roi shape

will-moore:
Useful to be able to use e.g. `webclient/api/annotations/?roi=1062&roi=11216&roi=1064&type=map` to load ROI and Shape annotations.


sbesson:
As discussed in https://github.com/ome/omero-web/issues/532, should the work be focused on extending the public JSON API as opposed to adding more endpoints/query parameters to the internal `webclient` API?

will-moore:
Ideally, it would be great to have annotations fully supported in the JSON API, but that's a lot of work.
I don't think the aspiration to extend the JSON API should be "opposed" to the the change proposed here. This PR is a very small change that improves the utility of the `/webclient/api/annotations` endpoint, and the work is already done.

sbesson:
Maybe taking a slight step back, is there a driver for this change beyond the fact it might be useful? Do you already have an Web app/use case that you could see directly benefiting from such an endpoint?

> I don't think the aspiration to extend the JSON API should be "opposed" to the the change proposed here. 

To me, the disconnect is at the level of statement published years ago at https://omero.readthedocs.io/en/stable/developers/Web.html#omero-web that `webgateway` and `api` are the only stable public APIs moving forward. One of the implications is that the historical endpoints of `webclient/api` used by third-party apps should be considered as legacy with the long-term goal to replace them by `webgateway` or `api` endpoints.

> Ideally, it would be great to have annotations fully supported in the JSON API, but that's a lot of work.

Fully agree that this will necessitate more than 10 lines of changes. Part of the concern might be that we haven't extended this API in years and we're out of practice. On the other side, you already did the scoping work for that particular issue in https://github.com/ome/omero-web/issues/532. Do you have a feeling of the decisions would need to be made to unblock this discussion and start moving it to the implementation phase? If the work is large, is there a way to break it down into smaller steps or would you need additional people involved?

jburel:
I understand the convenience of adding the functionality in ``webclient`` but I don't think the change should be ported to the mainline as proposed by this PR, since it is not public API
Moving forward BioFileFinder could play a key role in exploring metadata so it will be preferable to ensure that the endpoints are available through the public APIs. This might mean taking a step back and reworking portions of code.

will-moore:
webclient/api/annotations/?type=map&parents=true&image=1005 gives 

```
    lineage_d, to_query = tree.marshal_lineage(
                          ^^^^^^^^^^^^^^^^^^^^^

TypeError: marshal_lineage() got an unexpected keyword argument 'roi_ids'
```

will-moore:
What are the next steps here? I was just preparing a demo for tomorrow's OMERO BioFile Finder call and I realise that the demo relied on this PR.

 - Promote the `webclient/api/annotations/` endpoint to be a public API? This endpoint is already useful outside of the webclient. E.g.  API is of this form: `webclient/api/annotations/?roi=1062&roi=11216&roi=1064&type=map`
 - Use the same logic in the JSON `api` app and simply use omero-marshal to do the marshalling?
 - Port the query logic to `omero-py`, which is the current pattern for the JSON api. NB: the JSON api doesn't support querying multiple objects. E.g. `api/image/1/annotations` is for just 1 image, whereas `webclient/api/annotations/?image=1&image=2` supports multiple images.

PR #638, opened 2025-11-03
Allow display of ex-members data in group

will-moore:
Currently, if you set the user context (for filtering data in the tree) to a user who is not in the current group, the filtering will automatically switch to "All Members". However, this prevents filtering by an Experimenter who has left a Group: see https://forum.image.sc/t/visibility-of-data-from-a-user-that-left-the-group/117309

This PR removes that restriction, allowing you to filter by a user who is not a member of the current Group.

With these changes you can: use `?experimenter=123` or `?image-123` (image owned by ex-member) to switch the filtering  of a group to the Ex member (without changing the group context or switching to "All Members").

E.g. this workflow...

 - Filter a group by "All Members" to allow you to find an image owned by an Ex Member
 - Copy the Link to that image
 - Switch back to only showing your own images (to switch off "All Members")
 - Then go to the image link. This will switch to filtering to by Ex Member - (and will show the chosen image)

A follow-up PR could add options to the Group-User dropdown menu to check for the existence of any data that belongs to non-group members (see discussion on image.sc above).

### DIFF
--- a/omeroweb/webclient/tree.py
+++ b/omeroweb/webclient/tree.py
@@ -261,17 +261,6 @@ def marshal_experimenter(conn, experimenter_id):
     params.add("id", rlong(experimenter_id))
     qs = conn.getQueryService()
 
-    join_clause = ""
-    where_clause = ""
-    print("Is Admin: %s" % str(conn.isAdmin()))
-    # if not conn.isAdmin():
-    #     group_ids = conn.getEventContext().memberOfGroups
-    #     user_gid = conn.getAdminService().getSecurityRoles().userGroupId
-    #     if user_gid in group_ids:
-    #         group_ids.remove(user_gid)
-    #     params.addIds(group_ids)
-    #     join_clause = "join experimenter.groupExperimenterMap gem"
-    #     where_clause = "and gem.parent.id in :ids"
     q = """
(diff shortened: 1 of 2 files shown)


will-moore:
Apologies - haven't tested much! Hopefully that commit should fix that.

knabar:
@will-moore since removing the membership check happens in methods used for different templates, are there any other features impacted by this change?  Will it be possible to access data in other places now that was prohibited before?  I am trying to determine what the general security implications are.

knabar:
After additional review I don't think that this change will make any data available that has not been available before via e.g. the API calls, so from a security standpoint I have no objections.

Are there plans to address the usability questions, i.e. giving users a way to utilize this without having to manipulate the URLs directly?

will-moore:
@knabar It would be nice to add UI functionality, but could be a fair bit of work for a use-case that's not been requested by a lot of users. So unfortunately I don't think I can spend that time on it just now.

The motivation behind this PR was to align webclient permissions with the server (more like a bug-fix). This enables you to filter a group by ex-members using the workflow in the description; not ideal but at least possible.

will-moore:
Does this PR need anything else to be done or is it good to merge?

PR #639, opened 2025-11-13
Split view user label

Rdornier:
Hello @knabar, @will-moore,

Small contribution on Split_View_Figure plugin, asked by one of our users. I added support to enter a custom label, which enable users to also not put any labels.

image

The corresponding python script will be updated accordingly in another PR


Rdornier:
Should be fine now

Rdornier:
Thanks @will-moore for the review.
I've put the custom label in a new row, to not interfere with the above table.

Issue #641, opened 2025-11-27
Histogram: Big Images, log scale and gamma option


will-moore:
Feature request from https://forum.image.sc/t/olympus-whole-slide-scanner-histogram-feature-not-available-in-omero-iviewer/56308/7

Main issue: support for histograms from BIG images.

Also: .... we have some wishes for the improvement of the histogram:

 - Add logarithmic scale option and

 - Add possibility for gamma value adjustment

sbesson:
On the backend side, support for computing the histogram of pyramidal images using an appropriate resolution level was introduced:
- in the image region micro-service via http://github.com/glencoesoftware/omero-ms-image-region/pull/125
- in the server via https://github.com/ome/omero-server/pull/162

These endpoints are routinely used for large XY images in OMERO Plus and PathViewer - see https://www.glencoesoftware.com/blog/2023/07/14/whats-new-in-omero-plus.html. I suspect OMERO.web and other viewers should be able to similarly relax the checks for image dimensions assuming they are talking to a recent version of OMERO.server

PR #643, opened 2025-11-28
Don't disable the histogram for big images

will-moore:
See #641 and request at https://forum.image.sc/t/olympus-whole-slide-scanner-histogram-feature-not-available-in-omero-iviewer/56308/7

# TODO: do the same for iviewer!

As @sbesson mentioned on that issue, the server does indeed support histograms on Big images!

To test:

 - Show histogram on a big image
 
Screenshot 2025-11-28 at 17 37 00

### DIFF
--- a/omeroweb/webclient/templates/webclient/annotations/metadata_preview.html
+++ b/omeroweb/webclient/templates/webclient/annotations/metadata_preview.html
@@ -473,14 +473,14 @@
             <input id="showactive" type="checkbox" />
             Active
           </label>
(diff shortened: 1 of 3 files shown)


will-moore:
We already handled the `omero.ApiUsageException` in the JSON loading.
With the exception message displayed in an `alert()` as added in this PR, it looks like this:

Screenshot 2026-05-22 at 13 48 30

I expect we'd see the same thing with older servers that don't support big image pyramids too

PR #645, opened 2025-12-02
Update changelog for 5.30.0 release

knabar:
Update changelog for 5.30.0 release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 ## Features
 
 - Allow installation of sub-packages as plugins [#619](https://github.com/ome/omero-web/pull/619)
-- Wrap top links if they don't fit in one row [#612](https://github.com/ome/omero-web/pull/612)
+- Make top links scrollable if they don't fit in one row [#612](https://github.com/ome/omero-web/pull/612)
 
 ## Bug fixes
 


UPGRADING.md
```diff
@@ -4,6 +4,25 @@ This document highlights steps that may need to be taken by developers
...
+OMERO.web 5.30.0 supports Django versions 4.2 to 5.2.
+
+Django 5.2 will be required in the next minor release.
+
+### Session serialization in Django
+
+Due to the deprecation and removal of the PickleSerializer session serializer in Django, the 
+sessions store must be cleared before restarting OMERO.web following this upgrade - see
+https://omero.readthedocs.io/en/stable/sysadmins/omeroweb-upgrade.html#clear-the-sessions-store-optional
+for more information.
```
sbesson: By default `pip install omero-web==5.30.0` on an existing deployment will not upgrade Django from 4.2 to 5.2, right?

Do we want to state this cleanup only applies if Django being upgraded or is it easier to suggest it should be done in all cases?  This probably means we'll need to duplicate this recommendation for the next minor release of OMERO.web as end-users who are still on Django 4.2 will be unconditionally upgraded to Django 5.2

knabar: > By default `pip install omero-web==5.30.0` on an existing deployment will not upgrade Django from 4.2 to 5.2, right?

Correct

> Do we want to state this cleanup only applies if Django being upgraded or is it easier to suggest it should be done in all cases? This probably means we'll need to duplicate this recommendation for the next minor release of OMERO.web as end-users who are still on Django 4.2 will be unconditionally upgraded to Django 5.2

I think it's simpler to just require it now (and again with the next release)

will-moore:
One minor comment above, but otherwise looks good 👍

PR #649, opened 2025-12-08
Strip line endings from LUTs in text format

Tom-TBT:
If a LUT is uploaded to a server in Windows CRLF style, the `load_lut_to_rgb` fails to read the LUT value (so the luts.json or luts.png can't be generated on the fly).

The line endings can be changed in the LUT file directly https://askubuntu.com/a/803166
But this would prevent the issue in the first place.


Tom-TBT:
Thanks Seb, I applied your suggestion. Tested and give the same result as windows file style

PR #653, opened 2026-01-14
Update imgData endpoint to return more metadata about the resolutions

### LINKED ISSUE #584: Improve the resolution descriptions of the imgData endpoint
Bio-Formats and OMERO introduced the ability to work natively with multi-resolution file formats 12 years ago. See notably the PR introducing the sub-resolution API in Bio-Formats [^1] and in OMERO [^2]. In OMERO.web in particular, the return value of `PixelBuffer.getResolutionDescriptions()` is getting converted into a 1D array of floating point values representing the zoom level scaling computed as the ratio of the sizeX for the current resolution divided by the sizeX of the largest resolution and this `zoomLevelScaling` list is exposed in the `webclient/imgData` endpoint of OMERO.web [^3].

The approach above has been working quite successfully for the last 12 years and provides an simple representation of the resolution levels available for a given image. In the scenario where the  image is downsampled uniformly according to a scale factor without rounding issues e.g. `[48000,24000], [12000,6000], [3000,1500], [750,375]`, this 1D representation is sufficient.
However we have been recently exposed to limitations of this representation in the scenarios where:
- the ratio between zoom level can be represented exactly as a decimal number e.g. scaleFactor of 3 or dimensions that do not divide exactly into an integer
- there is some anisotropy in the X vs Y downsampling for certain resolution levels

In these scenarios, computing the sizes of each resolution using the size of the largest resolution (i.e. the `sizeX` and `sizeZ`) and the knowledge of the `zoomLevelScaling` can easily lead to off-by-one (or more) errors in the sizeY [^4]. This is problematic as the computed resolution dimensions are translated into tile requests which in turn exceed the effective dimensions.

From a quick investigation of mathematical workaround, there is no canonical way to fix all scenarios using the current API. Instead a more robust approach would be to update the `webclient/imgData` endpoint and return a new field that marshals and exposes the fullness of `PixelBuffer.getResolutionDescriptions()` including the sizes alongside each dimension for each resolution level. If available, clients including the OMERO.web viewers should be able to consume these dimensions directly rather than recomputing them internally and make the appropriate requests
 
[^1]: https://github.com/ome/bioformats/pull/118
[^2]: https://github.com/ome/openmicroscopy/pull/634
[^3]: https://github.com/ome/omero-web/commit/4c8de98a6e24eb671474681729e7d73aa394102f
[^4]: https://github.com/glencoesoftware/qupath-extension-omero-web/issues/8

sbesson:
Fixes #584 

The response to the `webgateway/imgData` endpoint uses `ImageWrapper.getZoomLevelScaling` which reduces the resolution description as a one dimensional array of scalar zoom levels.
This reduction suffers from limitations both due to rounding errors as well as resolution levels where sizeX and sizeY might have different zoom level scaling.
This commit marshals the sizeX/sizeY pair returned from PixelBuffer.getResolutionDescriptions into a new dictionary for all resolutions and returns it as an additional `resolutions` field in the JSON response for multi-resolution images.

The proposed value of the `resolutions` attribute mirrors `zoomLevelScaling` by using a dictionary where each key is the resolution index and only being populated for images with more than one resolution. As a possible discussion point, `resolutions` could be populated for all images independently of the number of resolution level.

/cc @melissalinkert @kkoz

### DIFF
--- a/omeroweb/webgateway/marshal.py
+++ b/omeroweb/webgateway/marshal.py
@@ -185,11 +185,17 @@ def imageMarshal(image, key=None, request=None):
     rv["tiles"] = tiles
     if tiles:
         width, height = image._re.getTileSize()
-        zoomLevelScaling = image.getZoomLevelScaling()
-
         rv.update({"tile_size": {"width": width, "height": height}, "levels": levels})
-        if zoomLevelScaling is not None:
-            rv["zoomLevelScaling"] = zoomLevelScaling
+
+        resolution_descriptions = image._re.getResolutionDescriptions()
+        rv["resolutions"] = {
+            index: {"sizeX": r.sizeX, "sizeY": r.sizeY}
+            for index, r in enumerate(resolution_descriptions)
+        }
+        rv["zoomLevelScaling"] = {
+            index: r.sizeX / resolution_descriptions[0].sizeX
+            for index, r in enumerate(resolution_descriptions)
+        }
 
     nominalMagnification = (
         image.getObjectiveSettings() is not None


will-moore:
The test has `assert 'zoomLevelScaling' not in img_data` but with this PR, `img_data` will always contain `zoomLevelScaling`.

I understand we don't want to call `_re.getResolutionDescriptions()` more than once, but if there's only a single item then don't set `zoomLevelScaling`

knabar:
> My reading of the current implementation is that for single-resolution images, the `tiles` boolean will be `False` and the `zoomLevelScaling` population will be skipped altogether as per

After another reading I agree, the presence of `zoomLevelScaling` should not change.  If tests pass, :+1: from me.

jburel:
Tests are green

will-moore (approved):
LGTM 👍

knabar (approved):

mabruce (approved):
Changes look good to me.

PR #676, opened 2026-05-26
Fix tree child counts for Tags view to enable paging in center panel

knabar:
Large datasets only load a limited number (by default 200) images into the left-hand tree view and center panel. Under the "Explore" tag, if a dataset has more than 200 images, paging links are added to the bottom of the center panel. Paging updates both the tree and center panel.

For the "Tags" tab, if a tag has more than 200 matching images, the same limitation of loading only the first 200 existed, but the paging links at the bottom of the center panel were not added.

The apparent cause for this was that the "child count" property for the items in the tree (tag sets and tags) was not in fact the number of images, but rather tags.  This worked well for showing how many tags a tag set has, but failed to indicate the number of images and subsequently enable the paging if there are more than 200.

This PR changes the behavior of the tag tree API call, which now returns as the "child count" the number of tags for tag sets, and the number of images for tags.  This enables paging for large numbers of images, but retains the count shown in the tree for tag sets.

Other tagged items (datasets, projects, etc.) are still not considered in the indicated count.

cc: @muhanadz

### DIFF
--- a/omeroweb/webclient/tree.py
+++ b/omeroweb/webclient/tree.py
@@ -1238,8 +1238,16 @@ def _marshal_tag(conn, row):
     @type row L{list}
 
     """
-    tag_id, text_value, description, owner_id, permissions, \
-        namespace, tag_child_count, image_child_count = row
+    (
+        tag_id,
+        text_value,
+        description,
+        owner_id,
+        permissions,
+        namespace,
+        tag_child_count,
+        image_child_count,
+    ) = row
 
     tag = dict()
     tag["id"] = unwrap(tag_id)
--- a/omeroweb/webclient/tree.py
+++ b/omeroweb/webclient/tree.py
@@ -1229,6 +1229,8 @@ def _marshal_tag(conn, row):
       * details.owner.id (rlong)
       * details.permissions (dict)
       * namespace (rstring)
+      * tag child count (rint)
+      * image child count (rint)
 
     @param conn OMERO gateway.
     @type conn L{omero.gateway.BlitzGateway}
@@ -1236,7 +1238,8 @@ def _marshal_tag(conn, row):
     @type row L{list}
 
     """
-    tag_id, text_value, description, owner_id, permissions, namespace, child_count = row
+    tag_id, text_value, description, owner_id, permissions, \
+        namespace, tag_child_count, image_child_count = row
 
     tag = dict()
     tag["id"] = unwrap(tag_id)
@@ -1252,10 +1255,10 @@ def _marshal_tag(conn, row):
         and unwrap_to_str(namespace) == omero.constants.metadata.NSINSIGHTTAGSET
     ):
         tag["set"] = True
+        tag["childCount"] = unwrap(tag_child_count)
     else:
         tag["set"] = False
-
-    tag["childCount"] = unwrap(child_count)
+        tag["childCount"] = unwrap(image_child_count)
 
     return tag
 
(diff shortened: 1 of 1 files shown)


sbesson:
A few integration tests are failing due to these changes and will need to be adjusted - https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/137/testReport/

sbesson (commented):
Functionally evaluated with a 702 images dataset where all the images (as well as the datasets) were annotated with the `paging` tag

Without this PR

Image

With this PR

Image

Overall, the experience is much more consistent with the `Explore` tab with both counts values available for all tags included annotated images and the standard paging controls available in the center panel. This means the tags tab can be used to explore large collections of tagged images.

One UI discrepancy exposed by this bug fix is the fact that the total value only counts images and ignores other annotated objects (projects & datasets). Addressing this would require some additional considerations on how to represent these mixed collections which are currently not applicable to the main view and feels outside the scope of this PR. Being able to navigate through tagged images is already an improvement.

Do we anticipate any performance implication related to the extra query to retrieve the count of annotated images for each tag? If so, what would be typical orders of magnitude for the numbers of tags and images that should be included into a test environment for evaluation?

will-moore:
On merge-ci, I've set the `page_size` to `10` to enable pagination.
This is working as in the screenshots above.

The `childCount` is showing the image count for regular tags and the child tags count for Tag groups.
Both are now showing up in the Tag tree (previously, Tags didn't show image count in the tree)
 
```
{
"id": 1131,
"value": "10min",
"ownerId": 5,
"permsCss": "canEdit canAnnotate canLink canDelete canChgrp isOwned",
"set": false,
"childCount": 41
},
```

With this PR:

Screenshot 2026-05-27 at 21 47 52

Without this PR (different server):

Screenshot 2026-05-27 at 21 48 46

will-moore (approved):
LGTM

sbesson (approved):

knabar:
> Do we anticipate any performance implication related to the extra query to retrieve the count of annotated images for each tag? If so, what would be typical orders of magnitude for the numbers of tags and images that should be included into a test environment for evaluation?

@sbesson I've run some repeated timings to compare the old code vs. new code against one of our dev servers; for the top-level call response time increased by about 15% on average, for an individual tag the response time was basically unchanged:

new | old |   |  
-- | -- | -- | --
top level | top level | Change | Average change
0.74747 | 0.67424 | 1.10860 | 1.14593
0.74782 | 0.64762 | 1.15471 |  
0.75849 | 0.65095 | 1.16521 |  
0.73501 | 0.66690 | 1.10214 |  
0.74882 | 0.64313 | 1.16434 |  
0.73884 | 0.65473 | 1.12846 |  
0.75712 | 0.64006 | 1.18289 |  
0.74835 | 0.66653 | 1.12276 |  
0.74623 | 0.64890 | 1.14999 |  
0.79516 | 0.67377 | 1.18017 |  
tag | tag |   |  
0.61856 | 0.63391 | 0.97579 | 1.00196
0.63263 | 0.62359 | 1.01449 |  
0.62652 | 0.61669 | 1.01595 |  
0.62791 | 0.62252 | 1.00866 |  
0.63648 | 0.62842 | 1.01284 |  
0.62355 | 0.62430 | 0.99879 |  
0.62145 | 0.62830 | 0.98909 |  
0.63073 | 0.63123 | 0.99920 |  
0.62580 | 0.62770 | 0.99698 |  
0.62730 | 0.62245 | 1.00778 |

PR #679, opened 2026-06-19
Allow POST requests from public user for specific URLs

knabar:
### Important: The implementation approach has changed again as of July 29, 2026, relevant discussion starting at https://github.com/ome/omero-web/pull/679#issuecomment-5116214618.

---

With the introduction of the table slice call in https://github.com/ome/omero-web/pull/564, the option to use `POST` requests (in addition to `GET`) was included to allow for large number of rows to be fetched in one request. `GET` requests are limited by the maximum length of URIs.

An issue not previously considered is when `omero.web.public.enabled` is enabled in combination with `omero.web.public.get_only`.  In that case there is no way to use `POST` for the table slice call. All `POST` requests would have to be allowed.

~~This PR proposes a new setting to selectively allow `POST` for some URLs, similar to how a set of URLs can be enabled for public mode.~~

~~With the latest commits, this PR adds a new setting to allow flexible configuration of URLs per request method, without impacting the functionality of the current settings, i.e. if this new setting is not configured, the behavior remains unchanged.~~

~~Example:~~

```bash
omero config set omero.web.public.url_filter_by_method '{"POST": "^/webgateway/table/\d+/slice/"}'
```

~~This may not be the best or simplest approach, any thoughts or ideas on how else we could approach this?~~

~~Some tests for the old and new settings have been added via https://github.com/ome/openmicroscopy/pull/6460~~

cc: @will-moore @sbesson @jburel

### DIFF
--- a/omeroweb/decorators.py
+++ b/omeroweb/decorators.py
@@ -126,6 +126,15 @@ class login_required(object):
         raise an Exception since cleanup is intended to be immediate; if
         False, connection cleanup will be skipped ONLY when a
         ConnCleaningHttpResponse is returned.
+
+    allowPublicPost:
+        Makes the request available to the public user, independently of
+        any configured PUBLIC_GET_ONLY restriction. The PUBLIC_URL_FILTER
+        setting still applies.
+        Important: This should only be set to True for POST requests that
+        do not modify any data; usually this would be used for requests
+        that could be GET but require transmission of more request data
+        than allowed by usual URL length restrictions.
     """
 
     def __init__(
@@ -136,6 +145,7 @@ class login_required(object):
         doConnectionCleanup=True,
         omero_group="-1",
         allowPublic=None,
+        allowPublicPost=None,
     ):
         """
         Initialises the decorator.
@@ -146,6 +156,7 @@ class login_required(object):
         self.doConnectionCleanup = doConnectionCleanup
         self.omero_group = omero_group
         self.allowPublic = allowPublic
+        self.allowPublicPost = allowPublicPost
 
     # To make django's method_decorator work, this is required until
     # python/django sort out how argumented decorator wrapping should work
@@ -277,6 +288,8 @@ class login_required(object):
                 )
                 settings.PUBLIC_ENABLED = False
                 return False
+            if self.allowPublicPost and request.method == 'POST':
+                return settings.PUBLIC_URL_FILTER.search(request.path) is not None
             if settings.PUBLIC_GET_ONLY and (request.method != "GET"):
                 return False
             if self.allowPublic is None:
--- a/omeroweb/decorators.py
+++ b/omeroweb/decorators.py
@@ -288,7 +288,7 @@ class login_required(object):
                 )
                 settings.PUBLIC_ENABLED = False
                 return False
-            if self.allowPublicPost and request.method == 'POST':
+            if self.allowPublicPost and request.method == "POST":
                 return settings.PUBLIC_URL_FILTER.search(request.path) is not None
             if settings.PUBLIC_GET_ONLY and (request.method != "GET"):
                 return False
--- a/omeroweb/webgateway/views.py
+++ b/omeroweb/webgateway/views.py
@@ -3680,7 +3680,7 @@ def table_get_where_list(request, fileid, conn=None, **kwargs):
         table.close()
 
 
-@login_required()
+@login_required(allowPublicPost=True)
 @jsonp
 def table_slice(request, fileid, conn=None, **kwargs):
     """
--- a/test/unit/test_login_required.py
+++ b/test/unit/test_login_required.py
@@ -166,8 +166,28 @@ def test_is_valid_table_slice_call(monkeypatch):
     monkeypatch.setattr(settings, "PUBLIC_PASSWORD", "password", raising=False)
     monkeypatch.setattr(settings, "PUBLIC_GET_ONLY", True)
     monkeypatch.setattr(settings, "PUBLIC_URL_FILTER", re.compile("^/api"))
-    assert decorator_allow_public_post.is_valid_public_url(1, rf.get("/webgateway/table/123/slice")) is False
-    assert decorator_allow_public_post.is_valid_public_url(1, rf.post("/webgateway/table/123/slice")) is False
+    assert (
+        decorator_allow_public_post.is_valid_public_url(
+            1, rf.get("/webgateway/table/123/slice")
+        )
+        is False
+    )
+    assert (
+        decorator_allow_public_post.is_valid_public_url(
+            1, rf.post("/webgateway/table/123/slice")
+        )
+        is False
+    )
     monkeypatch.setattr(settings, "PUBLIC_URL_FILTER", re.compile("^/api|webgateway"))
-    assert decorator_allow_public_post.is_valid_public_url(1, rf.get("/webgateway/table/123/slice")) is True
-    assert decorator_allow_public_post.is_valid_public_url(1, rf.post("/webgateway/table/123/slice")) is True
+    assert (
+        decorator_allow_public_post.is_valid_public_url(
+            1, rf.get("/webgateway/table/123/slice")
+        )
+        is True
+    )
+    assert (
+        decorator_allow_public_post.is_valid_public_url(
+            1, rf.post("/webgateway/table/123/slice")
+        )
+        is True
+    )
--- a/test/unit/test_login_required.py
+++ b/test/unit/test_login_required.py
@@ -25,6 +25,7 @@ import re
 
 rf = RequestFactory()
 decorator = login_required()
+decorator_allow_public_post = login_required(allowPublicPost=True)
 
 
 def test_is_valid_public_url_default(monkeypatch):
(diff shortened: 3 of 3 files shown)


sbesson:
To demonstrate the original issue,  the following requests can be executed after authenticating into OMERO.web using the Console tab of the Developer tools. They should return the first four cells (2 rows, 2 columns) of a table:

- using the GET request

```
$.get("/webgateway/table/2060/slice/?rows=0,1&columns=0,1", rsp=>console.log(rsp))
```

- or a POST request


```
$.post("/webgateway/table/2060/slice/", {rows: "0,1", columns: "0,1"}, rsp=>console.log(rsp))
```

Executing the same request against a table in a public OMERO.web instance deployed with the default `omero.web.public.get_only` configuration e.g. https://idr.openmicroscopy.org/webclient/?experimenter=-1 currently succeeds when using GET but fail with a 403 error when using POST.

The ability to slice the table using POST request is critical when slicing large table using large numbers of rows/columns. While GET requests are typically restricted by the maximum URL length, the body sent via a POST request has a larger upper limit. With this code, the expectation is to support the GET and POST requests to the `/webgateway/table/<id>/slice` endpoint either authenticated or anonymously (i.e. using the public user).

will-moore:
This looks like a pretty clean solution to the issue.

The only thought I had was since *only* `GET` requests are allowed via the `settings.PUBLIC_GET_ONLY`, if you wanted to allow other methods (PUT, DELETE, PATCH, OPTIONS, and HEAD) then we probably don't want a setting for each one. So, maybe this new setting should allow all of those too, not just POST?

knabar:
> The only thought I had was since _only_ `GET` requests are allowed via the `settings.PUBLIC_GET_ONLY`, if you wanted to allow other methods (PUT, DELETE, PATCH, OPTIONS, and HEAD) then we probably don't want a setting for each one. So, maybe this new setting should allow all of those too, not just POST?

I agree that we don't want a separate setting for each method, but I'm also hesitant to make it so you have to allow all methods on a URL if you just need one. 

For maximum flexibility we could have a setting that takes a `dict` of methods and regular expressions:

```
omero config set omero.web.public.url_method_filter '{"POST": "^/webgateway/table/\d+/slice/", "HEAD": "/some-other-regex/"}'
```

Is that going too far?

knabar:
Suggestion to bundle all read (`GET`, `OPTIONS`, `HEAD`) and all write (`POST`, `PUT`, `DELETE`, `PATCH`) requests together in two settings

will-moore:
I think we can leave the `omero.web.public.get_only` setting named the same, since renaming is probably more trouble than it's worth. But this would also allow `OPTIONS` and `HEAD` requests.

For the new setting, I wonder if we should simply allow ALL http methods for the specified URLs?
Currently, if there's a URL where you want to allow GET and POST, you have to add that URL to both the `omero.web.public.url_filter` AND the new filter.
Can we imagine a scenario where you'd want to allow POST but *not* allow GET for a URL?

If not, then a new setting could be named `omero.web.public.url_filter_all`. This is a nicer name than `omero.web.public.url_filter_unsafe`, which would be the correct term for methods that are not "safe" https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP

sbesson:
> For the new setting, I wonder if we should simply allow ALL http methods for the specified URLs?
Currently, if there's a URL where you want to allow GET and POST, you have to add that URL to both the omero.web.public.url_filter AND the new filter.
Can we imagine a scenario where you'd want to allow POST but not allow GET for a URL?

I don't have any scenario in mind but it would be useful for the sake of this conversation to compile the list of non-GET endpoints available via OMERO.web.

I think some level of redundancy/consistency between these settings will be necessary. For instance, if you decide to block some URLs in `omero.web.public.url_filter` as described in https://omero.readthedocs.io/en/stable/sysadmins/public.html but declare them in the new settings, what is the expected behavior?

> I think we can leave the omero.web.public.get_only setting named the same, since renaming is probably more trouble than it's worth. But this would also allow OPTIONS and HEAD requests.

In all cases, what is proposed here is either deprecating or overriding the behavior of `omero.web.public.get_only` and we should also update the documentation of this property accordingly.

omeroweb/settings.py
```diff
@@ -621,14 +622,31 @@ def check_session_engine(s):
...
+        "PUBLIC_URL_FILTER_BY_METHOD",
+        "{}",
+        parse_json_dict_of_regexes,
+        (
+            "Dictionary of `request-method: url-regex`, where `request-method`"
+            "is a valid HTTP request method, and `url-regex` is a regular "
+            "expression that matches URLs the public user is allowed to access "
+            "using that request method. If `GET` is not one of the specified "
+            "request methods, the `omero.web.public.get_only` and "
+            "`omero.web.public.url_filter` settings apply for `GET` requests, "
```
sbesson: I appreciate the current proposal simplifies the uprade of an existing OMERO.web deployment to allow , I am concerned by the confusion it creates in terms of deployment and support. In particular, the following configuration feels ambiguous:

```
omero.web.public.url_filter=^/api|webgateway
omero.web.public.get_only=True
omero.web.public.url_filter_by_method={"POST": "^/webgateway/table/\d+/slice/"}
```

What would be the impact of requiring that `omero.web.public.get_only` must be set to `False` in order to allow a POST request to `/webgateway/table/1/slice`?

sbesson (changes requested):
While validating the backwards compatibiility of the current proposal, I found the easiest path was to write a series of unit tests testing the `is_valid_public_user()` logic under various OMERO.web settings. https://github.com/sbesson/omero-web/commit/3456ec818353dfac40b2d3fcb9988f776a55d11d contains a series of tests for the existing logic that pass both on top of `master` and with these commits included. Could you cherry-pick this commit and extend the tests with examples demonstrating how one could use `settings.PUBLIC_URL_FILTER_BY_METHOD` as an alternative method?

As demonstrated by the PR discussion, the limitation of the current configuration is that the public URL are controlled by the combination of 2 settings: `omero.web.public_url_filter` and `omero.web.public_get_only` which are insufficient for the granularity.

A dictionary of `method: regexp` entries appears to be the best compromise to add the control access based on the request method AND the request path and preserve backwards compatibility. The default configuration could be translate into `{"GET": r"(?#This regular expression matches nothing)a^"}` using the new configuration.

As per the inline comment, my biggest remaining question is whether we should `omero.web.public.url_filter` and `omero.web.public.url_filter_by_method` to be set simultaneously and what the behavior should be. In the deployments when a non-GET request must be made as the public, a possible alternative to the current proposal would be to allow the two following mutually exclusive configurations:
1. either set `omero.web.public.get_only` to `False` and leave `omero.web.public.url_filter_by_method` unset. This is the current behaviour and means all URLs specified by `omero.web.public.url_filter` would be publicly available using any request method
2. or set `omero.web.public.get_only` to `False`, translate `omero.web.public.url_filter` into a `omero.web.public.url_filter_by_method` dictionary(`{"GET": ...}`) supplemented with `POST/HEAD/...` keys as required and unset `omero.web.public.url_filter`

omeroweb/settings.py
knabar: @sbesson I'm concerned that if we require `omero.web.public.get_only` to be `False`, the result will be confusing in e.g. this situation:

```
omero.web.public.url_filter=^/api|webgateway
omero.web.public.get_only=False
omero.web.public.url_filter_by_method={"POST": "^/webgateway/table/\d+/slice/"}
```

If I now submit `PUT: /api/something`, is it allowed or not? Under the old rules (still valid in this PR), the request would be allowed, but I don't think that is the intention if somebody added the `url_filter_by_method` setting.

We could make `omero.web.public.url_filter_by_method` work only if the two old settings are unset, i.e. everything needs to be configured exclusively and explicitly via `filter_by_method`, which would still be backwards compatible and would remove the confusion.  Arguably then OMERO.web should not start up if all three settings are configured.

chris-allan:
The more I look at this and the implications the less and less I like it. I raised an alternative privately on the original source issue we had internally and I think I still prefer it: a `POST` allow decorator. The level of flexibility here is interesting but I think it's too much, too error prone and exposes risks too many users getting things wrong.

knabar:
Based on discussion with @chris-allan and @sbesson, I changed tack and implemented a new option for the `login_required` decorator.

When `omero.web.public.get_only` is set, `allowPublicPost=True` will now allow `POST` requests in addition to `GET` for methods so decorated, while leaving the URL filter restrictions in place.  This is intended for methods that do not perform any write operations even when called via `POST`, for example when a large request body needs to be included that exceeds the allowable maximum URL length.

The only endpoint affected right now is `/webgateway/table/*/slice`.

will-moore:
Just catching up with the discussion...
The decorator `allowPublicPost` looks like a nice solution 👍 . 

I wonder if we would ever advise someone to set `omero.web.public.get_only = False` (is there any use-case for that?) and if we should consider deprecating & removing that option, making `omero.web.public.get_only = True` the only behaviour we support?
That would reduce the ability of users to (accidentally?) make their installations less secure and reduce settings confusion if we do decide to add more config options in the future.

sbesson (approved):
The latest iteration definitely feels less complex than the new configuration proposed previously. It also establishes an interesting pattern allowing us to annotate individual endpoints to support specific public requests as needed in the future. For now, the `webgateway/table/*/slice` is the only endpoint that requires this special handling and the unit tests are covering the different combination of configurations.

Would https://omero.readthedocs.io/en/stable/sysadmins/public.html need to be updated  as a follow-up to document the exception to the current behavior?

In a future PR, I can see us deprecating the `allowPublic` keyword of the `login_required` decorator as this is not used anywhere in our code or any consumer I am aware of.

chris-allan (approved):
Looks great to me and is consistent with the intent without adding a lot of complexity.

To @will-moore's question: there was a time when the current structure, and this also extends to the `allowPublic` argument of `login_required()` itself, made sense but that's long past. I definitely support a separate careful review of the public user configuration options, decorator options and documentation at a later date.

Yajing826 (approved):
I tested the fix on our internal system with public datasets as a public user.
Without the PR, trying to turn on and off label groups as a public user triggered the error below:

Image

With the PR, I was able to turn label images on and off and create new displays, while saving is disabled as expected:

Image

Image

knabar:
> Would https://omero.readthedocs.io/en/stable/sysadmins/public.html need to be updated as a follow-up to document the exception to the current behavior?

@sbesson I added a note in https://github.com/ome/omero-documentation/pull/2541 - do you think this is sufficient?

PR #681, opened 2026-06-26
Relax Python<3.13 requirement for OMERO.web

sbesson:
As Django 5.2 is the requirement since OMERO.web 5.31.0, the package should be compatible with all versions of Python greater than 3.10


knabar (approved):

will-moore (approved):
👍

PR #682, opened 2026-06-30
Initial support for JSON api /annotations endpoint

### LINKED ISSUE #532: JSON API support for Annotations

will-moore:
Fixes #532 - also discussed at #635

This wraps https://github.com/ome/omero-py/pull/489 and passes query parameters to getObjects().

Adds support for `/api/v0/m/annotations/` endpoint, with various query parameters:
 - "image", "dataset", "project", "screen", "plate", "well", "plateacquisition", "roi"
 - ns
 - types of annotation, ("file", "map", "tag", "long", "comment" etc) accessed via e.g. `/api/v0/m/mapannotations/`

E.g:
 - /api/v0/m/annotations/?type=map&dataset=251&dataset=102
 - /api/v0/m/fileannotations/?project=102&limit=5&offset=1&ns=openmicroscopy.org/omero/bulk_annotations

### DIFF
--- a/omeroweb/api/views.py
+++ b/omeroweb/api/views.py
@@ -93,6 +93,10 @@ def api_base(request, api_version=None, **kwargs):
         "url:save": build_url(request, "api_save", v),
         "url:schema": OME_SCHEMA_URL,
     }
+    for atype in ("file", "map", "tag", "long", "timestamp", "comment",
+                  "boolean", "double", "xml", "term"):
+        rv["url:%sannotations" % atype] = build_url(
+            request, "api_namedannotations", v, ann_type=atype)
     return rv
 
 
--- a/omeroweb/api/views.py
+++ b/omeroweb/api/views.py
@@ -474,8 +474,9 @@ class ObjectsView(ApiView):
         group = getIntOrDefault(request, "group", -1)
         normalize = request.GET.get("normalize", False) == "true"
         # Get the data
-        marshalled = query_objects(conn, self.get_omero_type(request),
-                                   group, opts, normalize)
+        marshalled = query_objects(
+            conn, self.get_omero_type(request), group, opts, normalize
+        )
         for m in marshalled["data"]:
             self.add_data(m, request, conn, self.urls, **kwargs)
         return marshalled
@@ -883,6 +884,7 @@ class AnnotationsView(ObjectsView):
         elif ann_type is not None:
             raise BadRequestError("Invalid annotation type: %s" % ann_type)
         return super(AnnotationsView, self).get(request, conn, **kwargs)
+
 
 class ExperimentersView(ObjectsView):
     """Handles GET for /experimenters/ to list Experimenters."""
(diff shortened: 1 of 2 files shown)


omeroweb/api/views.py
```diff
@@ -850,9 +854,19 @@ def get_opts(self, request, **kwargs):
...
-        ann_type = request.GET.get("type")
 
-        # Not strictly a pure function, but we need to set the OMERO_TYPE
+        return opts
+
+    def get(self, request, conn=None, **kwargs):
+        """Override get() to allow filtering by annotation type."""
+
+        # set self.OMERO_TYPE, then call super().get() to get the list of Annotations
+        # E.g. conn.getObjects("TagAnnotation") - not actually case-sensitive
```
pwalczysko, line 879: I think that the tests of ``test_show`` in https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/43/testReport/ might be failing because of this change - the timing of the commit and the timing of the tests starting to fail matches.

pwalczysko:
excluded to verify https://github.com/ome/omero-web/pull/682#discussion_r3703765668

pwalczysko:
> excluded to verify [#682 (comment)](https://github.com/ome/omero-web/pull/682#discussion_r3703765668)

wrong call, including again, sorry

pwalczysko:
```
flake8...................................................................Failed
- hook id: flake8
```

pwalczysko:
Excluding again, as this PR is coupled with https://github.com/ome/openmicroscopy/pull/6459 and https://github.com/ome/omero-py/pull/489. Will exclude all three to be sure, thanks @sbesson

pwalczysko:
Removed the exclude label. We need to keep an eye on https://github.com/ome/omero-web/pull/682#discussion_r3703765668 tomorrow.

PR #684, opened 2026-08-14
add config "omero.web.annotation_order"

JoshuaFortriede:
This accepts values of 'namespace', 'date', or a list JSON list of namespaces.

This affects the order of the annotation groups (namespaces) in the right pane of the webclient. Currently the code hardcodes based on the namespace name. This allows to specify order based on the namespace name (default), date (of creation), or a JSON list. For the JSON list, it will order based on the list, and then default to the namespace name. 

This allows flexibility from system administrators on how they want to display the data to ensure the most relevant annotations are at the top without having to resort to ascii-ordered naming conventions.

**Examples:**
omero config set omero.web.annotation_order namespace
omero config set omero.web.annotation_order date
omero config set omero.web.annotation_order '["my_annotation/space/group4","my_annotation/space/sample2","my_annotation/space/group2"]'

### DIFF
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -222,6 +222,7 @@ def check_session_engine(s):
         )
     return s
 
+
 def parse_annotation_order(value):
     try:
         parsed = json.loads(value)
@@ -231,6 +232,7 @@ def parse_annotation_order(value):
         pass
 
     return str(value)
+
 
 CUSTOM_HOST = CUSTOM_SETTINGS.get("Ice.Default.Host", "localhost")
 CUSTOM_HOST = CUSTOM_SETTINGS.get("omero.master.host", CUSTOM_HOST)
@@ -1017,7 +1019,7 @@ CUSTOM_SETTINGS_MAPPINGS = {
             "A value of 0 or -1 pre-selects All groups."
         ),
     ],
-    "omero.web.annotation_order":[
+    "omero.web.annotation_order": [
         "ANNOTATION_ORDER",
         "namespace",
         parse_annotation_order,
--- a/omeroweb/settings.py
+++ b/omeroweb/settings.py
@@ -222,6 +222,15 @@ def check_session_engine(s):
         )
     return s
 
+def parse_annotation_order(value):
+    try:
+        parsed = json.loads(value)
+        if isinstance(parsed, list):
+            return parsed
+    except Exception:
+        pass
+
+    return str(value)
 
 CUSTOM_HOST = CUSTOM_SETTINGS.get("Ice.Default.Host", "localhost")
 CUSTOM_HOST = CUSTOM_SETTINGS.get("omero.master.host", CUSTOM_HOST)
@@ -1006,6 +1015,16 @@ CUSTOM_SETTINGS_MAPPINGS = {
         (
             "ID of the group to pre-select in search form. "
             "A value of 0 or -1 pre-selects All groups."
+        ),
+    ],
+    "omero.web.annotation_order":[
+        "ANNOTATION_ORDER",
+        "namespace",
+        parse_annotation_order,
+        (
+            "Controls annotation ordering. "
+            "Allowed values are 'namespace', 'date', or a JSON list "
+            "of namespaces defining a custom order"
         ),
     ],
     "omero.web.ui.top_links": [
--- a/omeroweb/webclient/tree.py
+++ b/omeroweb/webclient/tree.py
@@ -2008,20 +2008,18 @@ def marshal_annotations(
     service_opts.setOmeroGroup(group_id)
 
     # Determine annotation order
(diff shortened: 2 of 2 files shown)


will-moore:
Hi @JoshuaFortriede - Thanks for your PR. I've included it in our daily build for testing...

Before I forget, we'll need a CLA form from you before we can accept a contribution to any of our GPL repos. Please see https://ome-contributing.readthedocs.io/en/latest/cla.html for info and instructions. Thanks.

JoshuaFortriede:
Signed and sent.

will-moore:
Testing on merge-ci...

With the PR merged and no config:

Screenshot 2026-08-19 at 15 04 50

After setting config:

```
$ omero config set omero.web.annotation_order '["openmicroscopy.org/mapr/phenotype","openmicroscopy.org/mapr/gene","openmicroscopy.org/omero/bulk_annotations"]'
```

I'm not seeing any change in the ordering, either in the UI or in the JSON response.
I'll test locally and see if I can debug...

will-moore:
This seems to be working fine with the same config (and similar annotations) on my local server:

Query generated is:

```
select oal from ImageAnnotationLink as oal
            join fetch oal.details.creationEvent
            join fetch oal.details.owner
            left outer join fetch oal.child as ch
            left outer join fetch oal.parent as pa
            join fetch ch.details.creationEvent
            join fetch ch.details.owner 
            left outer join fetch ch.file as file
            where pa.id in (:ids) and ch.class=MapAnnotation
            order by 
            case
                when ch.ns = 'openmicroscopy.org/mapr/phenotype' then 0
when ch.ns = 'openmicroscopy.org/mapr/gene' then 1
when ch.ns = 'openmicroscopy.org/omero/bulk_annotations' then 2
                else 999999
            end,
            ch.ns
```

Result is:

Screenshot 2026-08-19 at 15 43 11

will-moore:
This seems to be working fine (not sure why it's not working on merge-ci server...).

My slight concern is that it's a bit unexpected for the sorting to happen at the point of the DB query, while relying on the JavaScript rendering of the JSON-to-html to preserve the ordering of annotations.

https://github.com/ome/omero-web/blob/116a77766563756a351f2ec3f6cf98550c7f2cc8/omeroweb/webclient/static/webclient/javascript/ome.right_panel_mapanns_pane.js#L174

However, I appreciate that's the way the sorting is already defined *before* this PR. I just worry that in the future, I (or someone else) might choose to add sorting functionality in JavaScript (e.g. with a "sort by" chooser/button for users to pick their sorting) and then the sorting from this PR could get ignored/missed.

Maybe all it needs is some comment in the JS code above to say: "Annotations are ordered in the JSON response according to the `omero.web.annotation_order` setting and this should be preserved in the rendering".
Then, if we choose to move the sorting functionality to JavaScript, we can take that setting into account.

cc @knabar

pwalczysko:
> This seems to be working fine (not sure why it's not working on merge-ci server...).

Did you check that this PR is included ?

will-moore:
@pwalczysko Yes, looked at last commit at https://github.com/snoopycrimecop/omero-web/commit/6de07acf1880daf29c6a4cd890caeb7f0be40b6e which lists this PR as included.
Let's see tomorrow if rebuild has helped...
I see that merge-ci webclient is down now. It was OK after I last re-ran the OMERO-web job - not sure what's up...

will-moore:
This is working on the `merge-ci` server now. See #686 for explanation.

As I understand it, @knabar is happy with this PR.

cc @Tom-TBT may be interested in this feature.

For me, I'd just like to see a comment in the JavaScript as mentioned above, just so future devs are aware that order of annotations in the JSON should be preserved.

Tom-TBT:
Hi, thanks Will for the ping, and thanks @JoshuaFortriede for the PR.

It works as expected and I think it's useful.

Only without setting the parameter `omero.web.annotation_order`, there is an implicit default to alphabetical ordering. So alphabetical sorting should be a valid and documented option.

JoshuaFortriede:
@Tom-TBT, yes it defaults to alphabetical ordering of the namespace which is the current method. Is there a better way you would like this to be documented?

omeroweb/settings.py
```diff
@@ -1008,6 +1019,16 @@ def check_session_engine(s):
             "A value of 0 or -1 pre-selects All groups."
         ),
     ],
+    "omero.web.annotation_order": [
+        "ANNOTATION_ORDER",
+        "namespace",
+        parse_annotation_order,
```
Tom-TBT: @JoshuaFortriede I'm thinking here first, but also in the documentation https://omero.readthedocs.io/en/stable/sysadmins/config.html
How is this documentation updated @will-moore ? Is it in a repo?

will-moore (approved):
LGTM. Would be nice to have a comment in the JavaScript as mentioned above, but not a blocker.
@knabar Any feedback? Good to merge?

PR #687, opened 2026-09-03
Preserve formatting in script activity messages

TorecLuik:
## Summary

OMERO script `Message` outputs may contain multiline or whitespace-sensitive
content, such as reports, tracebacks, tables, and command output.

The Activities panel currently renders these messages as ordinary text,
causing newlines and repeated spaces to collapse.

This PR adds narrowly scoped styling for script result messages so that they:

- begin below the floated action buttons instead of flowing alongside them;
- preserve explicit newlines and repeated spaces while still allowing long lines to wrap;
- use a monospace font for predictable alignment;
- have a small visual separation from the script title.

The selector only targets messages in script activity rows. Other activity
types are unaffected.

I'll add it via `base_include_template`, but thought you might want it upstream too?

## Testing

Manually verified in a local OMERO.web deployment by loading the same CSS rule
through `omero.web.base_include_template`.

Tested with ordinary reports and a script that returns multiline ASCII
output through its `Message` result. Explicit whitespace is preserved, while
long output continues to wrap within the Activities panel.

A minimal test script is available at:
https://github.com/NL-BioImaging/cowscript/blob/master/Cow.py

## Screenshots

| Before | After |
| --- | --- |
| Script output before formatting fix | Script output after formatting fix |

### DIFF
--- a/omeroweb/webgateway/static/webgateway/css/ome.header.css
+++ b/omeroweb/webgateway/static/webgateway/css/ome.header.css
@@ -835,6 +835,14 @@ li.menu_back a {
 			    color: #666;
 			}
 
+			#activities_content tr.script .activity_title + .message {
+				clear: both;
+				display: block;
+				margin-top: 0.25rem;
+				white-space: pre-wrap;
+				font-family: ui-monospace, Consolas, "Liberation Mono", monospace;
+			}
+
 
 			#activities_panel tr.hidden {
 			    display: none;
@@ -1197,5 +1205,3 @@ li.menu_back a {
     left: 0;
     right: 0;
 }
-
-


will-moore:
Hi, thanks for the PR.

I'm not sure that we want to encourage the usage of long formatted message strings in the activities panel.

We recommend that you use `print("log message")` within the script. Any stdout from the script will be available to users under the info (i) icon in the activities panel.
Without your PR, and with your script modified to `print(cowsay(text))`, it looks like this (once you click the (i) button):

Screenshot 2026-09-03 at 14 03 11

With your PR we get a less compact layout, so we don't see so many Activities:

Screenshot 2026-09-03 at 14 01 02

However, I'm willing to accept that there is a use-case for what you're proposing, so maybe it's worth it.
I'll discuss with others and let you know what we decide...

NB: Before we can accept any contributions to the OME *GPL repos we need a Contributors License Agreement from you - see https://ome-contributing.readthedocs.io/en/latest/cla.html for details.

TorecLuik:
Fair enough, whatever you like. 

I found personally that I'd rather have info shown in the activities than behind the (i) button. 
As in, the (i) button feels to me more like a debug log - if I really want to see details. 

So it feels like low-hanging fruit to have a bit more style there - like accepting newlines as printed.

PR #691, opened 2026-09-17
Honor the configured login view when redirecting as the public user

sbesson:
When accessing a webclient URL with a `show=<object>-<id>` query parameter as the public user and the object is not found, the application is redirected to the login page for authentication. Consistently with other redirects, this fixes the redirect target to use settings.LOGIN_VIEW rather than the hardcoded weblogin which is the default OMERO.web login page.

This also updates every call to `HttpResponseRedirect` to includes an explicit `url_has_allowed_host_and_scheme` check for absolute URLs. This validation logic was previously delegated to `validate_redirect_url` but this pattern was flagged as a false positive by the CodeQL reports. Making the check explicit fixes the report  while the `validate_redirect_url` API remains available for downstream projects. 

This PR should be tested in a standard OMERO.web deployment configured with the public user both with the default login page as well as an alternate login workflow by setting the [omero.web.login_view](https://omero.readthedocs.io/en/stable/sysadmins/config.html#omero-web-login-view) configuration.

With the current release of OMERO.web, accessing a non existing object ID via `/webclient/?show=image-99999999` should redirect to `/webclient/login` in both scenario with the original path in the `url` query parameter. With this PR included, the same request should redirect to `/webclient/login` if `omero.web.login_view` is unset or the custom login page if `omero.web.login_view` page


omeroweb/webclient/views.py
```diff
@@ -174,19 +174,6 @@ def get_bool_or_default(request, name, default):
     return toBoolean(request.GET.get(name, default))
 
 
-def validate_redirect_url(url):
```
sbesson: Given it's a patch release, we should probably keep the helper method (and maybe deprecate it) as downstream consumers might depend on it.

will-moore:
With or without this PR, NO public user or login_view configured, `/webclient/?show=image-9999999` gets redirected to:

```
/webclient/login/?url=%2Fwebclient%2F%3Fshow%3Dimage-9999999
```

To test `login_view`, I simply chose an existing view that didn't need a login (rather than creating a custom login page):

```
$ omero config set omero.web.login_view api_versions
```

Without public user enabled, going to `/webclient/?show=image-9999999` gets redirected to

```
/api/?url=%2Fwebclient%2F%3Fshow%3Dimage-9999999
 ```

With the public user enabled but *without this PR*, the same URL gets (incorrectly) redirected to:

```
/webclient/login/?url=/webclient/?show=image-9999999
```

With the public user enabled *WITH this PR* the URL redirects to:

```
/api/?url=/webclient/?show=image-9999999
```
👍 

NB: both these don't encode the url in the same way that happens without public user at https://github.com/ome/omero-web/blob/52c47fd51495d54943e2170c51ad4637a51a5d3d/omeroweb/decorators.py#L232

Now, without `omero config set omero.web.login_view` set... with public user enabled, without this PR, redirects to:

```
/webclient/login/?url=/webclient/?show=image-9999999
```

and same redirect with this PR:

```
/webclient/login/?url=/webclient/?show=image-9999999
```

sbesson:
Thanks for the thorough testing @will-moore including the scenario where the public user is disabled. Is the table below an accurate summary of your testing with/without this PR under various configurations?

|  | OMERO.web 5.33.1 | OMERO.web PR 691 |
|---|---|---|
| omero.web.login_view unset, public user disabled | /webclient/login/?url=%2Fwebclient%2F%3Fshow%3Dimage-9999999 | /webclient/login/?url=%2Fwebclient%2F%3Fshow%3Dimage-9999999 |
| omero.web.login_view unset, public user enabled | /webclient/login/?url=/webclient/?show=image-9999999 | /webclient/login/?url=/webclient/?show=image-9999999 |
| omero.web.login_view set, public user disabled | /api/?url=%2Fwebclient%2F%3Fshow%3Dimage-9999999 | /api/?url=%2Fwebclient%2F%3Fshow%3Dimage-9999999 |
| omero.web.login_view set, public user enabled | /webclient/login/?url=/webclient/?show=image-9999999 | /api/?url=/webclient/?show=image-9999999 |

Can you think of any other scenario that would need to be adjusted?

will-moore (approved):
Yes, that looks correct 👍

knabar (approved):

